PR #16, opened 2026-02-05
Refactor

kasperpawlowski:


services/pricing/priceProvider.ts
```diff
@@ -0,0 +1,556 @@
...
+
+  if (!priceInfo) {
+    return undefined
+  }
+
+  // Return price if available, even if queryFailure is true
+  // (queryFailure may indicate stale price but we can still use it)
+  if (priceInfo.queryFailure && !priceInfo.amountOutMid) {
+    return undefined
+  }
```
WingsDevelopment, line 115: what do you think about having blocknumber cached with queries like this?

This way we can compare the block numbers on UI later on and show some warning icon to the user 'this data is last time fetch at 24234 block number and might be stale' ?

WingsDevelopment: we don't have to do it now here, but just to plan things out for new project

services/pricing/priceProvider.ts
```diff
@@ -0,0 +1,556 @@
...
+export const getAssetUsdValue = async (
+  amount: number | bigint,
+  vault: AnyVault | null | undefined,
+  source: PriceSource = 'on-chain',
+  backend?: BackendConfig,
+): Promise<number> => {
+  if (!vault) return 0
+
+  const price = await getAssetUsdPrice(vault, source, backend)
+  if (!price) return 0
```
WingsDevelopment: is this okay? or do we want to return undefined here?

services/pricing/priceProvider.ts
kasperpawlowski: for the on-chain data there's no need. you either get it or not. but fyi, lens return types already expose timestamps if you'd like to use them

services/pricing/priceProvider.ts
kasperpawlowski: in this case undefined would be a better choice. I'll check if it's easy to propagate

services/pricing/priceProvider.ts
WingsDevelopment: that is cool, i was thinking every data point should maybe have fetchedAt and/or fetchedAtBlock, I am not looking to complicate things, just to design properly

PR #41, opened 2026-02-24
validate swapverifier instead of swapper

dglowinski:


dglowinski:
Swapper itself is an untrusted black box, it's only swap verifier that needs to be verified

damarnez (approved):

PR #134, opened 2026-03-30
Migrate token list and prices to v3 API

kasperpawlowski:
## Summary
- Migrate token list from `/v1/tokens` to `/v3/tokens` with pagination support (100 per page, using `meta.total` for continuation)
- Migrate off-chain prices from `/v1/prices` to `/v3/prices` with parallel chunk fetching (max 100 addresses per request)
- `BackendPriceData` now matches the v3 response shape directly (`priceUsd`, ISO-8601 timestamp)
- Consolidate `PRICE_API_URL` into `EULER_API_URL` — both tokens and prices use the same backend
- Remove hardcoded `indexer.euler.finance` URLs from docs and env defaults
- Deduplicate in-flight price requests — subsequent batch windows piggyback on pending fetches instead of re-requesting (208 → 67 addresses on initial load)
- Increase batch collect delay from 50ms to 100ms

## Notes
- Both features are best-effort with full fallback: token list falls back to Uniswap/DefiLlama, prices fall back to on-chain oracles
- The v3 staging server is slow (~3-10s/request); production should be faster

## Test plan
- [ ] Set `EULER_API_URL` to v3 backend, verify token logos display on lend page
- [ ] Verify USD prices appear on vault cards (total supply, available liquidity)
- [ ] Check network tab: requests go to `/v3/prices` and `/v3/tokens`, no duplicate address fetches
- [ ] Unset `EULER_API_URL` entirely, verify app still works (on-chain prices, Uniswap/DefiLlama tokens)
- [ ] Test on a chain with >100 tokens to verify pagination

## Summary by CodeRabbit

* **Configuration**
  * Cleared default public API URL values in examples; no hard-coded Euler domains remain. Pricing defaults removed; pricing is now routed via the Euler API configuration.

* **Improvements**
  * Token and pricing endpoints upgraded to v3 with paginated token fetching and address-chunked price requests.
  * Request batching window increased from 50ms to 100ms.

* **Documentation**
  * Deployment, setup, and pricing docs updated with placeholder endpoints and new backend API contract details.


LeonardEulerXYZ (dismissed):
## Walkthrough

### Overview
The migration to v3 token/pricing endpoints may be directionally right, but as written it hard-switches callers onto `/v3/*` under `EULER_API_URL` without proving that the configured host actually serves those paths.

### Verdict
**Changes requested**

### Findings
1. **Potential issue · blocking** — this regresses environments still pointing `EULER_API_URL` at the existing indexer host.
   - I checked the previously documented/default host (`https://indexer.euler.finance`) and `/v3/tokens` / `/v3/prices` return `404` there.
   - So this PR breaks token metadata and backend pricing unless the host migration is already guaranteed alongside it, or there is a compatibility fallback.

### Suggested next step
Either ship the required host migration in lockstep and document it clearly, or add compatibility handling/fallback so existing deployments do not hard-break.

LeonardEulerXYZ (commented):
## Walkthrough

### Overview
This migrates token metadata and off-chain prices from the old v1 indexer surface to the v3 API, removes the separate price API config, adds token pagination, chunks v3 price requests at 100 addresses, and deduplicates in-flight single-price calls.

### Review effort
Medium. I reviewed the migration paths, config fallout, token pagination, price chunking/dedup, and fallback behavior.

### Verification
- `npm run test:run` — passed: 41 files, 621 tests; 1 skipped.
- `npm run build` — passed; only existing Sentry/Reown/chunk-size style warnings observed.
- Preview smoke:
  - `/` returned 200.
  - `/api/token-list?chainId=1` returned 200 with 6,885 merged tokens.
  - v3 staging `/v3/tokens` pagination returned `data` + `meta.total` as expected.
  - v3 staging `/v3/prices` returned `data[].priceUsd` as expected.
- Local throwaway regression checks, removed before reporting:
  - `fetchBackendPrices()` chunks 205 addresses into 100/100/5 requests and maps `body.data[].priceUsd`.
  - `fetchBackendPrice()` piggybacks same-address calls while the first fetch is still in flight.

### Findings
No blocking correctness issues found in the migration code.

One non-blocking semantic point to keep explicit before merge: consolidating `PRICE_API_URL` into `EULER_API_URL` means any deployment that sets `EULER_API_URL` for token list/logos also enables the global off-chain price backend. `getUnitOfAccountUsdRate()` already tries the backend whenever configured, even when callers later request `source: 'on-chain'` through oracle paths. That behavior existed with `PRICE_API_URL`, but the config separation is gone now, so operators can no longer enable Euler token metadata while keeping off-chain price backend disabled except by leaving `EULER_API_URL` empty entirely. If that is intended, the PR is coherent; if not, keep a separate “prices enabled” knob or thread the requested source into UoA/USD lookup.

### Verdict
Commented — no blockers from my pass. The code path is a neat migration: v3 shapes line up, pagination has sensible bounds, partial price failures remain best-effort, and the in-flight dedup does what it says on the tin.

Seranged (approved):

PR #138, opened 2026-04-02
fix: make wallet screening fail-closed on all error paths

Seranged:
Previously, any error (missing config, non-2xx response, network failure, timeout) would silently pass the address through as clean. This changes all four error paths to return addressIsSuspicious: true instead, so a misbehaving or unreachable screening service blocks access rather than granting it.

The only path that returns the actual TRM result is a clean 200 response — everything else fails closed.

Tested locally against stubs covering all failure modes (missing URI, 503, ECONNRESET, 5s timeout) and verified the BlockedAddressModal appears in each case.


kasperpawlowski (commented):
use `logWarn` instead of `console.warn`

kasperpawlowski (approved):

PR #153, opened 2026-04-04
feat: add vault cache for faster first load

kasperpawlowski:
## Summary

- The RPC proxy passively intercepts vault lens calls (batchSimulation + direct eth_call) and caches raw hex results per vault address in a 5-minute TTL in-memory cache
- New `POST /api/vault-cache/:chainId` endpoint serves cached hex for requested vault addresses
- `loadVaults()` tries the cache after factory detection — on hit, renders immediately and continues the full load (including escrow, pricing, Pyth) silently in background
- Escrow vault fetching refactored to use EVC batchSimulation instead of individual readContract calls
- Extracted `processRawEarnVaultData` and `processRawSecuritizeVaultData` for cache decoding
- Generation-aware race condition protection for chain switches during cache fetch

## How it works

No client push, no pre-warming schedule, no idle RPC calls. The server is the only writer — it reads from its own proxy traffic. Cache is populated as a side effect of normal user RPC calls. Subsequent users benefit from the warm cache.

## Test plan

- [ ] First user loads app — vaults load normally (cold cache), cache warms
- [ ] Second user loads within 5 min — vaults render instantly from cache, background refresh updates
- [ ] Borrow page loads correctly (escrow vaults loaded via background refresh)
- [ ] Chain switch during cache fetch doesn't show stale chain data
- [ ] Server restart gracefully falls back to normal load

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -104,9 +104,9 @@ watch(chainId, () => {
   resetBalances()
   const targetChainId = chainId.value
   const labelsPromise = loadLabels()
-  void loadTokenList()
+  const tokenListPromise = loadTokenList()
   void loadCountry()
-  void labelsPromise.then(() => {
+  void Promise.all([labelsPromise, tokenListPromise]).then(() => {
     if (chainId.value !== targetChainId) return
     void loadVaults()
   })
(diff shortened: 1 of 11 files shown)


kasperpawlowski:
this appeared not to be a straightforward optimization

PR #161, opened 2026-04-06
fix: require CF-Connecting-IP in production to close X-Forwarded-For rate limit bypass

Seranged:
In production, consume() now rejects requests with 403 if CF-Connecting-IP is absent, preventing attackers from bypassing per-IP rate limits by rotating the X-Forwarded-For header when accessing the origin directly. Dev and stg are exempt as they don't always run behind Cloudflare.

### DIFF
--- a/server/utils/rate-limit.ts
+++ b/server/utils/rate-limit.ts
@@ -15,24 +15,30 @@ interface RateLimiterConfig {
   label: string
 }
 
-// NOTE: This rate limiter is a best-effort defense-in-depth measure, not a
-// security boundary. Known limitations:
+// NOTE: This rate limiter relies on Cloudflare being in the request path for
+// production (DOPPLER_ENVIRONMENT=prd). CF-Connecting-IP is set by Cloudflare's
+// edge and cannot be spoofed by clients going through Cloudflare. In production,
+// requests arriving without this header are rejected fail-closed, which closes
+// the X-Forwarded-For rotation attack that was possible via the old fallback path.
 //
+// Residual limitation: an attacker who knows the origin IP and bypasses Cloudflare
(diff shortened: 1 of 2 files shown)


server/utils/rate-limit.ts
```diff
@@ -70,8 +76,22 @@ export function createRateLimiter(config: RateLimiterConfig) {
...
      */
     consume(event: H3Event, cost = 1): void {
+      // In production, CF-Connecting-IP must be present. Requests that arrive
+      // without it bypassed Cloudflare entirely — reject them fail-closed.
+      // Outside production (stg, preview, dev) Cloudflare may not be in the
+      // path, so the check is skipped.
+      const cfIp = event.node.req.headers['cf-connecting-ip']
+      // Fail-closed in production when CF-Connecting-IP is absent.
+      // stg and dev are exempt: they don't always run behind Cloudflare.
+      const hasCfIp = typeof cfIp === 'string' && !!cfIp.trim()
```
Seranged, line 89: Note: the typeof cfIp === 'string' && !!cfIp.trim() check is intentional. A plain !cfIp would miss a whitespace-only CF-Connecting-IP header (e.g. cf-connecting-ip:   ), which is truthy in JS but meaningless as an IP. Without .trim(), an attacker bypassing Cloudflare could send spaces as the header value, causing the code to fall through to X-Forwarded-For (attacker-controlled) in production.

kasperpawlowski:
@Seranged could you update the docs too? we should clearly say now that the prod is expected to be behind cloudflare. you can say that it's needed both to enforce this and the geo-gate feature

kasperpawlowski (approved):

PR #173, opened 2026-04-08
feat: add oracle checks display to borrow pair oracle section

Seranged:
- Surface per-adapter check results (pass/fail, severity) from the oracle-checks metadata as a Checks column in the oracle adapters grid
- Green/yellow/red dot based on pass status and highest failure severity
- Hover tooltip via Teleport shows full check breakdown with icons
- Failed checks rendered inline below the grid for quick visibility
- Suppress internal 'Adapter whitelist' check per monorepo parity

Oracles

Overview

PYUSDUSD 0x2289 BIRE C


kasperpawlowski:
1. poor ux on mobile
image
2. I didn't notice that the oracle-checks contain something called "label". if it's defined, as an improvement, we can try to display it in place we currently display the raw base/quote? this will give user more info about the oracle adapter

Seranged:
1. lmk if this order is ok for 2by2
image

1b. I also added a draw for when you click on the checks 
image


2. Labels looks ok like this? (desktop/mobile renderers) 
image
image

kasperpawlowski (approved):

PR #181, opened 2026-04-09
fix: prevent RPC retry amplification on transport errors

kasperpawlowski:
## Summary
- When a batch RPC call failed (e.g. Cloudflare 403), `evcBatchCall` swallowed the error and returned `{ success: false }` for every item — indistinguishable from individual on-chain reverts. This caused `fetchBatch` to retry each item individually, amplifying 1 failed batch of 25 into 25+ separate requests against an already rate-limited endpoint, creating a cascading flood.
- Transport errors (HTTP failures, timeouts) are now distinguished from on-chain reverts using viem's error hierarchy (`HttpRequestError`, `TimeoutError`, `WebSocketRequestError`). Individual retries only fire for on-chain reverts. Transport errors skip retries and stop processing further batch chunks.
- `batchLensCalls` returns a new `BatchLensResult<T>` type with a `transportError` flag. Existing consumers (`useAccountPositions`, pyth utils) are unaffected — they already skip `success: false` items gracefully.

## Test plan
- [ ] Verify normal vault loading works (all vaults appear)
- [ ] Simulate 403 by throttling `/api/rpc/1` in devtools — confirm console is no longer flooded with individual retry errors
- [ ] Confirm vaults missing due to transient transport error appear after the 60s poll cycle

### DIFF
--- a/composables/useAccountPositions.ts
+++ b/composables/useAccountPositions.ts
@@ -145,11 +145,14 @@ const updateBorrowPositions = async (
       rpcUrl.value,
     )
 
+    let hasTransportError = false
     for (let i = 0; i < results.length; i++) {
-      if (results[i].success && results[i].result) {
+      if (results[i].transportError) hasTransportError = true
+      else if (results[i].success && results[i].result) {
         accountInfoMap.set(nonPythEntries[i].key, results[i].result!)
       }
     }
+    if (hasTransportError) logWarn('useAccountPositions/accountInfo', 'RPC transport error — account info results may be incomplete')
   }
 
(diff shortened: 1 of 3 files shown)


Seranged (commented):
Disabled the RPC on localhost and found another retry amplification bug — one failed batch was spawning 25 individual retries.

Root cause was in how we walked the viem error chain: we were drilling past the HttpRequestError all the way down to a raw browser TypeError, which wasn't in our known transport error list, so the retry guard never fired.

One-line fix to treat any non-viem root cause as a transport error. Verified before/after

Seranged (approved):

PR #183, opened 2026-04-09
fix: fetch Merkl Earn vault rewards independently of protocol tagging

kasperpawlowski:
## Summary

- Drop `mainProtocolId=euler` filter from the Merkl ERC20LOGPROCESSOR endpoint — Merkl does not reliably tag all Earn vault campaigns (e.g. Clearstar vaults had `protocol: N/A` and `action: HOLD` instead of `LEND`)
- Fetch all ERC20LOGPROCESSOR opportunities with pagination (`items=100`, `page=N`) and filter client-side against known Earn vault addresses from `euler-labels`
- Defer ERC20LOGPROCESSOR fetch until labels are loaded to avoid wasted requests; a watcher on `earnVaults` triggers a force-refresh once available
- Relax `Opportunity.action` type from `'LEND' | 'BORROW'` to `string` to accommodate `HOLD` action on some Earn vaults

## Test plan

- [x] Verify Earn vault rewards APY displays correctly on Monad (eeWETH, CSUSDC, CSAUSD vaults)
- [x] Verify EVK vault rewards (supply + borrow) are unaffected
- [x] Verify no duplicate Merkl API calls on initial load (ERC20LOGPROCESSOR only fetched after labels load)
- [x] Verify rewards refresh correctly on chain switch and wallet change

### DIFF
--- a/composables/useMerkl.ts
+++ b/composables/useMerkl.ts
@@ -161,7 +161,7 @@ const loadOpportunities = async (chainId: number, isInitialLoading = true, force
     // ERC20LOGPROCESSOR is fetched without mainProtocolId filter (unreliable tagging by Merkl)
     // and filtered client-side against known Earn vault addresses from labels.
     // Merkl API caps items at 100, so we paginate ERC20LOGPROCESSOR to get all results.
-    const fetchAllPages = async (baseUrl: string): Promise<Opportunity[]> => {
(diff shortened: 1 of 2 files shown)


Seranged (approved):
Fixed chainswapping recalling the endpoints in my commits

PR #195, opened 2026-04-11
feat: add back navigation to drill-down pages

Seranged:
## Summary
**- Several drill-down pages had no back navigation this forced the user to navigate back via the navbar, which resets their position back to baseline**

- Adds a new `BaseBackIconButton` component with history-aware navigation: prefers `router.back()` when browser history state exists (in-app navigation), falls back to an explicit route when the page is cold-loaded via direct URL or deep link
- The fallback prevents the back button from silently doing nothing or exiting the app when a user arrives at a position or vault action page via an external link or bookmark

## Changes
- **`BaseBackIconButton`**: uses `window.history.state?.back` (set by Vue Router on every in-app push) to distinguish in-app navigation from cold loads. Accepts `fallback` (default `/portfolio`) and `exact` props — `exact` skips the history check entirely and always navigates to the fallback, used on the position overview page where the destination should always be `/portfolio` regardless of history
- **`VaultForm`**: extended with optional `back` and `backFallback` props so action sub-pages can declare their back destination without needing to manage the button outside the form panel
- **Vault detail pages** (`lend/[vault]`, `earn/[vault]`, `borrow/[collateral]/[borrow]`): replaced mobile-only `BaseBackButton` with `BaseBackIconButton` rendered inline alongside `VaultLabelsAndAssets`. Known limitation: the inline placement shifts the vault header horizontally — a follow-up is needed to position the button without displacing the header (e.g. absolute positioning into the page gutter or a layout-level slot)
- **Position action sub-pages** (`borrow`, `borrow/swap`, `collateral/swap`, `multiply`, `repay`, `supply`, `withdraw`): back rendered inside `VaultForm` header, fallback to `/position/${positionIndex}`
- **Earn/lend sub-account pages** (`[subAccount]/withdraw`, `[subAccount]/swap`): same VaultForm pattern, fallback to the parent vault detail page. `supply.vue` and `withdraw.vue` were also missing `usePositionIndex()` — added

## Test plan
- [ ] Cold-load a position action page (e.g. `/position/1/borrow`) and verify back button navigates to `/position/1`
- [ ] Navigate in-app from position overview → borrow and verify back button returns to position overview via history
- [ ] Cold-load position overview (`/position/1`) and verify back button always goes to `/portfolio` (exact mode)
- [ ] Navigate in-app to a lend/earn/borrow vault detail page and verify back button returns to the list page
- [ ] Cold-load a vault detail page and verify back button falls back to `/lend`, `/earn`, or `/borrow`
- [ ] Verify no back button appears on mobile for vault detail pages (button is desktop-only on those pages)
- [ ] Verify lend/earn sub-account pages (withdraw, rebalance) show back button pointing to the parent vault

Desktop:

image

image

image

Mobile:

image

image

image


kasperpawlowski (changes requested):
apply the same pattern on pages/explore/[market].vue

Seranged:
> apply the same pattern on pages/explore/[market].vue

Pushed
https://github.com/euler-xyz/euler-lite/pull/195/changes/9472fa15cad93a347d3b77aa061e9b0553fbceb7

kasperpawlowski (approved):

PR #201, opened 2026-04-13
feat: link to previous app from header dropdown and migration modal

kasperpawlowski:
## Summary
- Adds opt-in `NUXT_PUBLIC_CONFIG_MIGRATION_OLD_APP_URL` (wired in `nuxt.config.ts` + `useDeployConfig.ts`). When set, a "Go to the previous app" link is rendered as the first item in the logo dropdown in `TheHeader.vue`. When unset, the dropdown is unchanged.
- Updates `MigrationAnnouncementModal.vue` to tell users they can still reach the previous app via that top-left link, and that it will no longer be maintained.

This is intentionally a temporary, transitional flag, so it is not added to `.env.example` — only consumers who need it (production deploys) configure it via Doppler.

## Test plan
- [x] With `NUXT_PUBLIC_CONFIG_MIGRATION_OLD_APP_URL` unset: header dropdown shows only Resources + socials (no migration row, no extra divider).
- [x] With the var set to a URL: dropdown shows "Go to the previous app" at the top, separated by a bottom border, opens in a new tab with `rel="noopener noreferrer"`.
- [x] With `NUXT_PUBLIC_CONFIG_MIGRATION_ANNOUNCEMENT_URL` set (modal trigger): the modal includes the new paragraph about the previous app.
- [x] Modal still dismisses correctly and persists `migration-announcement-seen` in localStorage.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -5,7 +5,7 @@ import { MigrationAnnouncementModal } from '#components'
 
 const route = useRoute()
 const router = useRouter()
-const { migrationAnnouncementUrl } = useDeployConfig()
+const { migrationAnnouncementUrl, migrationLegacyAppUrl } = useDeployConfig()
 const migrationAnnouncementSeen = useLocalStorage('migration-announcement-seen', false)
 const modal = useModal()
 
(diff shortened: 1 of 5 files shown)


Seranged (approved):
1. Tweaked dropdown styling a TINY bit (spacing)
2. Allowed the paragraph in the migration modal to be turned off when the ENV doesn't exist
3. Moved dropdown panel outside <button> to fix invalid nested interactive elements.

PR #221, opened 2026-04-14
fix: align hook warning icon with badges in earn exposure rows & sync with master

kasperpawlowski:
## Summary
- VaultWarningIcon in earn vault exposure rows was vertically centered against the full VaultLabelsAndAssets block, causing it to float between the name and asset rows
- Added a default slot to VaultLabelsAndAssets so extra inline content renders in the badges row (alongside Deprecated/Restricted badges)
- Moved the hook warning icon into that slot so it aligns on the same line as the vault name and badges

## Test plan
- [ ] Navigate to a deprecated earn vault with hooked strategies (e.g. Euler Arbitrum Yield on Arbitrum)
- [ ] Verify the hook warning icon sits inline with the "Deprecated" badge, not floating between rows
- [ ] Verify other pages using VaultLabelsAndAssets are unaffected (lend vault pages, borrow pages, portfolio)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed page navigation on market exploration pages when fetching market data.

* **Refactor**
  * Simplified vault label verification and initialization logic.
  * Improved earn vault loading and caching mechanism.
  * Enhanced vault header component structure for better content flexibility.

* **Documentation**
  * Updated vault label verification documentation.

### DIFF
--- a/components/entities/vault/VaultLabelsAndAssets.vue
+++ b/components/entities/vault/VaultLabelsAndAssets.vue
@@ -131,6 +131,7 @@ const displayAssetsLabel = computed(() => assetsLabel || assets.map(asset => ass
           />
           Restricted
         </span>
+        <slot />
       </div>
 
       <p class="text-p2 font-semibold text-content-primary">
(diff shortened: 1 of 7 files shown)


Seranged (approved):
Approved — not addressing the review comments as the timeout concern is pre-existing (not introduced by this PR), and the flash is not reproducible since resetVaultsState() sets all loading flags to true synchronously before any component renders.

PR #223, opened 2026-04-14
feat: CowSwap collateral swap and close position flows

kasperpawlowski:
## Summary

- Add CowSwap integration for **collateral swap** (`CowEvcCollateralSwapWrapper`) and **close position / repay from collateral** (`CowEvcClosePositionWrapper`) flows
- Rename open-position-specific symbols to clarify scope across three order types
- Generalize `CowSwapReviewModal` to accept flow-agnostic step props
- Thread `includeCowSwap` through quote infrastructure for collateral swap and repay-from-collateral (not savings, not wallet repay)
- Close position uses Inbox contract with EIP-1271 signing (domain: `name="Inbox"`, `version="1"`)

### Infrastructure
- New ABIs for `CollateralSwapWrapper` and `ClosePositionWrapper` (including `getInboxAddressAndDomainSeparator`)
- New types: `CowSwapCollateralSwapParams`, `CowSwapClosePositionParams`, and their execute param variants
- Generalized `buildCowSwapAppData` (appCode param), `buildCowSwapOrderTypedData` (kind, domain overrides), `buildCowSwapOrderPayload` (signingScheme)
- New `entities/cowswap/inbox.ts`: Inbox signature building, order struct encoding (384 bytes with keccak256-hashed string fields), domain separator verification
- Extracted `useCowSwapExecutionCore` shared composable; flow-specific composables for open position, collateral swap, close position

### Form integration
- Collateral swap page: CowSwap provider detection, `CowSwapCollateralSwapExecuteParams` construction, `disableSourceCollateral` auto-set on 100% swap
- Repay from collateral tab: CowSwap provider detection, `CowSwapClosePositionExecuteParams` with order kind (`buy` for full close, `sell` for partial), interest buffer on full repay
- Both: terminal status watcher for position refresh + portfolio redirect

### CowSwap eligibility
| Flow | CowSwap? | Reason |
|------|----------|--------|
| Multiply (open position) | Yes | Already implemented |
| Collateral swap | Yes | Cross-asset only |
| Repay from collateral | Yes | Cross-asset only |
| Repay from savings | No | Cross-subaccount — wrapper is single-subaccount |
| Repay from wallet | No | No matching wrapper |

## Test plan

- [ ] Verify existing multiply CowSwap flow still works after rename
- [ ] Collateral swap: CowSwap quote appears on Ethereum cross-asset, absent same-asset and other chains
- [ ] Collateral swap: 2 signature prompts (EVC permit + CoW order), order on barn explorer
- [ ] Close position: CowSwap quote appears in "From collateral" tab, absent in "From savings"/"From wallet"
- [ ] Close position: order uses Inbox as receiver, `signingScheme: eip1271`
- [ ] Full close: `kind=buy`, partial: `kind=sell`
- [ ] Cancel order works for collateral swap; test cancel behavior for close position (EIP-1271)
- [ ] Non-Ethereum chains: no CowSwap quotes in any flow


LeonardEulerXYZ (commented):
## Walkthrough

### Overview
This draft adds CowSwap collateral-swap and close-position flows. The architecture is headed in a sensible direction, but I found two blocking correctness gaps in execution behavior.

### Verdict
**Commented**

### Findings
1. **Potential issue · blocking** — the multiply flow can no-op when CoW is the best quote but the user has not manually selected a route card. The branch treats CoW as the active provider from the effective/best quote, but submission still reads only `multiplySelectedQuote` and returns early when it is null.
2. **Potential issue · blocking** — close-position orders are submitted as `eip1271` / Inbox-owned orders, but cancellation is hardcoded as an `eip712` EOA cancellation signed by the wallet. That cannot authenticate as the Inbox contract that owns the order, so the cancel action is effectively broken for this path.

### Suggested next step
Before this leaves draft, make the default-best CoW route actually submit, and align cancellation with the signing scheme / owner used for close-position orders.

PR #244, opened 2026-04-15
fix: clamp source-Max to debt in repay flow when source value exceeds debt

kasperpawlowski:
## Summary

Fixes accidental over-repay when clicking **Max** on the source input in the repay form. Each repay tab (wallet / collateral / savings) has two Max buttons — one for the source balance and one for the debt. Previously, if the wallet/collateral/savings balance was worth more than the debt, clicking source-Max set the amount to the full source balance and the form failed with *"You repaying more than required"*.

Now: when the source is worth at least as much as the debt, source-Max behaves like debt-Max — switches to `TARGET_DEBT` mode with the full debt amount, so the user fully repays without over-paying. When the source is worth less than the debt, behavior is unchanged (EXACT_IN with full source).

### What changed

- `AssetInput.vue` — added optional `maxHandler` prop; when provided, it replaces the default Max behavior.
- `useRepaySwapCore.ts` — added `onSourceMax()` shared by collateral + savings tabs. Compares against already-reactive `sourceValueUsd` / `borrowValueUsd` for cross-asset cases; native-unit comparison for same-asset.
- `useWalletRepay.ts` — added `onSourceMax()` for wallet no-swap (same asset).
- `useWalletSwapRepay.ts` — added `onSourceMax()` + two new reactive USD refs that use `getTokenUsdValue` to price arbitrary wallet tokens (including tokens without a vault, via the backend price feed).
- `useCollateralSwapRepay.ts` / `useSavingsRepay.ts` — re-exported the handler from the core, and upgraded the `isSubmitDisabled` balance check to use `getSwapInputAmount(quote, direction)` so it compares against the slippage-padded `amountInMax` rather than the base `amountIn`. This closes a pre-existing gap where a tight slippage budget in TARGET_DEBT could silently pass submit-disabled and revert on-chain.
- `priceProvider.ts` — promoted `getTokenUsdValue` (vault-or-backend fallback) from inside `useSwapPriceImpact.ts` to a shared export so both the price-impact composable and the wallet-swap flow share one helper.
- `repay.vue` — wired `:max-handler` on the four source inputs (wallet no-swap, wallet-swap, collateral, savings). Debt inputs and percent slider are unchanged.

### Slippage edge case

If the oracle says `sourceValueUsd > borrowValueUsd` but slippage-padded `amountInMax` exceeds the source balance, the switched-to-TARGET_DEBT mode now surfaces *"Insufficient collateral/savings balance to cover the required swap amount"* in the form — clear feedback instead of a silent on-chain revert. The user can then adjust slippage or manually switch back to a partial repay.

If USD refs haven't resolved yet when the user clicks Max, behavior gracefully falls back to EXACT_IN with full source (the pre-existing behavior).

### Scope

- Max-button only. Manual typing of an over-repay amount still shows the existing *"repaying more than required"* error; manual typing is explicit intent.
- The same-asset clamp inside the debounced `requestQuote` (useRepaySwapCore.ts) is left as a safety net for manual entries.

## Test plan

- [x] **Wallet no-swap, balance < debt** — source-Max sets amount to walletBalance, submit enabled, partial repay.
- [x] **Wallet no-swap, balance > debt** — source-Max sets amount to debt, percent slider jumps to 100, submit enabled, full-repay plan selected. (Previously: error.)
- [x] **Wallet swap, source USD < debt USD** — source-Max → EXACT_IN with walletBalance.
- [x] **Wallet swap, source USD > debt USD, slippage fits** — source-Max → TARGET_DEBT with full debt, quote resolves, submit enabled.
- [ ] **Wallet swap, source USD > debt USD, slippage too tight** — "Not enough balance" error shown (already wired in `updateSyncEstimates`).
- [x] **Collateral same-asset, balance > debt** — source-Max → TARGET_DEBT, percent=100, `buildSameAssetFullRepayPlan` is selected.
- [x] **Collateral cross-asset, source USD > debt USD, slippage fits** — source-Max → TARGET_DEBT full debt, quote fills, submit enabled.
- [ ] **Collateral cross-asset, source USD > debt USD, slippage tight** — "Insufficient collateral balance…" error shown (new check).
- [x] **Savings tab** — same scenarios as collateral.
- [x] **Debt-Max button still works** on all tabs (regression check).
- [x] **Percent slider still works** on all tabs (regression check).
- [x] **Manual typing over-debt in source** still shows "repaying more than required" (regression check).

## Summary by CodeRabbit

* **New Features**
  * Added "set max amount" functionality to all repay flows. Users can now quickly set repayment amounts to their maximum available balance across wallet direct repay, cross-asset swaps, collateral swaps, and savings repay options with a single click.

### DIFF
--- a/components/entities/asset/AssetInput.vue
+++ b/components/entities/asset/AssetInput.vue
@@ -22,6 +22,7 @@ const props = defineProps<{
   priceOverride?: number // USD unit price for assets without a vault (e.g., swap-to-deposit)
   swappable?: boolean // When true, asset pill shows dropdown arrow and emits click-asset
   selectedSource?: 'wallet' | 'saving' // Source indicator chip when multiple collateral options exist
+  maxHandler?: () => void // When provided, replaces the default "Max" button behavior
 }>()
 const emits = defineEmits(['input', 'change-collateral', 'click-asset'])
 const model = defineModel<string>({ default: '' })
@@ -99,6 +100,10 @@ const price = computed(() => {
 
 const hasPrice = computed(() => price.value !== null)
 const setMax = () => {
+  if (props.maxHandler) {
+    props.maxHandler()
+    return
+  }
   model.value = trimTrailingZeros(formatUnits(props.balance ?? 0n, Number(props.asset.decimals)))
   emitInputNow()
   if (inputEl.value) {
--- a/composables/repay/useCollateralSwapRepay.ts
+++ b/composables/repay/useCollateralSwapRepay.ts
@@ -16,6 +16,7 @@ import { useEulerProductOfVault } from '~/composables/useEulerLabels'
 import { useRepaySwapCore } from '~/composables/repay/useRepaySwapCore'
 import { useRepaySwapDetails } from '~/composables/repay/useRepaySwapDetails'
 import { useRepayHealthMetrics } from '~/composables/repay/useRepayHealthMetrics'
+import { getSwapInputAmount } from '~/composables/useEulerOperations/swaps/verify'
 import { nanoToValue, valueToNano } from '~/utils/crypto-utils'
 import { normalizeAddressOrEmpty } from '~/utils/accountPositionHelpers'
 import { createRaceGuard } from '~/utils/race-guard'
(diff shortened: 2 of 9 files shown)


LeonardEulerXYZ (changes requested):
## Walkthrough

### Overview
The general approach makes sense: clamp source-side Max to debt when the source is already worth more than the debt, instead of blindly filling the whole wallet balance. But there is still a hole in the wallet-swap path.

### Verdict
**Changes requested**

### Findings
1. **Potential issue · blocking** — native-token wallet swaps can still bypass the clamp.
   - In `useWalletSwapRepay`, the new comparison path prices the source via `getTokenUsdValue(selectedAsset.value.address, null)`.
   - For native inputs, that raw address is not normalized to the wrapped native token before backend pricing, so the price lookup can come back `undefined`.
   - When that happens, `onSourceMax()` falls back to `EXACT_IN` with the full wallet balance, which is exactly the over-repay behavior this PR is trying to prevent.

### Suggested next step
Normalize native input to the wrapped native address before the USD lookup/comparison.

Seranged (approved):
Did lots of manual testing

PR #250, opened 2026-04-16
Fix/slippage edit accessibility

Seranged:
## Summary
- When a swap quote failed, the slippage tolerance edit button disappeared because `SwapDetailsSummary` was gated behind a quote-success condition (`swapEstimatedOutput`). Users had no way to adjust slippage to retry the quote unless they navigated to the global Settings panel.
- The block now renders on quote success (full details) or quote failure (slippage row only), while staying hidden during loading — preserving the original UX for the happy path.

## Changes
- Widened the `v-if` on 6 pages from `swapEstimatedOutput` to `swapEstimatedOutput || swapQuoteError` so the swap details block also appears when a quote fails.
- Updated `SwapDetailsSummary` to hide data-dependent rows (swap in/out, price impact, routed via) when their values are empty. The slippage tolerance row always renders, so on quote failure only the editable slippage is shown instead of a card full of dashes.
- Affected pages: borrow, position supply, position withdraw, position repay (wallet tab), lend deposit, lend withdraw.

## Test plan
- [ ] On any swap-enabled page (e.g. borrow with "Pay with" a different token), trigger a quote failure (select a token pair with no swap route available)
- [ ] Verify the slippage tolerance row with its edit button appears after the quote fails
- [ ] Verify clicking the edit button opens the slippage settings modal and changes persist
- [ ] Verify the block does NOT flash during quote loading — it should appear only after loading completes
- [ ] Verify successful quotes still show all detail rows (swap in/out, price impact, slippage, routed via)
- [ ] Verify the multiply page and dedicated swap pages (which were not affected) still behave normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed empty placeholder values from swap detail summaries for cleaner display
  * Swap information containers now display when quote errors occur, providing better error context alongside messages

### DIFF
--- a/components/entities/swap/SwapDetailsSummary.vue
+++ b/components/entities/swap/SwapDetailsSummary.vue
@@ -17,22 +17,31 @@ const emit = defineEmits<{
 </script>
 
 <template>
-  <SummaryRow label="Swap in">
+  <SummaryRow
+    v-if="inputDisplay"
+    label="Swap in"
+  >
     <p class="text-p2 text-right">
-      {{ inputDisplay ?? '-' }}
+      {{ inputDisplay }}
     </p>
   </SummaryRow>
-  <SummaryRow label="Swap out">
+  <SummaryRow
+    v-if="outputDisplay"
+    label="Swap out"
+  >
     <p class="text-p2 text-right">
-      {{ outputDisplay ?? '-' }}
+      {{ outputDisplay }}
     </p>
   </SummaryRow>
-  <SummaryRow label="Price impact">
+  <SummaryRow
+    v-if="priceImpact !== null"
+    label="Price impact"
+  >
     <p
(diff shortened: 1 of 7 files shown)


LeonardEulerXYZ (dismissed):
## Walkthrough

### Overview
This keeps the slippage control reachable when quote generation fails by rendering the swap details summary on quote error as well as on quote success, while hiding empty rows.

### Verdict
**Approved**

### Findings
No blocking findings. The changed entry points and `SwapDetailsSummary` behavior look internally consistent.

### Note
I could not complete local frontend validation in this VM because the repo currently expects Node 20+ and the lockfile is out of sync, so this is a grounded diff review rather than a successful local build.

kasperpawlowski (approved):

PR #252, opened 2026-04-16
Feat/exact amount hover tooltip

Seranged:
## Summary
- Restore the ability to see full-precision token amounts (all decimal places) by hovering over abbreviated values — a feature from the legacy app that was missing
- Desktop-only: tooltips appear on mouse hover and are completely inert on touch devices

## Changes
- New `UiExactAmount` component using pure CSS `@media (hover: hover) and (pointer: fine)` for reliable desktop-only detection — no JS event listeners, so no iOS Safari double-tap or sticky-hover issues on touch devices
- New `formatExactAmount` utility that does pure bigint arithmetic to format amounts with full decimal precision, avoiding JS `Number` precision loss (which caps at ~15-17 significant digits, not enough for 18-decimal tokens like WETH)
- Applied to: position detail page (borrow/collateral market values), portfolio items (supply value, debt, collateral), vault exposure blocks (allocation amounts/caps), asset input balances, and withdraw page balances

## Test plan
- [ ] Desktop: hover over any approximate token amount (e.g., "~ 1.23 WETH") on the portfolio page — tooltip should show the full-precision value
- [ ] Desktop: hover over token amounts on the position detail page (borrow market value, collateral market value)
- [ ] Desktop: hover over wallet balance in any supply/withdraw form's asset input
- [ ] Desktop: hover over exposure allocation amounts in earn vault overview
- [ ] Mobile: tap on any of the above amounts — no tooltip should appear, no double-tap required for navigation
- [ ] iPad: same as mobile — touch interactions should be completely unaffected

image

## Summary by CodeRabbit

* **New Features**
  * Exact-amount tooltips: hoverable displays show full-precision token amounts (including decimals) across portfolio items, vaults, asset inputs and withdraw/earn dialogs, with a copy-to-clipboard action for the exact value.

* **Improvements**
  * Swap info visibility: swap detail cards now also appear when quote retrieval fails, improving visibility into error states and guidance.

### DIFF
--- a/components/entities/asset/AssetInput.vue
+++ b/components/entities/asset/AssetInput.vue
@@ -3,7 +3,7 @@ import { formatUnits } from 'viem'
 import type { Vault, SecuritizeVault, VaultAsset, CollateralOption, EarnVault } from '~/entities/vault'
 import { getAssetUsdPrice } from '~/services/pricing/priceProvider'
 import { nanoToValue } from '~/utils/crypto-utils'
-import { compactNumber, formatSmartAmount, trimTrailingZeros } from '~/utils/string-utils'
+import { compactNumber, formatSmartAmount, trimTrailingZeros, formatExactAmount } from '~/utils/string-utils'
 import { ChooseCollateralModal } from '#components'
 import { useModal } from '~/components/ui/composables/useModal'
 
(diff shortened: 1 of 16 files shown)


LeonardEulerXYZ (dismissed):
## Walkthrough

### Overview
This mostly does two things: add desktop hover tooltips for exact token amounts, and keep `SwapDetailsSummary` visible on quote errors so slippage remains editable. The swap-summary side looks reasonable, but I found one blocking tooltip bug.

### Verdict
**Changes requested**

### Findings
1. **Potential issue · blocking** — `components/entities/portfolio/PortfolioBorrowItem.vue` now shows an “exact” tooltip for aggregated collateral positions using only `collateralItems[0]`.
   - The visible label can say `SYMBOL & others`, but the tooltip value is still just the first collateral.
   - That makes the “exact amount” tooltip misleading for multi-collateral borrow positions.

### Suggested next step
Either compute an exact representation for the full basket, or suppress the exact tooltip on aggregate multi-collateral rows.

kasperpawlowski (changes requested):
make it in a way that users are able to copy the value from the tooltip (apparently this is what they often do)

kasperpawlowski (dismissed):

kasperpawlowski (approved):

PR #253, opened 2026-04-16
fix: add spy mode support to reward composables

kasperpawlowski:
## Summary
- All four reward composables (`useREULLocks`, `useMerkl`, `useBrevis`, `useFuul`) only fetched data for the connected wallet address, ignoring spy mode (`?spy=` query param) entirely
- Portfolio rewards page showed no/wrong data when viewing another address via spy mode
- Introduces `effectiveAddress` (`spyAddress || wagmiAddress`) for data reads and `isActive` (`isConnected || spyAddress`) as the lifecycle trigger, preserving the original boolean-toggling watcher pattern
- Write operations (claim/unlock) still require the connected wallet to sign

## Test plan
- [ ] Open `/portfolio/rewards?network=1&spy=<address>` without connecting a wallet — all reward sections should load data for the spy address
- [ ] Connect a wallet while in spy mode — rewards should still show spy address data (spy takes priority)
- [ ] Clear spy mode while connected — rewards should switch to the connected wallet's data
- [ ] Verify no excessive RPC/API polling in the network tab (same cadence as normal connected mode)
- [ ] Verify claim/unlock buttons are disabled or require wallet connection in spy mode

### DIFF
--- a/composables/useBrevis.ts
+++ b/composables/useBrevis.ts
@@ -82,6 +82,10 @@ export const useBrevis = () => {
   const { chainId } = useEulerAddresses()
   const { client: rpcClient } = useRpcClient()
   const { enableIncentra } = useDeployConfig()
+  const { spyAddress } = useSpyMode()
+
+  const effectiveAddress = computed(() => spyAddress.value || wagmiAddress.value || '')
+  const isActive = computed(() => isConnected.value || Boolean(spyAddress.value))
 
   const ensureWalletOnCurrentChain = async () => {
     const targetChainId = chainId.value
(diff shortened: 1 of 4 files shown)


LeonardEulerXYZ (dismissed):
## Walkthrough

### Overview
Adds spy-mode support to the rewards composables by deriving an effective address from `spyAddress || wagmiAddress` and treating spy mode as an active read-only context.

### Verdict
**Approved**

### Findings
No blocking findings. The read paths now follow the effective address, while write paths still remain wallet-gated, which is the right separation.

Seranged (approved):

PR #254, opened 2026-04-16
fix: preserve network query param on post-transaction redirects

Seranged:
All post-transaction redirects used bare path strings without the network query parameter, triggering the network middleware to fire a second redirect. This double-navigation race condition with keep-alive and the out-in page transition could leave the portfolio child page empty and select the wrong tab.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network persistence after transactions. The selected network parameter is now preserved when users are redirected following transaction completion on earn, lend, and position pages, providing a more consistent user experience.

### DIFF
--- a/composables/useSwapPageLogic.ts
+++ b/composables/useSwapPageLogic.ts
@@ -84,6 +84,7 @@ export const useSwapPageLogic = (options: UseSwapPageLogicOptions) => {
   const otherAmountField: SwapQuoteAmountField = displayAmountField === 'amountIn' ? 'amountOut' : 'amountIn'
 
   const router = useRouter()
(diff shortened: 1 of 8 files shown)


LeonardEulerXYZ (dismissed):
## Walkthrough

### Overview
Small, targeted fix to preserve the `network` query param on post-transaction redirects that previously navigated to bare portfolio routes.

### Verdict
**Approved**

### Findings
No blocking findings. The redirect updates are applied consistently across the touched success paths, and the added `useRoute()` usage in the composable is appropriate.

kasperpawlowski (approved):

PR #258, opened 2026-04-16
feat: public /api/public/is-known vault verification endpoint

kasperpawlowski:
## Summary

- Adds a new public endpoint `GET /api/public/is-known?chainId=X&addresses=0x…,0x…` that returns `{ [checksumAddress]: boolean }` indicating whether each vault is verified by this app.
- "Verified" merges three sources per chain: `products.json` active vaults (EVK + Securitize), `earn-vaults.json` entries where `deprecated !== true`, and on-chain `escrowedCollateralPerspective.verifiedArray()`. Deprecated vaults are excluded; `notExplorable` vaults stay verified.
- New `/api/public/` path prefix: the CORS middleware sets `Access-Control-Allow-Origin: *` for any file under it. Other endpoints keep the strict allowlist.
- Labels proxy refactor: all four files (`products.json`, `entities.json`, `earn-vaults.json`, `points.json`) are now optional. Missing or unreachable files return a type-appropriate empty payload (`{}` or `[]`) with HTTP 200 instead of 502, so new chains and partial deployments degrade gracefully.

## Implementation

- `server/api/public/is-known.get.ts` — handler with address validation (viem `isAddress`/`getAddress`), 100-address batch cap, rate limit 100/min per IP, chainId allowlist (must have `RPC_URL_HTTP_\<chainId\>` configured).
- `server/utils/verified-vaults.ts` — per-chain verified-set builder with 5-min TTL cache, stale-while-revalidate, and inflight request deduplication. Escrow addresses fetched via a direct eth_call to the configured RPC (perspective address from `EulerChains.json`); labels fetched via internal `$fetch` to reuse the proxy's cache and validation.
- `server/middleware/cors.ts` — public-path branch added ahead of the allowlist check.
- `server/api/labels/[file].get.ts` — removed the required/optional split; `EMPTY_SHAPES` table drives both whitelist validation and fallback payloads.

## DDoS posture

- Per-IP rate limit: 100/60s via the existing `createRateLimiter` (fail-closed on missing `CF-Connecting-IP` in prod).
- chainId must be in the deployment's RPC allowlist (`resolveRpcUrl`) so bogus IDs 400 before any upstream work.
- Per-chain in-memory cache (5-min TTL, max 64 chains) and inflight dedup prevent rebuild stampedes; cold lookups cost exactly one rebuild across concurrent callers.
- Max 100 addresses per request; response bounded to ~5 KB.

## Docs

- New `docs/public-api.md` covering request/response shape, error codes, CORS, caching, rate limit, and curl/JS examples.
- `docs/README.md` index entry added.
- `docs/vault-labels-and-verification.md` cross-references the public endpoint and updates the label-files table to reflect that all files are now optional.

## Test plan

- [x] `curl 'http://localhost:3000/api/public/is-known?chainId=1&addresses=<active-evk>,<deprecated-evk>,<active-earn>,<deprecated-earn>,<escrow>,0x000…01'` returns `true,false,true,false,true,false`
- [x] Batch of 101 addresses returns 400 "Too many addresses"
- [x] Unsupported chainId (e.g. `?chainId=99999`) returns 400 "Unsupported chainId"
- [x] Malformed address returns 400 "Invalid address: 0x…"
- [x] 101st request within 60s from same IP returns 429
- [x] OPTIONS preflight from foreign origin returns 204 with `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Methods: GET, OPTIONS`
- [x] `/api/labels/earn-vaults.json?chainId=130` returns `200 []` (chain without earn labels)
- [x] `/api/labels/products.json?chainId=999999` returns `200 {}` (bogus chain)
- [x] `/api/labels/secrets.json?chainId=1` still 400s (whitelist enforced)
- [x] Existing lend/borrow/explore flows still load correctly on chain 1

## Summary by CodeRabbit

* **New Features**
  * Public API endpoint for batched verified-address lookup (up to 100 addresses), with permissive CORS, OPTIONS preflight support, per-client rate limiting, strict validation/normalization, and boolean mapping responses.

* **Documentation**
  * Added comprehensive public API docs and README section with endpoint specs, examples, validation rules, caching, and error semantics.

* **Bug Fixes**
  * Label endpoints now return sensible empty payloads for missing/upstream errors and use short caching; verification lookups use a 5-minute cache with stale-while-revalidate.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -81,6 +81,12 @@ Welcome to the documentation for the Euler Lite project. This documentation is d
 - Country group aliases (EU, EEA, EFTA)
 - UI enforcement across browse, detail, action, and modal pages
 
+### 🌐 [Public API](./public-api.md)
+
+- Publicly reachable endpoints under `/api/public/` (CORS `*`)
+- `GET /api/public/is-known` — verified-vault lookup by address
+- Request/response shape, caching, rate limits, and examples
+
 ## 🎯 Project Overview
 
 **Euler Lite** is a lightweight multi-chain DeFi application that provides lending and borrowing services through the Euler Finance protocol. It supports multiple EVM chains and connects via standard EVM wallets. The application allows users to:
--- a/docs/public-api.md
+++ b/docs/public-api.md
@@ -27,7 +27,7 @@ See [vault-labels-and-verification.md](./vault-labels-and-verification.md) for h
 | Param       | Type     | Required | Description                                                                 |
 |-------------|----------|----------|-----------------------------------------------------------------------------|
 | `chainId`   | integer  | yes      | EVM chain ID (e.g. `1` for mainnet).                                        |
-| `addresses` | string   | yes      | Comma-separated list of vault addresses. Max 100. Any case accepted.        |
+| `addresses` | string   | yes      | Comma-separated list of vault addresses. Max 100. Lowercase, uppercase, or valid EIP-55 checksum accepted; mixed-case inputs must have a correct checksum. |
 
 ### Response
 
@@ -40,7 +40,7 @@ Status `200`, JSON object mapping each input address (in EIP-55 checksum form) t
 }
 ```
 
-Addresses are normalized via EIP-55 checksum before comparison, so callers may submit lowercase, uppercase, or mixed-case and get the same answer. Keys in the response always use the checksum form.
+Addresses are validated via viem's `isAddress` (strict EIP-55 checks) and normalized to checksum form before comparison. Lowercase and uppercase hex inputs are accepted; mixed-case inputs must already carry a valid EIP-55 checksum or the request is rejected with `400`. Keys in the response always use the checksum form.
 
 ### Errors
 
--- a/docs/public-api.md
+++ b/docs/public-api.md
@@ -1,0 +1,107 @@
+# Public API
+
+Endpoints under `/api/public/` are intentionally reachable from any origin (no CORS allowlist) and are intended to be consumed by external integrators.
+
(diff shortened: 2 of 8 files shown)


LeonardEulerXYZ (changes requested):
## Walkthrough

### Overview
A public `is-known` endpoint is useful, but the current semantics are not yet aligned with the app’s own known-vault behavior and failure handling.

### Verdict
**Changes requested**

### Findings
1. **Potential issue · blocking** — deprecated vaults are currently reported as unknown.
   - `collectProductsAddresses()` skips `deprecatedVaults`, and `collectEarnAddresses()` skips deprecated earn entries.
   - The client-side label logic still treats those deprecated vaults as known for the unverified-vault flows, so the new endpoint diverges from the app’s current behavior.
2. **Potential issue · blocking** — upstream label failures now degrade to empty payloads, which can silently turn backend outages into `false` results.
   - The new fallback behavior in `server/api/labels/[file].get.ts` returns `{}`/`[]` for broad fetch/parse/validation failures, not just true file absence.
   - That means `/api/public/is-known` can return `200 { addr: false }` during upstream outages instead of surfacing a backend failure.

### Suggested next step
Align “known” semantics with the current product behavior, and distinguish genuine absence from upstream failure so the endpoint does not fail open with false negatives.

Seranged (approved):
Few tweaks:
 
- Labels endpoint: products.json and entities.json return 502 on genuine
  upstream failures (5xx, timeouts) instead of silently degrading to empty;
  404/403 (absent) still degrade gracefully for all files
- Docs: correct address validation description (strict isAddress rejects
  all-uppercase inputs)
- Reuse fetchWithTimeout with optional RequestInit in verified-vaults
  instead of reimplementing the AbortController pattern

PR #295, opened 2026-04-22
fix: tag-based stage deploy and harden deployment workflows

Seranged:
## Summary

- Stage deployments now trigger on `v*-rc*`/`v*-beta*` tags (e.g. `v1.0.0-rc.1`) instead of every push to `development`, matching the tag-based pattern used for production
- Prod `v*` pattern explicitly excludes `-rc`/`-beta` suffixes so both workflows never fire for the same tag
- Adds branch ancestry verification to ensure tagged commits are actually on `development`
- Adds `workflow_dispatch` with an `image_tag` input for deploy-only reruns of an existing image
- **Security:** Fixes script injection vector in both prod and stage workflows — `inputs.image_tag` was interpolated directly into shell scripts via `${{ }}`, now routed through env vars so values are never parsed as shell syntax
- **Security:** Restricts `workflow_dispatch` to target branch only — stage must be dispatched from `development`, prod from `master`. Prevents accidental deploys from feature branches
- **Security:** Validates image tag format against `[a-zA-Z0-9_.-]+` before writing to workflow outputs. Rejects newlines, spaces, or special characters that could corrupt step outputs

## Summary by CodeRabbit

* **New Features**
  * Manual deployment trigger with a customizable image tag input (default "latest").

* **Improvements**
  * CI now runs builds for tagged releases and supports prerelease tags for non-prod; production excludes prerelease tags.
  * Deploy flow gated so manual dispatches and pushes behave appropriately; build runs only on push.
  * Resolved image tag is selected and validated before deployment.
  * Full-history checkout and commit ancestry verification added.

### DIFF
--- a/.github/workflows/build-push-deploy-stage.yaml
+++ b/.github/workflows/build-push-deploy-stage.yaml
@@ -128,10 +128,15 @@ jobs:
           BUILD_TAG: ${{ needs.build-push.outputs.image_tag }}
         run: |
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
-            echo "image_tag=$INPUT_TAG" >> $GITHUB_OUTPUT
+            TAG="$INPUT_TAG"
           else
-            echo "image_tag=$BUILD_TAG" >> $GITHUB_OUTPUT
+            TAG="$BUILD_TAG"
           fi
+          if ! echo "$TAG" | grep -qE '^[a-zA-Z0-9_.-]+$'; then
+            echo "::error::Invalid image tag: must match [a-zA-Z0-9_.-]+"
+            exit 1
+          fi
+          echo "image_tag=$TAG" >> $GITHUB_OUTPUT
 
       - name: Configure AWS credentials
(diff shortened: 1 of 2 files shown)


girts512 (approved):

.github/workflows/build-push-deploy-stage.yaml
```diff
@@ -1,9 +1,16 @@
 name: Build, Push & Deploy (stage)
 on:
   workflow_dispatch:
+    inputs:
+      image_tag:
+        description: 'ECR image tag to deploy (defaults to "latest")'
+        required: false
+        default: 'latest'
```
girts512: default is set to 'latest' but 'latest' tag is pushed from build-push-deploy-prod action only, not from stage

girts512 (dismissed):

.github/workflows/build-push-deploy-stage.yaml
Seranged: > Resolved in https://github.com/euler-xyz/euler-lite/pull/295/changes/46d3e4bae7da11f799555d49663c474ab1c300af

girts512 (approved):

PR #302, opened 2026-04-23
Add asset-level geo-blocking via assets.json

kasperpawlowski:
## Summary
- Geo-blocking in labels has only supported vault-level `block` / `restricted` rules, so maintainers expressed "asset X is blocked in country C" by listing every vault of X. That proxy breaks for the pay-with (repay), withdraw-to, borrow swap-in, and swap-deposit flows, where the user picks an arbitrary ERC-20 in `SwapTokenSelector` rather than a vault — the existing checks inspect the borrow/collateral vault and never see the selected asset.
- This PR introduces a first-class asset-level rule source (`assets.json`) and OR-composes it with the existing vault-level rules, so asset-level policy becomes a floor that vault-level entries can only add to. Existing behaviour is preserved; new flows are now sealed end-to-end.
- A follow-up extends this with **pattern rules** (`symbols` / `symbolRegex` / `names` / `nameRegex`) and a cross-chain **`all/assets.json`** so issuer-wide policy (e.g. every Ondo-issued token) no longer needs an exhaustive per-chain address list, and patterns apply to the external token lists (Uniswap, DefiLlama, Merkl) the picker pulls from.

### Phase 1 — address-keyed asset rules
- **New labels file**: `assets.json` (chain-scoped like the rest), entries shaped `{ address, block?, restricted? }` with the existing `EU`/`EEA`/`EFTA` aliases via `COUNTRY_GROUPS`. Whitelisted in `server/api/labels/[file].get.ts` and loaded alongside the other files in `useEulerLabels.loadLabels`, stored in new `assetBlocks` / `assetRestrictions` maps (lowercased address keys, matching the earn-vault pattern).
- **Core helpers** (`composables/useGeoBlock.ts`): new `isAssetBlockedByCountry` / `isAssetRestrictedByCountry`. `isVaultBlockedByCountry` / `isVaultRestrictedByCountry` now OR in the asset-level rules by resolving the underlying asset via `useVaultRegistry().getVault(addr)?.asset` (registry accessor pulled once at module scope to avoid per-call allocations on hot paths like browse tables). Every existing call site — discovery chips, pair overview, portfolio items — picks up asset-level rules automatically.
- **Flow enforcement**, each using the relevant soft-restrict semantics ("restricted = can't *acquire* exposure; can still *reduce*"):
  - Repay pay-with (`useWalletSwapRepay` / `repay.vue`): hard-block only (source is given up).
  - Withdraw-to (`useCollateralForm` / `withdraw.vue`): hard + soft block (output is acquired).
  - Supply swap-deposit source (`lend/[vault]/index.vue`, `position/[number]/supply.vue`): hard-block only (source is given up).
  - Borrow swap-in source (`useBorrowForm` / `borrow` page): hard-block only (source is given up).
- **Selector UI** (`SwapTokenSelector.vue`): `mode='input'` disables hard-blocked assets; `mode='output'` disables both hard-blocked and soft-restricted. Disabled rows get `opacity-50 cursor-not-allowed` and a "Restricted" chip; click handlers suppress selection. Same treatment on the custom-token import row. A `geoByAddress` computed map evaluates the geo state once per row to avoid 3× re-computation per render.
- **Migration posture**: OR semantics mean labels maintainers can do a clean cutover — move asset-driven entries from `products.json` / `vaultOverrides[*]` / `earn-vaults.json` into `assets.json` and remove them from the vault-level spots. Genuinely vault-specific blocks stay where they are. No transition flag needed; labels are version-controlled.

### Phase 2 — pattern rules + cross-chain `all/assets.json`
- **Extended schema**: `EulerLabelAssetEntry` gains `symbols` (exact, case-insensitive), `symbolRegex`, `names` (exact, case-insensitive), `nameRegex`. Match fields are OR-composed within an entry. Regexes are compiled with the `i` flag at load time; invalid regexes are dropped with a warning.
- **Cross-chain file**: a new `all/assets.json` upstream hosts rules that apply on every chain. `refreshLabelFile` is refactored to `scope: number | 'all'`; the `/api/labels/assets.json?chainId=N` handler transparently unions per-chain + `all/` so the client fetches a single merged array. `warm-cache.ts` warms the global file once on boot alongside the per-chain entries.
- **Helper signature widened**: `isAssetBlockedByCountry` / `isAssetRestrictedByCountry` now accept an `AssetLike` (`string | { address?, symbol?, name? } | undefined`). String-only callers still work (address-map only); callers with the full asset (selector, form-level flags, vault-level OR via registry lookup) unlock pattern matching. All existing call sites updated to pass the full asset.

### Phase 3 — review fixes + labels sanitization
Kicked off by the deep-review orchestrator and a separate sanitization audit.

- **ReDoS mitigation**: cap the haystack passed to `.test()` at 128 chars (`MAX_REGEX_INPUT_LEN`). Real ERC-20 symbols are ≤12 chars and names ≤64; an attacker-chosen on-chain symbol can't drag a curator-typo regex into catastrophic backtracking.
- **Memoization**: per-country `assetBlockCache` / `assetRestrictedCache` in `useGeoBlock` keyed by `country|address|symbol|name`, size-bounded at 1000, cleared from `loadLabels` so decisions computed against removed rules are never served.
- **Scope fix**: `isInputAssetBlocked` in `useCollateralForm` now applies only in `mode === 'supply'`. In withdraw the input asset is the collateral vault's own underlying (covered by `isGeoBlocked` via the vault-level OR); the previous wiring disabled submit without a corresponding toast.
- **Mixed-entry warning**: loader logs a warning when an `assets.json` entry mixes `address` with pattern fields, since they share one `block` / `restricted` pair with no way to scope different country lists to each match surface.
- **Labels sanitization** in `validateNode`: `MAX_STRING_LEN` (16KB/string), `MAX_ARRAY_LEN` (10000/array), and a `SAFE_LOGO_FILENAME_RE` allowlist (alphanumeric + common image extensions; no slashes/traversal/protocol-relative). Logo values are concatenated into `${CDN}/${logo}` and `/entities/${logo}` in several components; the allowlist prevents a curator slip or compromised labels from pointing the browser off-origin.

Existing defenses kept as-is: `autoLink` HTML-escapes every string branch (so XSS via label text is impossible regardless of `validateNode`), URL scheme allowlist, markdown-link-injection detection, regex compile check + 512-char cap.

## Test plan
- [x] Set `DEV_GEO_COUNTRY=DE` (or another real test value) in `.env`, run `npm run dev`. Confirm startup `HEAD /` populates the country ref.
- [x] Stage a local `assets.json` via a forked labels repo (`NUXT_PUBLIC_CONFIG_LABELS_REPO`) that hard-blocks USDC in `DE` and soft-restricts WETH, and a local `all/assets.json` with `[{ "symbols": ["USDY","OUSG"], "block": ["DE"] }, { "symbolRegex": "^OMPL$", "restricted": ["DE"] }, { "nameRegex": "^Ondo\\s", "block": ["DE"] }]`.
- [x] **Server-side union**: `curl "http://localhost:3000/api/labels/assets.json?chainId=1"` returns per-chain + global entries concatenated.
- [x] **Repay pay-with**: open non-USDC debt. Picking USDC disables submit with "Paying with this asset is not available in your region". Picking WETH stays enabled (soft-restrict doesn't apply). Selector row shows dimmed + Restricted chip.
- [x] **Withdraw-to**: pick USDC as output → submit disabled ("Receiving this asset…"). Pick WETH → also disabled (soft-restrict on acquire). Both show dimmed + chip in the selector.
- [x] **Swap-deposit source (lend page)**: on `/lend/<vault>` where the target asset ≠ USDC, USDC as source disables submit; WETH stays enabled.
- [x] **Borrow swap-in**: USDC as pay-with on the borrow form's borrow tab disables submit with the "Paying with…" toast; WETH stays enabled.
- [x] **Position supply (add collateral)**: USDC as pay-with on `/position/<n>/supply` with non-USDC collateral disables submit with the "Paying with…" toast.
- [x] **Pattern rules — symbol exact**: any token in an external token list with symbol `USDY` / `OUSG` shows dimmed + Restricted chip in the selector. A non-matching token is unaffected.
- [x] **Pattern rules — symbol regex**: token symbol `OMPL` is soft-restricted in `mode='output'` (disabled in receive-as) and allowed in `mode='input'` (pay-with), matching existing semantics.
- [x] **Pattern rules — name regex**: a token whose symbol doesn't match but whose name starts with `Ondo ` (e.g. "Ondo US Dollar Yield") is also blocked.
- [x] **Case-insensitivity**: an entry `{ "symbols": ["usdy"] }` still catches tokens advertised as `USDY`.
- [ ] **Vault-level OR with patterns**: if a vault's underlying asset matches a pattern (e.g. `symbols: ["USDY"]` with a USDY vault), `isVaultBlockedByCountry` returns true for that vault — verify via the Restricted chip on browse pages.
- [x] **Regression — missing `assets.json` / `all/assets.json`**: remove the files upstream. Server returns empty list / just the present side, client behavior matches pre-PR.
- [x] **Bad regex handling**: ship a rule with `"symbolRegex": "*invalid("` — server rejects the whole file at validation, client falls through to empty.
- [x] **Sanitization** (Phase 3): ship a `"logo": "../secret.svg"` in `entities.json` — server rejects. Ship a 20KB string — server rejects. Ship an array with 20000 entries — server rejects.
- [x] **Memoization**: browse a lend page with many vaults, verify `isVaultBlockedByCountry` no longer iterates pattern rules per-row on every render (behavior-visible as: no perf regression with many pattern rules configured).
- [x] **`isInputAssetBlocked` scoping**: in withdraw mode, confirm that swap-withdraw with a collateral whose underlying is blocked surfaces the `isGeoBlocked` toast rather than silently disabling submit.
- [x] **Server-side geo-gate**: set `DEV_GEO_COUNTRY` to a sanctioned country; `/api/*` still returns HTTP 451.
- [x] `npm run lint` (touched files) and `npm run typecheck` (no new errors beyond the pre-existing `useMarketGroups.ts` issue on `development`).

## Summary by CodeRabbit

* **New Features**
  * Enforces asset-level geo-blocking/restrictions across swap/pay-with flows; blocked assets are unselectable.
  * Token selector rows render as disabled with a visible "Restricted" badge; blocked selections are prevented.
  * Adds a Restricted info modal explaining blocked vs restricted asset states.

* **Improvements**
  * Distinct warning toasts and disabled-reason messages for asset-specific blocks vs general swap restrictions.

* **Documentation**
  * Expanded geo-blocking docs covering asset-level rules and pattern-based matching.


Seranged (dismissed):
Image

Smoketest image

Seranged (approved):

PR #305, opened 2026-04-24
fix: verified race on direct navigation to earn/securitize vault pages

kasperpawlowski:
## Summary
- Direct navigation to an Earn vault page (and, latently, to a Securitize vault page) could render the vault as unverified because `getEarnVault` / `getSecuritizeVault` would fall through to a direct fetch with an empty labels context, yielding `verified: false`.
- Gate the Earn and Securitize vault page IIFEs on a new `useEulerLabels().isReady` signal so they wait for labels to load before resolving the vault. Internal navigation is unaffected (the flag is already true).
- The EVK branch of the Lend page now explicitly waits on both `isLabelsReady` and `isEscrowLoadedOnce` (in parallel) instead of relying on the transitive `loadLabels → loadVaults` ordering in `app.vue`.

## Test plan
- [x] Reload directly on an Earn vault page (e.g. `/earn/0x8bF41Ad2b816F7c220b22F4BCD63fC2A35Ab4247?network=8453`); the unverified disclaimer modal should NOT appear for a verified vault.
- [ ] Reload directly on a verified Securitize vault page; the unverified disclaimer modal should NOT appear.
- [x] Reload directly on a verified EVK vault page; the page should still render correctly with no regression.
- [x] Reload directly on an escrow EVK collateral page; still verified, no regression.
- [x] Reload directly on a genuinely unverified vault of each type; disclaimer modal should still appear as before.
- [x] Navigate from the in-app Earn/Lend lists to each vault type; no delay regressions on the already-loaded path.
- [x] Switch chains while on a vault page; labels refetch, `isReady` resets and re-flips when done.

### DIFF
--- a/composables/useEulerLabels.ts
+++ b/composables/useEulerLabels.ts
@@ -9,6 +9,7 @@ import { CACHE_TTL_5MIN_MS } from '~/entities/tuning-constants'
 import { normalizeAddress } from '~/utils/normalizeAddress'
 import {
   isLoading,
+  isReady,
   loadState,
   products,
   entities,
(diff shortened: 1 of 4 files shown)


Seranged:
@coderabbit review

Seranged (approved):

PR #307, opened 2026-04-24
feat: surface unresolved portfolio positions in a banner

kasperpawlowski:
## Summary

When the subgraph reports a position but the app can't materialise it — either the vault metadata lookup fails (`getOrFetch → undefined`) or the per-account lens call fails (RPC transport error, individual revert) — the position was silently dropped with no user feedback. We hit this in production with Plasma vault `0xa9C251…` when Chainstack's `eth_call` started reverting on `getVaultInfoFull`.

This PR surfaces those silently-dropped positions via a counts-only banner on the portfolio page, and instruments the Sentry signal so infra-side regressions show up before users report them.

## What changed

- **`composables/useAccountPositions.ts`** — new `UnresolvedPosition` type + `_unresolvedPositions` state. `registerUnresolved` is called at **five** previously-silent drop sites:
  - savings: `getOrFetch(entry.vault)` returning `undefined` → `fetch-failed`
  - borrow controller: `getOrFetch(borrowAddress)` returning `undefined` → `fetch-failed`
  - borrow collateral: `getOrFetch(collateralAddress)` returning `undefined` → `collateral-unresolved`
  - **per-account lens failures** (the review's critical case): `getAccountInfo` transport errors or reverts in both non-Pyth and Pyth borrow paths → `lens-failed`
  - savings lens failures: `getAccountInfo` transport errors or reverts in the deposit batch → `lens-failed`
  
  Dedup on `(subAccount, vault)` ignores `kind` so a failed borrow-collateral isn't double-counted when the same vault reappears in the deposit feed. Expose `unresolvedBorrowCount` / `unresolvedDepositCount` computed refs.

- **`composables/useEulerAccount.ts`** — `beginRefreshCycle` at refresh start and `finalizeRefreshCycle` at end. Replaces the prior `clearUnresolvedPositions` up-front. Entries that remain failing stay in the list continuously (no banner flash); entries that recover are pruned once per cycle. Errored refreshes skip finalize so a transient error doesn't prune still-failing entries.

- **`components/entities/portfolio/PortfolioUnresolvedBanner.vue`** — minimal counts-only banner matching `PortfolioShowAllHint` style (same `bg-warning-100` / `text-warning-500`, `info-circle` icon, `hint` transition). One sentence: *"N positions and M deposits couldn't be loaded. This is usually temporary — your funds are safe on-chain. If it persists, please reach out to the team."* Session-only dismiss auto-resets when a fresh failure arrives.

- **`pages/portfolio.vue`** — mounts the banner between `PortfolioShowAllHint` and the performance card so it's visible across all portfolio tabs.

## Sentry

When `NUXT_PUBLIC_SENTRY_DSN` is set, each capture emits a structured `captureMessage` with tags `{chainId, vaultType, kind, reason}` and context `{vault, kind, reason}`. No `subAccount` in the payload — it's wallet-derived and paired with session-replay would correlate replays to users. Session-level dedup keyed `${chainId}:${vault}:${reason}` means a broken endpoint emits once per chain/session instead of every 60s poll; the set is cleared on chain/address switch.

No-op when Sentry isn't configured (the app is forkable).

## Scope decisions

- Portfolio page only. Vault deeplink pages keep their existing "Unable to load Vault" toast — if a single-vault page fails, the problem is genuinely fatal for that page.
- No contact method in the banner copy — forks host their own support channels.
- Sub-account addresses aren't surfaced elsewhere in the app, so the banner intentionally shows counts only (no per-position breakdown).

## Test plan

- [ ] With `.env` pointing `RPC_URL_HTTP_9745` at the broken Chainstack endpoint and spy set to `0x5933ea3EE3FC7c6EAee0bA505ed3C3F9Fb80A6dC` on chain 9745, visit `/portfolio` — banner appears: "1 deposit couldn't be loaded."
- [ ] Force a lens error mid-session (temporarily misconfigure the proxy for a specific vault, or use a chain where Pyth simulation drops a result). Refresh should surface the lens-failed position — regression test for the silent-drop path beyond `getOrFetch`.
- [ ] Swap `RPC_URL_HTTP_9745` to `https://rpc.plasma.to` and wait for the next refresh → banner disappears, position renders in the Savings tab.
- [ ] 60s refresh with persistent failures → banner does **not** flash off/on (regression check).
- [ ] Main/happy chain (e.g. mainnet) — banner does not render.
- [ ] Dismiss → hidden for session; trigger a fresh failure (count grows) → banner re-appears.
- [ ] Without `NUXT_PUBLIC_SENTRY_DSN` → no network traffic to `/api/sentry-tunnel`, no runtime error. With DSN → one event per `(chainId, vault, reason)` per session with the expected tags/context.

## Summary by CodeRabbit

* **New Features**
  * Added a dismissible warning banner that appears when there are unresolved borrow or deposit positions and indicates whether issues affect borrows, deposits, or both.
  * Banner automatically reappears if new failures arrive after a prior dismissal.
  * Portfolio now tracks unresolved borrow/deposit counts so the banner shows up-to-date status.


Seranged (approved):
LGTM now

added follow-up stale-generation guard that prevents old wallet/chain/spy refreshes from leaking counts into the current portfolio state. 

(spy/address switching during inflight calls)

PR #310, opened 2026-04-24
fix: fix spy mode balance lookup for native-token deposit flows.

kasperpawlowski:
## Summary
Fixes spy mode balance lookup for native-token deposit flows.

When viewing as a spy address and choosing to deposit native ETH/native token through a wrapped-native flow (for example ETH → WETH), the deposit form could show the wrong native balance or `0`. The affected UI path uses `useWallets()` / `SwapTokenSelector`, not `useWagmi()`.

## Fix
- Reverted the earlier `useWagmi()` balance change because it targeted the wrong layer.
- Keep `useWagmi()` as connected-wallet state.
- In `useWallets()`, split native currency (`zeroAddress`) out of the ERC-20/lens `tokenBalances` batch.
- Fetch native balance with `client.getBalance({ address: balanceAddress })`.
- Merge the result back into the balance map under `zeroAddress`, so `SwapTokenSelector` and deposit `AssetInput` use the spy target’s native balance when spy mode is active.

## Test plan
- [x] Open a WETH/native-wrap supply form with no `spy` param
  - Select **Pay with ETH/native token**
  - Balance uses the connected wallet’s native balance
- [x] Open the same flow in spy mode, e.g. `/position/1/supply?collateral=0xEE074ceB44adA094391B9C71A7Fb8ba6962D492c&network=42161&spy=0x80DE5fBe9238F5F8F22860675a523c56215844B6`
  - Select **Pay with ETH**
  - Balance uses the spy address’s ETH balance
- [x] Enter more than the spy address’s ETH balance
  - Form shows “Not enough balance”
- [x] Toggle spy mode off
  - Pay-with-native balance reverts to connected wallet context

## Verification
- [x] `npx eslint composables/useWallets.ts composables/useWagmi.ts`

### DIFF
--- a/composables/useWagmi.ts
+++ b/composables/useWagmi.ts
@@ -22,6 +22,8 @@ function initializeWagmi() {
   const { disconnect: wagmiDisconnect } = useDisconnect()
   const { switchChain } = useSwitchChain()
   const { screenConnectedAddress, resetScreeningCache } = useAddressScreen()
+  const { spyAddress } = useSpyMode()
+  const { chainId: appChainId } = useEulerAddresses()
 
   const chainId = computed(() => wagmiChain.value?.id)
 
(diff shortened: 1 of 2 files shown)


Seranged:
@coderabbit review

Seranged (approved):
Completely reworked PR as initial was out of scope to problem, not merging until another validation - fixed PR + PR description + PR title

PR #323, opened 2026-04-25
feat: lazy-resolve off-label collateral vaults

kasperpawlowski:
**Summary**
Stop dropping off-label collateral vaults from Euler Lite discovery. After the bulk loaders + escrow lazy-fetch settle, scan all member EVK vaults for *live* collateral edges and feed any address not yet in the registry through the existing per-category bulk loaders.

**Why**
`verifiedVaultAddresses` is built exclusively from product labels (`euler-labels`). Any EVK vault that exists on chain but isn't listed under any product is invisible to the bulk EVK loader — and when a labeled borrow vault references it as collateral, every downstream view silently drops the relationship:

- `useMarketGroups.augmentWithCollateralGraph` can't add it to `externalCollateral` (no entry in `vaultMap`)
- `getCollateralExposurePairs` skips the pair when the resolver returns `undefined`
- the matrix and graph have no row/cell to render

The diagnostic `logWarn`s added in PR #309 surface the gaps; this PR fixes them by resolving the missing addresses lazily.

**Resolution path**
By the time the lazy-resolve pass runs, `fetchChainVaultCategories` has already populated the per-address category cache earlier in the same `loadVaults` call. So `fetchVaultCategory(addr)` is a cache hit for every address indexed by the subgraph, and we can group addresses by category and hand each group to the existing bulk loader for that type — same multicall batching, same registry-write path, no parallel implementation:

| `fetchVaultCategory(addr)` returns | dispatched to |
|---|---|
| `'escrow'` | `fetchNeededEscrowVaults` |
| `'evk'` | `updateEVKVaults` (multicall lens generator) |
| `'earn'` | `updateEarnVaults` (multicall lens generator) |
| `'securitize'` | `updateSecuritizeVaults` |
| `null` (subgraph hasn't indexed) | **skipped** — diagnostic warn fires; the next `loadVaults` cycle resolves it once the subgraph catches up |

The bulk loaders are invoked with `silent=true` so loading flags stay untouched (this runs after the initial reveal). Skipping null-category addresses avoids the legacy probe-and-guess path that misidentifies brand-new escrows as plain EVK.

**What changed**
- New pure helper `extractUnresolvedCollateralAddresses(evkVaults, isInRegistry, nowSeconds?)` in `entities/vault/collateral-discovery.ts` — returns the unique set of collateral addresses referenced by *live* edges (`isLiveCollateralEdge`) but not in the registry. Skips zero address. Snapshots the clock once. Tested in isolation.
- `composables/useVaults.ts`: after the existing `Promise.all(...)` for EVK / Earn / Securitize / escrow loaders, run a lazy-resolve pass that categorises every unknown via `fetchVaultCategory`, groups by category, and delegates to the existing bulk loaders. Generation guard preserved so chain switches abort cleanly. Non-explorable addresses are filtered out before fetching.
- Single-pass (not fixed-point): discovery views (matrix, graph, collateral-exposure block) iterate only product-member vaults, so a newly resolved off-label vault is a leaf in those views. Any second-hop unknowns will produce diagnostic warns and resolve on the next `loadVaults` cycle.
- Unit coverage in `tests/entities/vault/collateral-discovery.test.ts`: live unknown collaterals are surfaced, registry hits are skipped, fully ramped-out edges are excluded, mid-ramp edges are kept, zero address skipped, dedupe across vaults, case-insensitive registry membership, snapshotted-clock contract pinned.

**Validation**
- `npm run test:run` — full suite (441 tests) passes.
- `npm run test:run -- tests/entities/vault/collateral-discovery.test.ts` — 8/8 pass.
- Manual smoke (post-merge of #309 + this): on a product whose borrowable vault references an EVK vault not in any product label, expect the collateral to resolve and appear in the matrix/graph/exposure block within one load cycle.

**Stacked on**
PR #309 (`feature/eul4-147-ramping-node-fix`) — depends on the shared `isLiveCollateralEdge` predicate introduced there. Base will be re-targeted to `development` once #309 merges.

## Summary by CodeRabbit

* **New Features**
  * Added automated collateral discovery that identifies and resolves unregistered collateral addresses referenced by vaults during loading.
  * Implemented lazy loading of collateral assets to improve efficiency and performance.

* **Tests**
  * Added comprehensive test coverage for collateral address extraction and resolution logic.

### DIFF
--- a/composables/useMarketGroups.ts
+++ b/composables/useMarketGroups.ts
@@ -456,8 +456,7 @@ export const useMarketGroups = () => {
     const memberVaults: Vault[] = []
 
     try {
-      const ctx = buildFetchContext()
-      for await (const result of fetchVaults(ctx, allAddresses)) {
+      for await (const result of fetchVaults(allAddresses)) {
         memberVaults.push(...result.vaults)
         if (result.isFinished) break
       }
--- a/composables/useMarketGroups.ts
+++ b/composables/useMarketGroups.ts
@@ -456,7 +456,8 @@ export const useMarketGroups = () => {
     const memberVaults: Vault[] = []
 
     try {
-      for await (const result of fetchVaults(allAddresses)) {
+      const ctx = buildFetchContext()
+      for await (const result of fetchVaults(ctx, allAddresses)) {
         memberVaults.push(...result.vaults)
         if (result.isFinished) break
       }
--- a/composables/useVaults.ts
+++ b/composables/useVaults.ts
@@ -317,11 +317,12 @@ const fetchUnresolvedCollaterals = async (addresses: string[], generation: numbe
 
   if (loadGeneration.value !== generation) return
 
+  // Bulk loaders short-circuit on empty input, so call unconditionally.
   await Promise.all([
-    evkAddrs.length ? updateEVKVaults(evkAddrs, generation, true) : null,
-    earnAddrs.length ? updateEarnVaults(earnAddrs, generation, true) : null,
-    securitizeAddrs.length ? updateSecuritizeVaults(securitizeAddrs, generation, true) : null,
-    escrowAddrs.length ? fetchNeededEscrowVaults(escrowAddrs, generation) : null,
+    updateEVKVaults(evkAddrs, generation, true),
+    updateEarnVaults(earnAddrs, generation, true),
+    updateSecuritizeVaults(securitizeAddrs, generation, true),
+    fetchNeededEscrowVaults(escrowAddrs, generation),
   ])
 }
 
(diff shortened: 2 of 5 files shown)


Seranged (dismissed):
Approved after adding bounded follow-up discovery so off-label collateral vaults loaded into the registry also bring in their own live collateral pairs, plus fixing the direct-market fetchVaults callsite; tests, typecheck, eslint, and targeted explore smokes are clean.

kasperpawlowski:
the last commit appears to be redundant

Seranged (approved):

PR #325, opened 2026-04-26
Fix swap quote slippage validation

Seranged:
## Summary
- Validate swap quote slippage against the user-selected tolerance before building verifier calldata.
- Surface quote-applied slippage in swap summaries when it differs from the selected tolerance.
- Keep Multiply transaction review focused on the actual swap delta rather than projected total position sizes.

## Changes
- Thread requested slippage through swap plan builders so verifier data is rebuilt and checked from the current quote and tolerance.
- Add shared exact-in and target-debt slippage validation helpers aligned with the swap verifier paths.
- Pass computed quote slippage into swap detail summaries across affected supply, withdraw, repay, borrow, collateral swap, debt refinance, and multiply flows.
- Add regression coverage for quote slippage display and verifier validation edge cases.

## Test plan
- [x] `npm run test:run`
- [x] `npm run test:run -- tests/utils/swap-slippage-validation.test.ts tests/utils/swap-quotes.test.ts`
- [x] `npx eslint 'pages/position/[number]/multiply.vue'`
- [x] `git diff --check`
- [x] Manually smoke tested supply with swap, withdraw with swap, collateral swap, debt refinance, and multiply at `0.05%` and `1%` slippage.
- [x] Checked `npm run typecheck`; it still fails on pre-existing unrelated `composables/useMarketGroups.ts` typing.

## Summary by CodeRabbit

* **New Features**
  * Swap UIs display quote-derived slippage alongside your configured slippage and show a warning when they differ.
  * Transaction builders now forward the chosen slippage into swap plans so displayed and executed slippage align.
  * Centralized slippage validation added to reject inconsistent or out-of-range quotes.

* **Chores**
  * Minimum allowed slippage raised slightly.

* **Tests**
  * Added comprehensive tests for slippage computation, application, rounding, and validation (including warning logs).

### DIFF
--- a/components/entities/swap/SwapDetailsSummary.vue
+++ b/components/entities/swap/SwapDetailsSummary.vue
@@ -2,11 +2,12 @@
 import { isPriceImpactWarning, isSlippageWarning } from '~/utils/priceImpact'
 import { formatNumber } from '~/utils/string-utils'
 
-defineProps<{
+const props = defineProps<{
   inputDisplay: string | null
   outputDisplay: string | null
   priceImpact: number | null
   slippage: number
+  quoteSlippage?: number | null
   routedVia: string | null
   multipliedPriceImpact?: number | null
 }>()
@@ -14,6 +15,20 @@ defineProps<{
 const emit = defineEmits<{
   (e: 'openSlippageSettings'): void
 }>()
+
+const SLIPPAGE_DIFF_TOLERANCE = 0.005
+
+const slippageDiffers = computed(() =>
+  props.quoteSlippage !== null
+  && props.quoteSlippage !== undefined
+  && Math.abs(props.quoteSlippage - props.slippage) >= SLIPPAGE_DIFF_TOLERANCE,
+)
+
+const quoteSlippageLabel = computed(() =>
+  props.quoteSlippage !== null && props.quoteSlippage !== undefined
+    ? formatNumber(props.quoteSlippage, 2, 0)
+    : '',
+)
 </script>
 
 <template>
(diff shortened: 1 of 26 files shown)


kasperpawlowski:
Pushed a follow-up addressing review nits, all in one commit (fa5ac1e):

- **Defensive parsing in `verify.ts`**: switched the four `BigInt(quote.amount*)` reads in `validateSwapQuoteSlippageData` to `BigInt(... || 0)` so a malformed quote payload surfaces a validation error instead of a SyntaxError.
- **Structured logging**: failures now go through `logWarn('swapQuoteSlippage', ...)` to match the rest of the codebase. Dropped the dev-only `console.info` success log — it was diagnostic noise with no logger equivalent.
- **`MAX_SLIPPAGE` import**: removed the local literal in `verify.ts` and re-used `entities/constants.ts`.
- **Slippage-warning flicker**: replaced the relative `1.0001` divergence factor with an additive `VALIDATION_DIVERGENCE_TOLERANCE_PP = 0.001` (in slippage points). It's now strictly less than `SLIPPAGE_DIFF_TOLERANCE = 0.005` at every slippage value (incl. `MAX_SLIPPAGE`), so the user-facing warning can never fire on a quote the validator just accepted. Tests updated with the new boundary values plus a regression test at 50%.
- **`MIN_SLIPPAGE = 0.01`**: a 0% slippage would force exact-equality validation and almost always fail with a generic toast — bumping the floor to 1bp avoids that footgun. The slippage settings UI clamps to this floor and the error message now reads "between 0.01% and 50%".
- **`buildMultiplyPlan` typing**: replaced the `requestedSlippage?: number` + runtime-throw + `requestedSlippage!` pattern with a discriminated union — `{ quote: SwapApiQuote, requestedSlippage: number }` xor `{ quote?: undefined, requestedSlippage?: never }`. Call sites in `useMultiplyForm.ts` and `multiply.vue` build args conditionally, so the swap branch can no longer be reached without a slippage. A local `swapInfo` capture preserves narrowing through the function body.

Tests: 447 pass. Lint clean. No new typecheck errors (one pre-existing `useMarketGroups.ts` error on this branch is unrelated).

kasperpawlowski (approved):

PR #346, opened 2026-04-28
feat: surface swap mode in operation review steps

kasperpawlowski:
## Summary

The operation review modal — the per-step list shown before a user confirms a transaction — now reflects the actual swap mode behind the operation:

1. **TARGET_DEBT swaps relabel as "Swap to repay".** When a user repays debt by swapping a collateral or wallet asset into the borrow asset, the step is labeled `Swap to repay` instead of just `Swap`. Plain swaps (deposit-side conversion, multiply, collateral→collateral) keep `Swap`.

2. **The unpinned leg of every swap is prefixed with `~`** to make clear which side is exact and which is an estimate that can move within slippage:

   | Mode | Pinned (exact) | Estimated (`~`) |
   |---|---|---|
   | `EXACT_IN` | input | output |
   | `EXACT_OUT` | output | input |
   | `TARGET_DEBT` | target debt amount | input |

## Implementation

- `utils/stepDecoding.ts` — adds `swapMode?: SwapperMode` to `StepDecodingContext` and `estimated?: boolean` to `StepAssetInfo`. `buildDisplaySteps` derives the relabel and the per-leg estimated flag from `ctx.swapMode` only when the step label is `Swap`.
- `OperationReviewModal.vue` — accepts an optional `swapMode` prop and forwards it into the decoding context.
- `OperationStepsList.vue` — renders the `~` prefix on each leg whose `estimated` is true.
- **No calldata decoding.** The form composables already pass `swapperMode` into their swap-quote builders, so this PR just plumbs that same value one prop further into the modal at every call site that already passes `swapToAsset`/`swapToAmount`.
- `useSwapPageLogic` gets a new required `swapperMode` config option so the three swap-page entry points can pass their static mode through.

## Test plan

- [ ] Lend → swap (`/lend/[vault]/[subAccount]/swap`): `Swap` label, output side prefixed with `~`.
- [ ] Position → collateral swap (`/position/[number]/collateral/swap`): `Swap` label, output side prefixed with `~`.
- [ ] Position → borrow swap (`/position/[number]/borrow/swap`): `Swap to repay` label, input side prefixed with `~`, target-debt output amount exact.
- [ ] Lend deposit with input-asset swap-in (`/lend/[vault]` choosing a non-asset input): `Swap` label, output (target asset) prefixed with `~`.
- [ ] Lend withdraw with output-asset swap-out (`/lend/[vault]/[subAccount]/withdraw` choosing a non-asset output): `Swap` label, output prefixed with `~`.
- [ ] Multiply (`/position/[number]/multiply`): `Swap` label, output (long asset) prefixed with `~`.
- [ ] Borrow with collateral-side swap (`useBorrowForm` swap-borrow path): `Swap` label, output (collateral) prefixed with `~`.
- [ ] Repay flows that swap (collateral repay, savings repay, wallet swap repay) with both `EXACT_IN` and `TARGET_DEBT` directions: relabel and `~` move correctly when toggling direction.
- [ ] Same-asset repay/transfer/swap (no actual swap step): no relabel, no `~`.
- [ ] Operations without any swap (plain supply/borrow/withdraw/repay): no `~` anywhere.

## Summary by CodeRabbit

* **New Features**
  * Operation review now accepts and shows swap mode and an estimated-side override for clearer swap reviews.
  * Swap steps can relabel (e.g., "Swap to repay") when in target-debt mode.

* **Bug Fixes**
  * Estimated amounts are now prefixed with "~" to distinguish them from exact values.

### DIFF
--- a/components/entities/operation/OperationReviewModal.vue
+++ b/components/entities/operation/OperationReviewModal.vue
@@ -22,7 +22,7 @@ interface REULUnlockInfo {
   daysUntilMaturity: number
 }
 
-const { type, asset, assetIconUrl, campaignInfo: _campaignInfo, reulUnlockInfo, amount, onConfirm, plan, swapToAsset, swapToAmount, swapMode, supplyingAssetForBorrow, supplyingAmount, transferAmounts, submittingLabel } = defineProps<{
+const { type, asset, assetIconUrl, campaignInfo: _campaignInfo, reulUnlockInfo, amount, onConfirm, plan, swapToAsset, swapToAmount, swapMode, swapEstimatedSide, supplyingAssetForBorrow, supplyingAmount, transferAmounts, submittingLabel } = defineProps<{
   type?: 'supply' | 'withdraw' | 'borrow' | 'repay' | 'swap' | 'transfer' | 'reward' | 'brevis-reward' | 'fuul-reward' | 'reul-unlock' | 'disableCollateral' | 'swap-supply' | 'swap-withdraw' | 'swap-borrow'
   asset: VaultAsset
   assetIconUrl?: string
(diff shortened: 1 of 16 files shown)


Seranged (approved):
Smoke-tested the swap review modal matrix and verified the follow-up fixes:

- `571c77fa` adds swap-mode display in review steps: `TARGET_DEBT` relabels to `Swap to repay`, and estimated swap legs get `~`.
- `e179850e` fixes review estimate edge cases:
  - borrow debt-swap marks the displayed input/new-debt side as estimated instead of the exact target debt side
  - wallet swap repay `TARGET_DEBT` shows the exact typed debt amount instead of the slippage-min output
- `9d4977af` makes the swap-mode-to-estimated-side mapping explicit/exhaustive.

Checked `TARGET_DEBT`, `EXACT_IN`, same-asset, and no-swap cases. UX looks right.

PR #372, opened 2026-05-05
feat: project utilization-aware APY and ROE across forms

Seranged:
## Summary
- Show utilization-aware APY and ROE estimates throughout borrow, multiply, collateral, repay, refinance, migration, and lend forms.
- Make projected values inspectable through consistent info and sparkle hover or tap breakdowns for market rates, yield contributions, and reward campaigns.
- Keep projected metrics unavailable until every required price, rate, reward, quote, and form identity input belongs to the current complete state.

## Changes
- Model complete post-action collateral and debt portfolios, including wallet versus savings cash semantics, swap-funded deposits, lend rebalances, same-vault savings repayments, and old-debt refinance repayments.
- Size same-asset refinance estimates with the 0.01% SDK cushion for Euler migrations and the existing 1% external-migration buffer.
- Re-evaluate borrow and looping rewards against the projected collateral set and preserve reward token, provider, action, debt or supply vault, and qualifying collateral identity.
- Display before and after lending or borrow rates and disambiguate same-symbol vaults only when their addresses are needed.
- Invalidate delayed and in-flight estimates on source, quote, vault, position, settings, and navigation changes; clear current baselines while replacements load or fail; and reject missing or partial rate projections.
- Align repay projections with execution by using mode-aware, source-capped swap inputs for TARGET_DEBT and clearing stale quotes before debounced replacements.
- Keep the calculations in display and estimation state; transaction plan construction, signing, batching, and submission paths remain unchanged.

## Test plan
- [x] npm test -- --run — 142 files, 1,330 tests
- [x] npm run typecheck
- [x] npm run lint — 0 errors; 6 existing warnings
- [x] npm run build
- [x] Local desktop smoke: collateral repay and multiply produced projected rate transitions; swap-funded collateral supply showed a quote and rate transition, and clearing the amount removed the estimate without a late restore
- [x] Local mobile smoke at 390x844: the projected breakdown remained accessible without horizontal overflow, and clearing the amount removed the estimate


Seranged:
@coderabbit review

LeonardEulerXYZ (changes requested):
## Re-review after latest fixes

I re-checked the current head (`9c9d53a`) against the CodeRabbit APY/ROE critiques and ran local validation from a fresh PR worktree.

Validation:
- `npm run test:run` passed: 623 passed, 1 skipped.
- `npm run build` passed. Warnings were the usual Sentry/Reown/chunk-size noise, not a PR-specific failure.

The latest commit resolves several of the original critique classes:
- repay `roeAfter` now receives `nextCollateralSupplyApy` instead of always reusing the before-state collateral APY;
- `useRepayHealthMetrics` now guards/clears projected borrow APY failures;
- `getCollateralApySnapshot()` now returns a safe `{ supplyUsd: 0, weightedSupplyApy: null }` fallback;
- multiply current/next collateral snapshot watchers now use race guards;
- same-asset refinance no longer depends on an external quote for the projected summary path.

I still see two correctness issues worth fixing before merge.

### Blocking: same-asset debt refinance can under-borrow from a truncated display amount

`setFromAmountToMax()` intentionally truncates the exact debt to ~6 significant digits for display. The same-asset debt-swap plan then parses that display string and passes it as `amount` to `buildSameAssetDebtSwapPlan()`.

That plan borrows `adjustForInterest(amount)`, skims it into the old vault, then calls `repayWithShares(maxUint256)` and `disableController()` on the old vault. For normal-sized debts the 0.01% cushion covers the truncation, but for small/dust debts it does not. Example with the current truncation rule:

- exact debt `0.000001234567`
- displayed/parsed amount `0.000001`
- after 0.01% cushion: `0.0000010001`, still only ~81% of the exact debt

In that case the plan can leave residual debt in the old vault and then fail at `disableController()`. Since the UI says this flow requires the full debt amount, the same-asset path should use the exact `currentDebt.value` for the plan/summary debt amount rather than the truncated display string, or otherwise preserve an exact max amount for this path.

Relevant code:
- `pages/position/[number]/borrow/swap.vue`: `setFromAmountToMax()` truncates `currentDebt` into `fromAmount`.
- `pages/position/[number]/borrow/swap.vue`: same-asset `buildPlan()` parses `fromAmount` into `amount`.
- `composables/useEulerOperations/swaps/same-asset.ts`: same-asset plan borrows/skims the adjusted amount, then repays all old debt and disables the old controller.

### Blocking: borrow refinance projected-ROE watcher still lacks failure fallback

The new `watchEffect` in `pages/position/[number]/borrow/swap.vue` is race-guarded, but the async `Promise.all()` is not wrapped in `try/catch`. It calls `getProjectedRates()` and pricing helpers; `getProjectedRatesBatch()` can still throw on the direct `readContract` fallback path.

If a projection request rejects, the watcher exits before the stale-generation check and before clearing/updating `nextBorrowValueUsd` / `nextBorrowApy`. That can leave stale after-ROE data on screen for the current selected vault pair instead of degrading to `null` / current APY. This is the same class of failure the latest fixes correctly hardened in `useRepayHealthMetrics` and `getCollateralApySnapshot()`.

Suggested shape:
- wrap the projected borrow `Promise.all()` in `try/catch`;
- if the generation is still active, clear `nextBorrowValueUsd` and `nextBorrowApy` to `null` on failure;
- optionally fall back to unprojected `fromBorrowApy` / `toBorrowApy` only if that is an intentional UX choice.

### Non-blocking coverage note

The current tests cover the helper-level APY refactor, but not the higher-risk reactive paths directly: `usePositionCollateralApy`, `useRepayHealthMetrics` stale/failure behavior, same-asset borrow-swap estimates, collateral-swap snapshot fallback, or multiply watcher races. I would add a small focused test around the two blockers above; the green suite is good compatibility evidence, but it does not discharge these edge cases.

LeonardEulerXYZ (approved):
## Incremental re-review

Re-checked latest head `504c9e7` (`fix: harden debt refinance estimates`) against the two blockers from my previous review.

Resolved:
- Same-asset debt refinance now uses `currentDebt.value` for the same-asset plan amount instead of reparsing the truncated display string.
- Same-asset projected summary now also uses exact `currentDebt.value`, so the displayed after-state is aligned with the actual plan path.
- The borrow-refinance projected-ROE watcher now wraps the async projection work in `try/catch` and clears `nextBorrowValueUsd` / `nextBorrowApy` for the active generation on failure.

Validation:
- `npm run test:run` passed: 623 passed, 1 skipped.
- `npm run build` passed. Warnings are the existing Sentry/Reown/chunk-size warnings.

I did not find any remaining blocker in the latest delta. Approval is bounded to the CodeRabbit/APY/ROE critique surface and the two follow-up blockers above.

LeonardEulerXYZ (dismissed):
## Deep agent-assisted re-review

Reviewed current head `504c9e7` with four independent passes: APY/ROE math, async/reactive behavior, transaction/same-asset trust boundaries, and tests/coverage. I then re-anchored the material findings against the current code before posting.

Validation run from `/home/leonard/euler-lite-worktrees/pr-372`:
- `npm run typecheck -- --noEmit` ✅
- `npm run test:run -- tests/entities/vault/apy.test.ts tests/utils/repay-utils.test.ts tests/composables/useSavingsRepay.test.ts` ✅ (`60` tests)
- `npm run test:run` ✅ (`623` passed, `1` skipped)
- `npm run build` ✅

### CodeRabbit critique status

I re-checked the 7 CodeRabbit critiques on the current head:

1. `useCollateralSwapRepay` post-repay weighted collateral APY for `roeAfter` — resolved; current and next weighted APYs are now separated.
2. `useRepayHealthMetrics` stale projected borrow APY updates — resolved; the projected borrow watcher uses a race guard and active-failure clearing.
3. `usePositionCollateralApy` safe fallback — resolved; `getCollateralApySnapshot()` now catches and returns `{ supplyUsd: 0, weightedSupplyApy: null }`.
4. borrow refinance same-asset projected path — same-asset path itself is resolved; the watcher no longer requires `quote.value` for same-asset and uses `currentDebt.value` there.
5. collateral swap ROE snapshot failure clearing — no longer actionable as originally stated because the shared snapshot helper now catches and returns the safe snapshot.
6. multiply current snapshot race guard — resolved.
7. multiply next snapshot race guard — resolved.

### Blocking finding

One blocker remains from the deeper transaction-path pass:

- Cross-asset debt refinance still derives the quote amount and projected old-debt reduction from the truncated display string `fromAmount`, while the actual transaction plan/verifier is full-refinance (`targetDebt: 0n`, `currentDebt: currentDebt.value`). The same-asset path was fixed to use exact `currentDebt.value`, but the cross-asset path still has a quote/summary-vs-plan mismatch. For example, `1234567890123456789` at 18 decimals is displayed as `1.23456`, which parses back to `1234560000000000000`, short by `7890123456789` wei. Dust-sized debts can be much worse proportionally.

Expected invariant: for a “refinance full current debt” flow, the quote request and projected debt summary should use the same exact debt amount as the transaction/verifier path, not the shortened UI display string.

### Non-blocking hardening

- The new collateral portfolio watchers in `useSavingsRepay` / `useCollateralSwapRepay` clear refs on invalid dependencies before advancing their race guard, so an older in-flight snapshot can theoretically repopulate stale derived ROE/APY values after invalidation. This is display/derived-metric stale-state hardening, not a direct transaction construction issue, so I’m not treating it as the blocker here.

### Test coverage note

The green suite is useful compatibility evidence, but there is still no focused regression coverage for the cross-asset debt-refinance exact-amount invariant. A good test would mount/harness `pages/position/[number]/borrow/swap.vue` with a high-precision `currentDebt`, assert the displayed max may be shortened, but verify the quote request/projected summary/full-refinance plan all use exact `currentDebt.value` for the debt being refinanced.

LeonardEulerXYZ (changes requested):
## Current-head re-review

Reviewed head `c0384d0` across the projected APY/ROE diff, the latest CodeRabbit feedback, and sibling form paths. CI lint, typecheck, and tests are green. Local focused tests and headed browser smoke could not run because the review host had no free disk space; no screenshot evidence was produced or posted.

### Blocking: invalid-state branches do not invalidate in-flight APY snapshots

The repay-page watcher clears `netAPY` when its position/vault inputs disappear, but advances `netApyGuard` only after that check. An older request can therefore finish after disconnect/navigation/input replacement, pass `isStale(gen)`, dereference the now-missing reactive inputs, and repopulate the cleared summary with stale APY. The latest CodeRabbit finding is valid.

The same newly introduced pattern appears in `useCollateralSwapRepay` and `useSavingsRepay`: their empty-input branches reset summary refs without advancing their guards, so prior snapshots can restore stale collateral APY/value/address state (and read `position.value` after it became undefined). Please advance/capture each generation before the early-return check and use captured input values across the await. A focused deferred-promise regression test should cover invalidation while a snapshot is in flight.

### Maintainability / sibling-flow pass

The shared `usePositionCollateralApy` abstraction is the right direction and most borrow, multiply, collateral, and repay surfaces now use it. Two hygiene gaps remain:

- refinance still independently composes a weighted projected collateral portfolio instead of using the shared snapshot abstraction, leaving two implementations with different missing-data semantics;
- the new race guards have no focused out-of-order/invalidation tests, despite being repeated across several user-editable form paths.

I did not elevate the existing lend-withdraw projected-APY unit issue because it is unchanged from `development`, not introduced by this PR. Earlier CodeRabbit APY composition findings are otherwise resolved on the current head.

Smoke coverage: incomplete — code/diff review plus hosted CI only; no browser, mobile, wallet, or signing smoke due to the disk-space constraint.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `e4eb613f` across the projected APY/ROE calculations, async state handling, transaction-estimate alignment, sibling form consumers, tests, and browser rendering.

### Blocking: a failed requested projection is reported as a complete after-estimate

`getCollateralApySnapshot()` requests utilization-aware rates for entries marked `projectRates`, but a `null` result is converted to `projectedRaw = null` and then passed through the current-rate fallback. The snapshot still returns `isComplete: true`. A transient/unsupported lens result can therefore make the after-action Net APY/ROE look valid while it is actually using the pre-action supply APY.

I reproduced this with a local throwaway regression: mock `getProjectedRatesBatch()` to return `[null]` for a requested collateral projection. The snapshot returned the current weighted APY with `isComplete: true`; the fail-closed expectation failed. Please reject missing results for requested projection entries before weighting, and commit a focused regression test for this branch.

### Validation

- `npm run test:run` — 136 files passed, 1 skipped; 1272 tests passed, 1 skipped.
- `npm run build` — passed; only non-fatal build warnings.
- Local throwaway null-projection regression — failed as described above, then removed.
- Headed Chromium smoke against the Railway PR preview and a clean `development` build at 1440×1000 and 390×844.
  - Base fixture: Base position 1, cbETH/WETH.
  - Exercised `/position/1/repay` and `/position/1/multiply`; entered `0.000001` rather than stopping at route load.
  - Net APY/ROE summaries rendered without clipping or overlap on desktop and mobile; no PR-specific visual blocker found.
  - Wallet/signing smoke was not run.

The shared snapshot abstraction is the right shape: it is reused by borrow, multiply, repay variants, and the shared collateral form rather than duplicating the weighting logic in each page. That also makes this fail-open behavior broad: every consumer requesting projected collateral rates inherits it. The fix and regression belong in `usePositionCollateralApy`, not in one visible route. I did not find a separate sibling-only implementation that should be patched instead.

### Reviewer-bot feedback

The earlier CodeRabbit race/failure findings were useful evidence but are superseded by later fixes on this head. The remaining blocker above is distinct: the batch call resolves successfully, but one requested item is `null`, so the outer `catch` does not help.

<details>
<summary>Validated browser evidence</summary>

Development comparator — desktop repay:

Development desktop repay

PR #372 — desktop repay:

PR 372 desktop repay

Development comparator — desktop multiply:

Development desktop multiply

PR #372 — desktop multiply:

PR 372 desktop multiply

PR #372 — mobile Net APY summary (390×844):

PR 372 mobile Net APY summary

PR #372 — mobile ROE summary (390×844):

PR 372 mobile ROE summary

</details>

LeonardEulerXYZ (changes requested):
## Current-head incremental review — changes requested

Reviewed head `766bf678` against the previous null-projection blocker, the broader projected APY/ROE paths, active bot feedback, and sibling form consumers. The latest commit correctly makes `usePositionCollateralApy` reject a missing requested collateral projection and adds the focused regression. One sibling implementation still accepts the same failure state.

### Blocking: refinance still presents incomplete projections as valid ROE

The refinance estimator batches projected-rate requests for the changed collateral and debt legs, but it does not reject a short result array or a `null` result. Missing entries are stored as `null`; the collateral and borrow APY helpers then fall back to current rates, so `roeAfter` can look complete even though one or more requested utilization-aware projections failed. This is the same fail-open class fixed by the latest commit in the shared collateral snapshot.

Please make the refinance path fail closed when any requested projection is absent and add focused coverage for a partial/`null` batch result.

### Validation

- `npm run test:run` — passed: 136 files passed, 1 skipped; 1,273 tests passed, 1 skipped.
- Focused APY/ROE suites — passed: 6 files, 22 tests.
- `git diff --check origin/development...HEAD` — passed.
- Current CI lint, typecheck, tests, and Railway deployment are green.
- Browser/mobile smoke was not rerun for this incremental head: the only delta from the previously smoke-tested head is the shared-helper null-projection fix plus its unit test. The prior GitHub review already contains desktop/mobile evidence; no new screenshots were captured or posted. Wallet/signing smoke was not run.

The shared `usePositionCollateralApy` abstraction and new regression are the right shape. The remaining issue exists because refinance independently reimplements projected portfolio weighting and therefore has different completeness semantics. I also checked borrow, multiply, and repay sibling consumers: projection-readiness policy is still distributed, but the concrete blocker here is the refinance batch path. A shared projection-completeness helper or explicit common invariant would reduce this drift.

### Reviewer-bot feedback

The seven active CodeRabbit inline claims describe earlier heads and are resolved or obsolete as written on this head. I did not promote them into current findings. The blocker above is distinct and verified against the current implementation.

Smoke coverage: incremental code/diff review plus full tests/typecheck and hosted CI; browser/mobile evidence inherited from the immediately preceding reviewed head; no wallet/signing smoke.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `7a84e0e`. The latest commit fixes the prior refinance partial/`null` batch blocker and applies the new completeness helper across the batch-projection call sites I checked. One failure-state class remains in the shared repay path, with the same stale-value variant in both wallet repay paths.

### Findings

1. **Repay ROE still falls back to the current borrow APY when the requested post-repay rate is unavailable.** `projectedBorrowApy` is correctly cleared on a `null` result/rejection, but `roeAfter` immediately substitutes `borrowApy`. This produces a complete-looking after-ROE using a pre-action rate. A local throwaway regression that makes `getProjectedRates()` return `null` failed: expected `roeAfter === null`, received `0` from the mocked `getRoe()` path.
2. **Wallet repay rejection paths retain an earlier Net APY estimate.** The resolved-`null` branches now fail closed, but the `catch` blocks in both `useWalletRepay` and `useWalletSwapRepay` only log. Since sync estimates have already set `hasEstimate = true`, a transient rejected projection can leave `_estimateNetAPY` from a prior amount/quote visible as the current post-action estimate.

### Validation

- Focused Vitest: 10 files, 49 tests passed.
- `npm run typecheck` passed.
- `npm run build` passed; non-fatal bundler warnings only.
- Local throwaway projection-failure regression: failed as described above, then removed.
- Headed Chromium against the Railway preview at 1440×1000 and 390×844: `/position/3/repay` and `/position/3/multiply` rendered the real app; repay-from-collateral produced quote/summary state, and multiply moved from `9.06x` to `9.07x` with ROE/LTV/liquidation-buffer deltas visible. No material page exceptions. Wallet/signing smoke was not run.
- Screenshots: not captured; this pass validated data-state transitions rather than a visual-layout delta.

The shared snapshot and completeness helpers are the right abstraction. The remaining gap is that failure completeness is not carried through the shared repay metric or the two sibling wallet estimators. Please make rejected and resolved-missing projections converge on the same unavailable state and add focused tests for those reusable paths.

### Bot feedback disposition

The older CodeRabbit race-guard critique is resolved, but its failure-fallback conclusion is incomplete: setting `projectedBorrowApy` to `null` does not fail closed while `roeAfter` coalesces it back to the current rate. The other previously active APY/ROE critiques checked here are resolved or stale on this head.

LeonardEulerXYZ (commented):
## Current-head incremental review — prior blockers resolved

Reviewed head `176b0e5a` against the two failure-state blockers from `7a84e0e`, the affected sibling flows, active reviewer feedback, and the current hosted preview.

### Verdict

No remaining blocker found in the latest delta.

- `useRepayHealthMetrics` now carries explicit projected-borrow completeness. A missing or rejected requested projection keeps `roeAfter` unavailable instead of silently reusing the current borrow APY.
- `useWalletRepay` and `useWalletSwapRepay` now clear the active Net APY estimate when projection work rejects, so a prior amount/quote cannot remain visible as the current after-estimate.
- The new regressions are discriminating: they cover resolved-`null` and rejected shared projections, plus success-then-rejection stale-value cases in both wallet repay variants.

### Validation

- Focused Vitest: 3 files, 8 tests passed.
- `npm run typecheck` passed.
- `git diff --check origin/development...HEAD` passed.
- Hosted CI lint, typecheck, tests, and Railway deployment are green.
- Headed Chromium smoke on the current Railway preview at 1440×1000 and 390×844:
  - `/position/3/repay?network=1` — entered a collateral-repay amount; ROE, LTV, health, and quote summary remained rendered without material page errors.
  - `/position/3/multiply?network=1` — changed multiplier `9.06x → 9.07x`; ROE/LTV summary deltas rendered on desktop and mobile.
- Wallet/signing smoke was not run.
- Screenshots were not captured or posted: this incremental change alters projection failure-state behavior rather than layout, and the exercised success-state UI had no new visual delta to document.

The repaired repay paths now converge on the intended fail-closed policy, but projection completeness is still represented independently across borrow, multiply, collateral, refinance, and repay flows. A shared projection-result state would reduce future drift between `null`, rejection, stale-generation, and fallback behavior. This is non-blocking for the current delta.

One small test-hardening gap remains: the shared repay tests prove both failure modes hide `roeAfter`, but do not explicitly assert that a successful projection restores a non-null after-ROE. The implementation does set completeness only on valid projection success, so I did not find a production defect here.

### Reviewer feedback disposition

The earlier CodeRabbit and Leonard failure/race comments describe older heads. The two still-visible Leonard inline findings were resolved by this commit and removed rather than left as stale instructions.

Smoke coverage: focused code/tests/typecheck plus desktop/mobile browser interaction; no wallet/signing coverage.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `e4207d8d` across projected APY/ROE calculations, repay/refinance/multiply form integration, async state handling, sibling consumers, tests, and the hosted preview.

### Blocking finding

The new collateral/savings projection paths use `core.spent`, which reads `quote.amountIn` for every cross-asset repay. In `TARGET_DEBT` mode, however, the submitted plan, balance gate, and review amount use the mode-aware input (`amountInMax` when present). When those fields differ, the new collateral `assetsDelta` and savings `cashDelta` understate how much leaves the source vault, so projected supply APY/ROE—and collateral-repay health—can be optimistic relative to the transaction. Near the health boundary this may present an actionable-looking estimate that simulation later rejects.

This should be fixed once in the shared repay core using the existing mode-aware input helper, then covered with a quote where `amountIn=100` and `amountInMax=105`; both collateral and savings consumers inherit the result. The live Nordstern/Kyber/etc. fixture happened to return equal fields, but the SDK contract and existing quote/review tests explicitly support distinct values, so equality is not a safe invariant.

### Validation

- Focused repay/multiply/projected-yield Vitest: 70 tests passed.
- `npm run typecheck` passed.
- `npm run build` passed; non-fatal bundler warnings only.
- Headed Chromium against the Railway preview at 1365×900 and 390×844.
  - Exercised `/position/4/repay?network=1` with **From collateral** and a 10 USDC debt target.
  - Quotes loaded; ROE, projected market rates, LTV, health, liquidation-price/buffer, and swap details rendered.
  - The Projected ROE modal fit on desktop and mobile without clipping/overlap.
- Wallet/signing smoke was not run.

The shared snapshot/completeness abstractions are the right shape. This finding is also a useful example of where the invariant belongs: source-spend semantics should be normalized in `useRepaySwapCore`, not independently patched in collateral and savings projections. I checked the sibling consumers; both currently inherit the same mismatch.

### Reviewer-bot feedback

The seven visible CodeRabbit inline claims refer to earlier heads and are resolved or obsolete as written on this head. I did not repeat them as current findings.

<details>
<summary>Validated browser evidence</summary>

Desktop collateral repay — Projected ROE breakdown:

Desktop collateral repay — Projected ROE breakdown

Mobile collateral repay — Projected ROE breakdown (390×844):

Mobile collateral repay — Projected ROE breakdown

</details>

Smoke coverage: headed desktop/mobile interaction smoke; no wallet/signing coverage.

composables/repay/useCollateralSwapRepay.ts
```diff
@@ -224,21 +230,70 @@ export const useCollateralSwapRepay = (options: UseCollateralSwapRepayOptions) =
...
+    collateralSnapshotComplete.value = false
+    nextCollateralSnapshotComplete.value = false
+    collateralSnapshot.value = null
+    nextCollateralSnapshot.value = null
+    const [currentSnapshot, nextSnapshot] = await Promise.all([
+      getCollateralApySnapshot(currentPosition, currentBorrowVault),
+      getCollateralApySnapshot(currentPosition, currentBorrowVault, {
+        deltas: [{
+          vaultAddress: currentSourceVault.address,
+          assetsDelta: -spent,
```
Seranged, line 283: Fixed in 042b9705. Shared repay depletion now uses the mode-aware, source-capped input, invalidates stale quotes before replacement, and fails closed on invalid or unavailable inputs. Added core plus collateral and savings consumer regressions.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed `042b9705` across projected APY/ROE math, refinance execution alignment, async invalidation/failure handling, active bot feedback, sibling form consumers, and scalability/maintainability.

### Blocking findings

1. Same-asset refinance projections size the new debt below the amount the execution path actually borrows. Internal Euler migration applies the SDK's 0.01% interest cushion; external same-asset migration uses the page's 1% buffered amount. The projected utilization, debt value, LTV/health, and liquidity/cap gate use unbuffered current debt instead.
2. The collateral form keeps the previous current-yield context while a replacement baseline is loading, and preserves it if the new baseline read rejects. A later projected estimate can therefore pair its after-state with an old before-state for the same position/vault key.

### Validation

- 32 focused tests across repay swap/collateral/savings, refinance/swap helpers, borrow, and multiply — passed
- `npm run build` — passed (existing bundler warnings only)
- Hosted preview headed-browser pass at desktop and 390×844 mobile — the app shell loaded, but the chosen spy fixtures did not hydrate the forms, so interaction smoke is incomplete and no screenshots were posted
- Hosted CI lint/test/typecheck and Railway checks are green

### Bot feedback and hygiene

The current CodeRabbit race/after-state concerns were checked against the latest code; the mode-aware repay input and multiply race guards are now present. I removed one stale Leonard inline comment that described the pre-fix repay input path.

Scalability/maintainability pass: the shared projection/rate utilities are a good direction, but async projected-yield orchestration remains duplicated across several form/page watchers. In this review, the concrete correctness consequence is the baseline lifecycle gap called out inline; a shared latest-estimate runner would reduce this class of sibling drift.

Smoke coverage: incomplete headed browser visual smoke; no wallet/signing smoke.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `8b91b983`. The latest commit tightens projected-yield label spacing, but it does not change the two blocking paths already identified on the prior head.

### Blocking findings

1. **Same-asset refinance projections still understate the debt opened by execution.** `nextDebtAmountNano` returns unbuffered `currentDebt` for the same-asset path. Internal Euler migration applies the SDK's 0.01% interest cushion, while external same-asset migration uses the page's 1% buffered amount. Projected utilization, debt value, APY/ROE, LTV/health, and liquidity/cap checks therefore model less target debt than the plan can borrow. This is most consequential near liquidity, cap, or health boundaries. The existing refinance tests cover request merging, not execution-sized same-asset debt.

2. **The collateral form can pair a new projected estimate with a stale current-yield baseline.** The baseline watcher advances its guards and clears the projected output, but leaves the previous `netAPY` / `currentYieldContext` installed while the replacement snapshot is loading. The baseline `Promise.all` also has no active-generation failure reset, so rejection preserves the old context indefinitely. A later estimate for the same position/vault key can then use fresh after-state data against an older before-state. Clear the baseline before the await and on active-generation failure, and add a deferred/rejected baseline regression.

The existing Leonard inline threads remain accurate coordinates for these two findings, so I retained them rather than adding duplicates.

### Validation

- Focused Vitest: 4 files, 21 tests passed (`useCollateralForm`, `refinance-apy`, `useRepaySwapCore`, `useCollateralSwapRepay`).
- `npm run typecheck` passed.
- `git diff --check origin/development...HEAD` passed.
- Hosted CI lint, typecheck, tests, CodeRabbit, and Railway deployment are green.
- Headed Chromium interaction smoke on the current Railway preview at 1365×900 and 390×844:
  - exercised collateral repay with a 10 USDC debt target rather than stopping at route load;
  - quote, projected ROE, projected market rates, LTV/health transitions, and contribution breakdown rendered;
  - the 24×24 breakdown trigger and modal rendered without horizontal overflow, clipping, or material page errors on desktop/mobile.
- Wallet/signing smoke was not run.

The shared projected-yield summary and snapshot/completeness utilities are the right direction, and the latest spacing fix correctly lands in the shared row. The remaining drift is in orchestration semantics: refinance has a separate notion of projected debt sizing from execution, while collateral-form baseline lifecycle is managed by another bespoke async watcher. Execution-sized amount derivation and latest-estimate/baseline invalidation need reusable invariants plus focused failure/race tests, rather than route-local patches.

### Reviewer-bot feedback

The visible CodeRabbit threads describe earlier-head APY/race issues that are resolved or obsolete as written on this head; I did not repeat them as current findings.

<details>
<summary>Validated desktop/mobile browser evidence</summary>

Desktop collateral repay — Projected ROE breakdown:

Desktop collateral repay — Projected ROE breakdown

Mobile collateral repay — Projected ROE breakdown (390×844):

Mobile collateral repay — Projected ROE breakdown

</details>

Smoke coverage: headed desktop/mobile interaction smoke; no wallet/signing coverage.

LeonardEulerXYZ (commented):
## Current-head incremental review — prior blockers resolved

Reviewed head `514b2997` against the two blockers from `8b91b983`, the affected sibling flows, active reviewer feedback, and the current hosted preview.

### Verdict

No remaining blocker found in the latest delta.

- Same-asset refinance projections now use the same execution-sized debt as the plan: the SDK 0.01% interest cushion for internal Euler refinance and the explicit 1% buffer for external migration. The buffered amount feeds projected utilization/rates, debt value, LTV/health, and liquidity/cap checks.
- The collateral-form baseline watcher now clears the prior baseline/context before awaiting replacement, invalidates pending projected estimates, and keeps both before/after state unavailable on active-generation failure. The new deferred/rejected regressions cover the stale pairing path.
- The repay-page baseline moved into a guarded `useRepayNetApy` composable with stale-completion and rejection coverage.

### Validation

- Focused Vitest: 5 files, 27 tests passed (`useCollateralForm`, `useRepayNetApy`, `refinance-apy`, `useRepaySwapCore`, `useCollateralSwapRepay`).
- `npm run typecheck` passed.
- `git diff --check origin/development...HEAD` passed.
- Hosted CI lint, typecheck, tests, CodeRabbit, and Railway deployment are green.
- Headed Chromium interaction smoke on the current Railway preview at 1365×900 and 390×844:
  - exercised `/position/4/repay?network=1` with **From collateral** and a 10 USDC debt target;
  - live quotes loaded and ROE, LTV, health, liquidation-price/buffer, and swap amounts updated without material page errors;
  - wallet/signing smoke was not run.
- Screenshots were not captured or posted for this incremental pass: the latest commit changes projection sizing and async baseline lifecycle, not layout; the immediately preceding review already contains desktop/mobile modal evidence for the same UI surface.

The new refinance amount helper and repay baseline composable are good shared abstractions. Two non-blocking gaps remain:

- projected-yield campaign/rate-line assembly is still duplicated across collateral, multiply, borrow, and repay flows; the local collateral builder includes unchanged rate rows while the shared helper filters them, so sibling display semantics can drift;
- the former refinance blocker is tested at helper level, but there is no page-boundary regression proving internal same-asset, external same-asset, and cross-asset route selection feed the correct amount into `nextDebtAmountNano`.

### Reviewer feedback disposition

The visible CodeRabbit comments describe earlier-head APY/race issues and are resolved or obsolete as written on this head. I did not repeat them. The two prior Leonard inline blockers are fixed and were removed rather than left as stale instructions.

Smoke coverage: headed desktop/mobile interaction smoke; no wallet/signing coverage.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `57b08ff7` across utilization-aware APY/ROE math, missing/partial data handling, async invalidation, sibling borrow/multiply/collateral/repay/lend/refinance surfaces, active bot feedback, and the hosted preview.

### Blocking finding

A nonzero debt whose off-chain USD price is unavailable is converted to a valid `$0` debt value in the new repay Net APY baseline. The estimate is then published as though the position had no borrowing cost rather than remaining unavailable. See the inline finding.

This sentinel is also used by several of the new projected-value paths, so the fix should be applied consistently: zero token amount may legitimately map to `$0`; a positive amount with no price must keep the estimate incomplete.

A local throwaway regression reproduced the issue: with positive debt and the USD helper resolving unavailable-as-zero, `useRepayNetApy` published `5` instead of `null`.

### Bot-feedback disposition

Several active CodeRabbit threads are stale on this head and should not be treated as current blockers:
- collateral-funded repay does propagate `nextSnapshot.weightedSupplyApy` into `roeAfter`;
- projected borrow APY and existing-position multiply snapshots now have generation guards and fail-closed completeness gates;
- `getCollateralApySnapshot()` catches read failures and returns an incomplete snapshot.

The shared `getProjectedYieldState`, collateral snapshot, projected-rate completeness, and race-guard primitives are the right abstraction direction, and the composable tests cover many stale-result and partial-rate cases. Remaining nonblocking hygiene gap: substantial page-local orchestration in existing-position multiply, refinance, and the three lend pages lacks focused consumer tests, while lend projected-supply detail assembly is repeated across those pages. Extracting the common page adapter and testing the page consumers would reduce drift between sibling flows.

### Validation

- `npm run test:run -- <12 focused APY/ROE files>` — 12 files, 81 tests passed
- Hosted preview, headed Chromium: desktop borrow and multiply plus mobile repay rendered; borrow/repay inputs were exercised; mobile width had no horizontal overflow
- Wallet/signing smoke: not run
- Screenshots: not posted because the available live fixtures did not render a valid projected-breakdown state; posting those captures would overstate the tested surface

No malicious-code, dependency, workflow, secret, or transaction-construction changes were found in the reviewed diff.

LeonardEulerXYZ (changes requested):
## Current-head incremental review — changes requested

Reviewed `47538ea71a160294aed3bd35f53d97d3c66dcdfd`.

The previous blocker is resolved for an absent price: the new estimate helper keeps positive unpriced debt unavailable, and the affected composables now clear their metrics instead of treating that debt as `$0`. The shared helper is the right maintainable abstraction.

One fail-closed boundary remains: for a positive amount, an explicit zero `marketPriceUsd` still passes the helper as a complete `$0` valuation. That can publish Net APY/ROE while omitting a real collateral or debt leg. See the inline finding.

- Good: the fix is centralized in `getAssetUsdValueForEstimate` and reused across the changed borrow, multiply, collateral, and repay consumers rather than duplicated locally.
- Gap: the positive-amount/zero-price boundary is not covered by the focused utility test. The route-local borrow-more implementation also lacks a focused missing-price regression test; its sibling composables have coverage.

### Bot feedback
Current CodeRabbit completed successfully. I rechecked the historical APY/ROE comments against this head; they are stale/addressed and do not add a current blocker.

### Validation
- `npm run test:run -- tests/utils/sdk-prices.test.ts tests/composables/useRepayNetApy.test.ts tests/composables/useBorrowForm.test.ts tests/composables/useMultiplyForm.test.ts tests/composables/useCollateralForm.test.ts tests/composables/useWalletRepay.test.ts tests/composables/useWalletSwapRepay.test.ts` — 53/53 passed
- `git diff --check 57b08ff..47538ea` — passed
- Hosted lint, test, typecheck, CodeRabbit, and Railway deploy checks — passed
- Headed Chromium preview smoke on the VOLONEW/USDC borrow form at desktop `1440x1000` and mobile `390x844`: route and inputs rendered, form interaction produced the expected disconnected-wallet balance guard, no app error shell, and no horizontal overflow. The missing-price branch was not visually reproducible with this live fixture.

Smoke coverage: browser interaction + mobile; no wallet/signing. Screenshots were not posted because this is a data-readiness edge and the live fixture did not produce useful visual evidence of it.

LeonardEulerXYZ (commented):
## Current-head review — no remaining blocker found

Reviewed head `920ebd2c` against `development`, including the latest zero-price fix, utilization-aware APY/ROE math, async invalidation/failure behavior, reward enrichment, sibling borrow/multiply/collateral/repay/lend/refinance surfaces, and active reviewer feedback.

### Verdict

No blocking correctness, security, or transaction-boundary finding remains on this head.

The previous positive-amount/zero-price blocker is resolved in the shared estimate-value helper and covered by the missing regression. I also traced the two remaining risk candidates:

- reward campaign/registry enrichment remains reactive: snapshot consumers read the reward version before their first await, while synchronous projected-yield computeds track the reward settings, eligibility address, and reactive vault registry through the reward getters;
- snapshot/rate projection failures now fail closed behind completeness gates and race guards. The uncaught off-chain USD watcher shape predates this PR; the changed paths clear the published estimate before awaiting, and the off-chain helper does not perform an RPC/backend request.

### Validation

- Focused Vitest covering projected yield, vault APY/rates, SDK price readiness, repay metrics, borrow/multiply/collateral, and wallet/collateral repay paths — passed.
- `npm run build` — passed.
- `git diff --check origin/development...HEAD` — passed.
- Headed Chromium interaction smoke on the Railway preview:
  - `/position/1/repay?network=1&spy=<fixture>` at `1440×1000` and `390×844`;
  - entered a 100 USDC wallet repayment without submitting a transaction;
  - Projected Net APY updated `-1.52% → -1.45%`, liquidation buffer `18.61% → 18.70%`, and LTV `65.11% → 64.80%`;
  - the projected-rate/contribution breakdown opened and remained readable at both viewports, with no horizontal overflow.
- Wallet/signing smoke: not run.

The shared projected-yield state/details, collateral snapshot, projected-rate batching/completeness, estimate-value, and race-guard utilities are the right abstraction direction and are used across the affected sibling flows. A non-blocking residual gap remains: campaign/rate-line assembly and async orchestration are still repeated in several page/composable adapters, so future work should continue converging those adapters and add page-boundary regressions for the highest-risk refinance/multiply routes.

### Reviewer-feedback disposition

The visible CodeRabbit findings about missing post-repay weighted APY, projected-borrow race guards, snapshot failure fallback, same-asset refinance handling, and multiply snapshot guards describe earlier heads; the current code addresses them. I did not repeat them as findings.

<details>
<summary>Validated desktop/mobile evidence</summary>

Desktop repay — Projected Net APY breakdown (`1440×1000`):

Desktop repay — Projected Net APY breakdown

Mobile repay — Projected Net APY breakdown (`390×844`):

Mobile repay — Projected Net APY breakdown

</details>

Smoke coverage: headed desktop/mobile interaction smoke; no wallet/signing coverage.

LeonardEulerXYZ (changes requested):
## Current-head incremental review — changes requested

Reviewed head `33a2218d` against the previously clean `920ebd2c` head, focusing on the two new commits, affected sibling APY/reward paths, active bot feedback, and scalability/maintainability.

### Blocking finding

The latest collateral snapshot change prefers layer-aware position vault state, which is correct when that state is enriched, but it no longer falls back to the registry when a matching `AccountPosition` exists with optional `vault` missing. The snapshot then returns incomplete even though the vault is resolvable, hiding projected Net APY/ROE across consumers. See the inline finding.

A local throwaway regression with an unenriched primary collateral and a resolvable registry vault failed (`snapshot.isComplete` was `false`); it was removed after validation.

### Validation

- Focused Vitest for the changed composable/reward/APY utilities: 4 files, 25 tests passed.
- `npm run typecheck` passed.
- `git diff --check origin/development...HEAD` passed.
- Hosted lint, test, typecheck, CodeRabbit, and Railway checks passed.
- Headed Chromium at 1440×1000 and 390×844 loaded the current Railway app without page exceptions or horizontal overflow, but the two bounded spy fixtures no longer hydrated actionable repay/refinance forms. Interaction smoke is therefore incomplete; no screenshots were posted because the captures did not show the changed state.
- Wallet/signing smoke was not run.

The layer-aware snapshot state and shared total-supply helper are sensible abstractions. Two non-blocking gaps remain: total supply APY composition and projected supply-detail assembly are still duplicated across sibling lend flows, and the snapshot resolves collateral entries with repeated linear scans. The immediate correctness fix should preserve the simulated vault when present while using the registry as fallback when `existingCollateral.vault` is absent.

### Reviewer feedback disposition

The visible CodeRabbit APY/race findings describe earlier heads: current repay snapshots, race guards, failure fallback, same-asset refinance, and multiply guards address them. I did not repeat them as current findings.

Smoke coverage: incomplete headed desktop/mobile route smoke; no wallet/signing coverage.

Comment lifecycle: no Leonard top-level walkthrough or inline comments found to edit/delete on this head.

LeonardEulerXYZ (changes requested):
Reviewed the incremental fix from `33a2218` through `8afb24e` and rechecked the full affected APY/ROE projection path. The earlier shared collateral-snapshot registry fallback is fixed, and the focused layered-vault/APY/batch/form tests now pass. Two remaining stale-state boundaries can still present incorrect projected APY/ROE, so I’m keeping this blocked.

Validation:
- Focused Vitest: 6 files, 59 tests passed (`useLayeredVaults`, `usePositionCollateralApy`, `useTxBatch`, refinance APY, borrow form, multiply form)
- `npm run typecheck` passed
- `npm run build` passed
- `git diff --check origin/development...HEAD` passed
- Desktop/mobile Railway preview loaded, but the available fixtures did not reach a usable refinance interaction state; no screenshots are attached because they would not prove the changed behavior

Scalability / maintainability pass: the shared `useLayeredVaults` and collateral-APY abstractions are the right direction and sibling borrow/multiply coverage is present. The remaining refinance source-leg path still performs a pre-fallback shape gate, and the shared batch overlay remains valid after a failed replacement simulation. Both are shared correctness boundaries rather than formatting-only concerns.

Security/supply-chain pass: no dependency, lockfile, workflow, shell-script, credential, or new external-call surface was introduced in the incremental delta.

Bot feedback: I checked the active automated feedback against the current head; I found no separate material current-head bot finding to carry forward beyond the two issues below.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `7ca8cc5d` across utilization-aware APY/ROE math, same-vault cash/borrow merging, batch-layer propagation, sibling borrow/multiply/collateral/repay/refinance/lend consumers, active bot feedback, and scalability/maintainability.

### Blocking finding

The new authoritative `simulatedVaultsLayers` path accepts every non-empty cardinality except the exact `plans.length` normalization case. For a multi-operation batch, a final-only or partial vault-layer response is therefore indexed from layer 0 and cumulatively propagated into earlier layers. Those earlier forms can display future vault utilization and projected APY/ROE. The account-layer path already rejects insufficient cardinality; the vault-layer path should likewise accept only a proven shape (`0`, `plans.length`, or `plans.length + 1`) and otherwise ignore/fail closed rather than shift snapshots onto the wrong operations. See the inline comment.

### Verified

- The latest same-address projection merge is coherent: cash and borrow deltas are combined atomically and inconsistent base states fail closed.
- Borrow, multiply, collateral, repay, and refinance projection consumers now separate current/projected values and use generation guards or safe clearing on the paths checked.
- The two prior Leonard blockers are resolved: refinance applies layered fallback before the shape gate, and a failed current replacement simulation invalidates the prior account/vault overlay. I removed those two obsolete Leonard inline comments.
- Active CodeRabbit APY/race comments were checked against current code; the material claims are addressed or obsolete and were not repeated as Leonard findings.
- Whole-diff security/supply-chain pass found no dependency, lockfile, workflow, secret, environment, signing, or transaction-plan changes requiring separate escalation.

The core projection math is appropriately centralized and has focused utility/composable tests. One non-blocking follow-up remains: projected-supply detail assembly is still duplicated across the three lend pages, while collateral/multiply rate-line adapters partly duplicate `getCollateralSnapshotRateLines`. A shared configurable adapter plus page-boundary tests would reduce drift in before/after campaign and unchanged-row semantics.

### Validation

- Hosted CI on this head: lint, typecheck, and test passed.
- `git diff --check origin/development...HEAD` passed.
- Focused local tests were not independently rerun: the PR worktree's `package.json` / `package-lock.json` are not `npm ci`-compatible (`cac`, `commander`, and `crossws` missing from the lock), so the clean managed bootstrap stopped before installing dependencies.
- Hosted preview was probed on desktop using the documented spy fixtures, but the borrow/refinance routes rendered only their page shells and no actionable APY/ROE form state. Browser interaction smoke is therefore incomplete; mobile and wallet/signing smoke were not run. No screenshots were posted because there was no valid changed state to evidence.

LeonardEulerXYZ (commented):
## Current-head incremental review — prior blocker resolved

Reviewed head `d0d2f4c2` against the prior blocked head `7ca8cc5d`, then rechecked the affected APY/ROE projection path, sibling consumers, active reviewer feedback, and scalability/maintainability.

### Verdict

No remaining blocking correctness, security, or transaction-boundary finding was found on this head.

The latest commit fixes the prior multi-operation batch-layer issue at the shared boundary: simulated vault snapshots now accept only the three proven cardinalities (absent, one per operation, or base-inclusive). Ambiguous final-only/partial shapes are discarded instead of being shifted onto earlier operations. The new regression proves that a final-only two-operation response cannot leak future vault utilization into earlier projected APY/ROE layers.

The incremental delta changes projection-layer validation and tests only; transaction construction is unchanged. The broader projection consumers checked—borrow, multiply, collateral, repay, refinance, and lend—continue to consume the shared layered state rather than applying a route-local workaround.

### Validation

- Focused Vitest: 5 files, 54 tests passed (`useTxBatch`, wallet repay, collateral APY snapshots, projected yield, and projected rates).
- `npm run typecheck` passed.
- `git diff --check origin/development...HEAD` passed.
- Hosted lint, test, typecheck, CodeRabbit, and Railway deployment checks are green.
- Browser smoke: not completed. The review runner had no usable Chromium binary and the bounded browser installation did not complete, so I did not claim desktop/mobile interaction coverage or post screenshots. Wallet/signing smoke was not run.

The fix is correctly centralized in `normalizeSimulatedVaultLayers` and tested at both utility and `useTxBatch` integration levels, so every APY/ROE form using layered vault state inherits the same fail-closed behavior. The existing non-blocking follow-up remains: projected-supply detail assembly is duplicated across the three lend pages, and collateral/multiply rate-line adapters still partly duplicate the shared snapshot adapter. Consolidating those adapters with page-boundary tests would reduce future sibling drift.

### Reviewer-feedback disposition

The active CodeRabbit APY/race comments describe older heads; their material concerns are addressed or obsolete in the current code and were not repeated. The prior Leonard inline finding about ambiguous vault-layer cardinality is fixed and was removed rather than left as stale guidance.

Smoke coverage: focused code/tests/typecheck; browser/mobile and wallet/signing smoke not completed.

LeonardEulerXYZ (changes requested):
## Current-head review — changes requested

Reviewed head `a7690de4`, including the incremental layered-collateral change since the previously clean head, utilization-aware APY/ROE math, exact amount and completeness boundaries, async invalidation, sibling form consumers, active reviewer feedback, and scalability/maintainability.

### Blocking finding
Switching directly between two savings collateral sub-accounts does not invalidate or rerun the borrow projection. The new calculation reads `savingCollateral.value?.subAccount` to decide self-transfer treatment and reward/weighted-collateral inputs, but the invalidation watches only `isSavingCollateral`. A savings-A → savings-B selection leaves that boolean `true`, so the already-published Net APY can continue to describe the old source account.

A local throwaway regression selected a second savings sub-account after the first estimate settled; no new valuation/projection call occurred, and the test failed. Please make the selected savings source part of the synchronous invalidation/rerun dependencies and keep the source-switch regression.

### Other review notes
- The incremental change otherwise correctly incorporates the pending layered sub-account, existing enabled collateral, existing debt, self-transfer deduplication, asset decimals, projected-rate completeness, and weighted rewards.
- Scalability/maintainability: the implementation continues to use the shared projection/reward helpers, and I found no additional missed sibling flow in repay, refinance, collateral swap, multiply, or lend. The missing source dependency is local to the borrow-form projection lifecycle.
- Active CodeRabbit findings inspected are addressed or stale on this head: repay before/after weighting, stale async guards, incomplete-snapshot fallback, same-asset refinance, and multiply snapshot races now have the corresponding current-head handling.

### Validation
- Focused APY/ROE/form suite: 27 files, 251 tests passed.
- `npm run typecheck` passed.
- Local throwaway source-switch regression: failed as described above; removed after validation.
- Headed hosted-preview interaction smoke: borrow form rendered and accepted inputs at desktop and narrow mobile viewports; Net APY/LTV/health summary remained present. This disconnected smoke could not exercise wallet savings-sub-account switching.
- Clean install reproducibility remains blocked by a pre-existing lockfile mismatch; package manifests are unchanged by this PR.

Smoke coverage: headed desktop/mobile browser interaction; no wallet/signing smoke. No screenshots posted—the blocker is account-state reactivity rather than a useful disconnected visual delta.

LeonardEulerXYZ (changes requested):
1|## Current-head review — changes requested
     2|
     3|Reviewed head `e07b4209`, including the incremental borrow-form fix, the utilization-aware APY/ROE paths, sibling repay/swap flows, active reviewer feedback, and scalability/maintainability.
     4|
     5|### Prior blocker resolved
     6|
     7|The latest commit correctly synchronizes `selectedSavingSubAccount` before the borrow projection runs. The focused borrow regression passes, and I found no stale borrow-form estimate on the fixed path.
     8|
     9|### Blocking finding
    10|
    11|The savings-repay sibling flow can still retain an account-sensitive quote after switching between two sub-accounts in the same savings vault. `getQuoteAccounts()` derives `accountIn` from `selectedSavingSubAccount`, but the shared quote-invalidation watcher observes the source vault, balance, shares, and slippage—not source account identity. If positions A and B hold equal assets in the same vault, switching A → B changes none of those dependencies, so neither quote engine resets.
    12|
    13|The plan builder then uses the newly selected B as `fromAccount` while `selectedQuote` can still be the quote fetched for A. That can leave the new utilization-aware `spent` / `debtRepaid` / APY-after state based on the old quote and can carry an account-context mismatch into planning or simulation.
    14|
    15|A local throwaway regression with equal-balance A/B positions reproduced this deterministically: the selected sub-account changed, but both quote-engine reset spies remained untouched. Please make source account identity part of quote invalidation and add the same-vault/different-sub-account regression.
    16|
    17|### Scalability / maintainability hygiene
    18|
    19|The projection, reward, completeness, and race-guard logic is mostly centralized and reused across borrow, multiply, collateral, repay, refinance, and lend surfaces. The remaining gap is at the shared repay-swap boundary: account identity is part of the quote key, but is not modeled as part of the shared invalidation key. Fixing this centrally is preferable to a savings-only UI workaround.
    20|
    21|### Reviewer-feedback disposition
    22|
    23|The visible CodeRabbit APY/race findings describe earlier heads and are addressed or obsolete in the current implementation; I did not repeat them as current findings. The prior Leonard borrow-form inline finding is fixed and was removed rather than left stale.
    24|
    25|### Validation
    26|
    27|- Focused Vitest: 3 files, 27 tests passed (`useBorrowForm`, `useSavingsRepay`, `useRepaySwapCore`).
    28|- `npm run typecheck` passed.
    29|- `git diff --check origin/development...HEAD` passed.
    30|- Local throwaway same-vault/different-sub-account regression failed at the missing quote reset as described, then was removed; the worktree is clean.
    31|- Headed Railway-preview interaction smoke on the borrow form at `1440×1000` and `390×844`: entered an amount, Net APY/LTV/health summaries remained present, no app error shell or horizontal overflow. The only console error was the preview's existing 403 resource noise.
    32|- Wallet/signing smoke was not run.
    33|- Screenshots were not captured or posted: the blocker requires two same-vault savings sub-accounts and is not a useful disconnected visual delta.
    34|
    35|Smoke coverage: headed desktop/mobile borrow-form interaction; no wallet/signing coverage.
    36|

LeonardEulerXYZ (commented):
## Current-head incremental review — prior blocker resolved

Reviewed head `a1db57a9` against the previously blocked `e07b4209` head, then rechecked the affected utilization-aware APY/ROE paths, sibling form consumers, active reviewer feedback, and the whole diff’s security/maintainability surface.

### Verdict

No remaining blocking correctness, security, or transaction-boundary finding was found on this head.

The latest fix correctly makes quote account identity reactive at the shared repay-swap boundary. Switching between equal-balance savings subaccounts now resets both quote engines and requests a replacement quote with the new `accountIn`; the focused regression covers the exact A→B case. I removed my now-obsolete inline blocker rather than leaving stale guidance.

### Validation

- Current-head hosted CI: lint, typecheck, and full tests passed; CodeRabbit and the Railway deployment are green.
- `git diff --check origin/development...HEAD` passed.
- Direct current-head diff/security sweep: no dependency, lockfile, workflow, environment/config, credential, signing, approval, permit, or new external-call surface was introduced.
- Local clean-install validation was not repeated because the unchanged repository manifests currently fail `npm ci` consistency checks (`cac`, `commander`, and `crossws` are absent from the lockfile). I did not mutate the lockfile to work around that baseline condition.
- Headed Chromium loaded the current preview at desktop `1440×1000` and mobile `390×844` without an app error shell, material page exception, or horizontal overflow. The bounded live refinance fixtures no longer hydrated an actionable APY/ROE form, so interaction coverage on this incremental head is incomplete. The immediately preceding reviewed head already had desktop/mobile form interaction coverage; this final delta changes account-sensitive quote invalidation and its test, not layout.
- Wallet/signing smoke was not run.
- Screenshots were not captured or posted because the available current-head fixtures did not show a valid changed interaction state; route-shell captures would overstate the evidence.

The fix is correctly centralized in `useRepaySwapCore`, so collateral and savings repay consumers inherit the same account-identity invalidation instead of receiving a route-local patch. Shared projected-yield, collateral snapshot, rate-completeness, estimate-value, and race-guard utilities now cover the main borrow, multiply, collateral, repay, refinance, and lend flows checked.

Non-blocking follow-up: substantial orchestration remains duplicated across wallet vs wallet-swap repay and the lend/refinance page adapters, while the largest refinance page still lacks a focused page-boundary reactivity test. Continuing to consolidate those adapters would reduce sibling drift.

### Reviewer-feedback disposition

The visible CodeRabbit comments concern earlier-head before/after APY, race-guard, snapshot-failure, and same-asset refinance paths. Those claims are addressed or obsolete in the current implementation; I did not repeat them as current findings.

Smoke coverage: current-head desktop/mobile route smoke was incomplete; prior-head desktop/mobile form interaction evidence remains applicable to the unchanged UI surface; no wallet/signing coverage.

kanvgupta (approved):

PR #374, opened 2026-05-07
fix: avoid negatively caching transport errors in pricing resolver

kasperpawlowski:
## Summary

Single transport-level RPC failures (timeout, HTTP 5xx, rate limit) were poisoning the asset price cache in `entities/vault/pricing.ts`: the `.catch` handler stored `null` for the asset, and since `clearPriceCaches()` only fires on chain switch, the next caller saw a permanent miss for the rest of the process lifetime — even after the RPC fully recovered. On the warm-cache server process this means a single Ankr blip can blank a vault's USD value until the next deploy.

The fix walks the cause chain via `summarizeViemError` and only writes to the negative cache when the failure is *not* a transport error. Genuine "no oracle for this pair" signals (the lens returning `queryFailure: true`, plus contract reverts) still cache as before — the in-flight dedup map prevents thundering-herd retries during a sustained outage.

Real-world trigger that surfaced this: wstETH on TAC (chain 239) hit a transient `code: -32062` from `rpc.ankr.com/tac`. The same call works now in 450ms, but the cached `null` would have kept that asset unpriced until the Fargate task restarted.

## Test plan

- [x] New unit tests in `tests/entities/vault/pricing.test.ts` covering: success cache, `queryFailure: true` cache, transport-error skip-cache (HTTP 5xx + timeout), non-transport-error cache, and concurrent in-flight dedup during a transport failure.
- [x] Full vitest suite: 627 passing, 1 skipped (no regressions).
- [ ] Manual: monitor BetterStack for `ctx: pricing/resolveAssetPrice` warnings after deploy — single-asset bursts should now self-recover instead of going silent at one warning per process.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved price-resolution caching: transient transport/RPC errors are no longer permanently cached, allowing retries; contract/query failures remain cached to avoid redundant lookups. Unit-of-account cache updates now occur only when a valid price is available or an existing asset entry exists, preventing unnecessary null writes.

* **Tests**
  * Expanded coverage for caching and error-handling: transient failures, timeouts, query failures, cache invalidation, and concurrent requests.

### DIFF
--- a/entities/vault/pricing.ts
+++ b/entities/vault/pricing.ts
@@ -153,7 +153,9 @@ export const resolveUnitOfAccountPriceInfo = async (
   }
 
   const priceInfo = await resolveAssetPriceInfo(rpcUrl, utilsLensAddress, unitOfAccount)
-  unitOfAccountPriceCache.set(normalized, priceInfo || null)
+  if (priceInfo || assetPriceCache.has(normalized)) {
+    unitOfAccountPriceCache.set(normalized, priceInfo || null)
+  }
   return priceInfo
 }
 
--- a/entities/vault/pricing.ts
+++ b/entities/vault/pricing.ts
@@ -152,8 +152,9 @@ export const resolveUnitOfAccountPriceInfo = async (
     return undefined
   }
 
+  const gen = cacheGeneration
   const priceInfo = await resolveAssetPriceInfo(rpcUrl, utilsLensAddress, unitOfAccount)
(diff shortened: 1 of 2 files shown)


Seranged (approved):
Added two follow-up fixes:

1. Stopped unitOfAccountPriceCache from caching null when the underlying failure was a transient RPC/transport error.

2. Added a cacheGeneration guard so an old in-flight unit-of-account price request cannot write stale data after clearPriceCaches() runs during a chain switch.

PR #376, opened 2026-05-07
Refactor info triggers to use hover modals

Seranged:
## Summary
- Refactor informational metric triggers to share the modal preview UX across vault lists, overview pages, position pages, and discovery surfaces.
- Keep click/tap behavior available for the full modal experience while hover-capable devices can preview the same breakdown content inline.
- Keep badge/chip labels, generic footnotes, and full-card overview opens on their existing click or tooltip behavior.

## Changes
- Applies `UiModalPreviewTrigger` to APY, rewards, Net APY, ROE, collateral exposure, hooks, and ramp-down info affordances across lend, borrow, earn, portfolio, position, securitize, and discovery views.
- Moves trigger-specific modal props into local computed data helpers so hover, click, and keyboard activation open the same modal content without leaking events into parent cards.
- Adds keyboard access to the preview trigger with button semantics, Enter/Space activation, and a visible focus state.
- Preserves existing `UiFootnote` behavior for compact explanatory notes and keeps badge/chip informational labels click-only.

## Test plan
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `git diff --check`
- [x] Smoke tested `/lend?network=1` and `/explore?network=1` locally with no console errors.
- [x] Smoke tested a migrated Supply APY trigger opening its modal without navigating the parent vault card.
- [x] Manually reviewed badge/chip, points, footnote, and portfolio full-card surfaces to keep only the metric-style affordances in the shared preview flow.


kasperpawlowski (dismissed):
I am wondering if there's a better way to do this. the problem is that every time the modal opens, user needs to click the close button or click away. it's a bit suboptimal UX and I don't want to make it worse than it is

Seranged:
Yeah there definitely is, I have a idea but is more of a rework, will scope out a proposition

kasperpawlowski (approved):

PR #382, opened 2026-05-07
Hide static share token exchange rates

Seranged:
## Summary
- Hide the share token exchange rate row when an EVK vault cannot accrue depositor interest, keeping the risk parameters block focused on useful values.
- Avoid the `convertToAssets` read for those static-rate vaults instead of fetching the value and hiding it later.

## Changes
- Adds a computed display predicate for escrow vaults and collateral-only EVK vaults.
- Reuses that predicate to gate both the row rendering and the contract read.
- Leaves Securitize vault overview behavior unchanged.

## Test plan
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] Checked a standard borrowable vault keeps showing the share token exchange rate row.
- [x] Checked a collateral-only EVK vault hides the row.
- [x] Checked an escrow vault hides the row.
- [x] Checked the escrow page RPC payloads did not include the `convertToAssets` selector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Share token exchange rate now only appears on vault overview when applicable, preventing misleading or irrelevant information.
* **Performance**
  * Reduced background data requests for vaults where the exchange rate is not applicable, improving responsiveness and efficiency.

### DIFF
--- a/components/entities/vault/overview/VaultOverviewBlockRiskParameters.vue
+++ b/components/entities/vault/overview/VaultOverviewBlockRiskParameters.vue
@@ -32,6 +32,7 @@ const isBorrowable = computed(() => borrowCount.value > 0)
 
 const supplyCapPercentageDisplay = computed(() => getSupplyCapPercentage(vault))
 const borrowCapPercentageDisplay = computed(() => getBorrowCapPercentage(vault))
+const showShareTokenExchangeRate = computed(() => vault.vaultCategory !== 'escrow' && vault.borrowCap !== 0n)
 
 const supplyCapDisplay = ref('-')
 const borrowCapDisplay = ref('-')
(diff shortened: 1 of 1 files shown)


kasperpawlowski:
`borrowCap === 0n` in EVK means "no new borrowing allowed", not "no interest accruing". A vault that was paused mid-life by setting borrowCap to 0 but still has outstanding debt (`vault.borrow > 0n`) keeps accruing interest on that debt, so the share-token exchange rate is still ticking up and is genuinely useful to surface. The current gate hides it in that state.

The codebase already recognises that configuration as a real wind-down state — `utils/discoveryCalculations.ts:934` flags `borrowCap === 0n && vault.borrow > 0n` as the "exceeded" cap-usage case for the discovery table.

Extending the predicate to keep the row visible while debt exists fixes this without needing a contract read:

`const showShareTokenExchangeRate = computed(() => vault.vaultCategory !== 'escrow' && (vault.borrowCap !== 0n || vault.borrow > 0n))`

kasperpawlowski (approved):

PR #446, opened 2026-05-14
refactor: stable slippage default

Seranged:
## Summary
- Stable-to-stable swap contexts use the `0.05%` default slippage in the settings modal and transaction slippage state.
- Slippage remains available from global user settings and per-flow slippage modals; high custom overrides apply within the pair-default bucket where they were set.
- Overrides at or below the active default remain usable across contexts; stale, future-dated, invalid, or pair-default mismatched high overrides fall back to the active default.
- Existing `swap-slippage*` storage keys hydrate into the explicit override shape for compatibility and are cleared after migration/reset so old values do not resurrect.
- Embedded borrow/pay-with and position supply/withdraw swap flows pass their selected swap token pair into the slippage context.
- The active custom slippage chip uses the active-chip text color for both `Custom` and the value.

## Slippage Product Behavior

Swap slippage defaults follow the selected swap token pair:

- **Stable-to-stable pairs** (for example, `USDC -> RLUSD`) default to `0.05%`.
- **Other pairs** (for example, `wstETH -> WETH`) default to `0.3%`.

The slippage control is globally accessible from user settings and is also available inside swap flows. Custom slippage is stored as a user override with the pair-default bucket where it was set:

- A high custom value such as `3%` applies within the same default bucket where the user set it.
- A `3%` override set from global settings or a normal `0.3%` pair applies to normal `0.3%` pairs, but does not carry into stable-to-stable `0.05%` pairs.
- A `3%` override set from a stable-to-stable `0.05%` flow applies to stable-to-stable pairs, but does not carry into normal `0.3%` pairs.
- High custom overrides reset to the active pair default after 24 hours.
- Lower-than-default overrides such as `0.01%` can carry across pair types and do not expire just because 24 hours passed.

Existing persisted `swap-slippage*` values hydrate into the current override model. Values with a stored stable `0.05%` bucket remain custom on stable pairs, while values without a stored bucket hydrate as normal `0.3%` overrides and fall back to the stable `0.05%` default on stable pairs.

## Changes
- Adds `swap-slippage-override` as the source of truth for persisted custom slippage.
- Uses explicit JSON serialization for the nullable override object.
- Computes effective slippage from the active swap pair default plus the stored override, without writing default/effective values back as custom overrides.
- Clears migrated legacy keys when the override is persisted, reset, expired, invalid, or pair-default mismatched.
- Keeps context-mismatched stable overrides from being cleared before the swap context is available.
- Keeps lower-than-default overrides active even after 24h, while high overrides expire after 24h within their pair-default context.
- Expands `useSlippage` coverage for stable/non-stable pairs, override context changes, existing storage hydration, stale/future timestamps, invalid overrides, and clearing custom values equal to the active default.

## Test plan
- [x] `npm run test:run -- tests/composables/useSlippage.test.ts tests/utils/swap-slippage-validation.test.ts` (28 passing tests)
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `git diff --check`
- [x] Local Nuxt slippage settings smoke on `127.0.0.1:3082` with deterministic injected swap contexts:
  - stable context defaults to `0.05%`
  - non-stable context defaults to `0.3%`
  - stable context keeps a stable-context `3%` override and shows `resets to 0.05% default after 24h`
  - non-stable context keeps a generic-context `3%` override and shows `resets to 0.3% default after 24h`
  - stable context rejects a generic-context `3%` override
  - non-stable context rejects a stable-context `3%` override
  - stable and non-stable contexts keep stale `0.01%` lower-than-default overrides
  - stale and future-dated stable-context `3%` overrides fall back to `0.05%`
  - existing stable-context `swap-slippage*` keys hydrate to `Custom 3%`
  - existing generic-context `swap-slippage*` keys fall back to stable `0.05%`
  - migrated legacy overrides are written to `swap-slippage-override` as JSON
  - active `Custom 3%` chip renders matching dark text for label and value

## Summary by CodeRabbit

* **New Features**
  * Slippage settings now adapt more accurately to the current swap context, including borrow, multiply, and collateral flows.
  * Added improved handling for custom slippage values, presets, and override state persistence.

* **Bug Fixes**
  * Fixed slippage overrides resetting and expiring more reliably, including cleanup of outdated saved values.
  * Improved default slippage behavior for stablecoin-related swaps and prevented settings from applying in inactive form modes.


LeonardEulerXYZ (changes requested):
Reviewed head `5837f6b70d87a7a4a72ef221157c7bbc2d33b930`.

Verdict: request changes — I found one user-visible slippage override regression.

Validation run:
- `npm run test:run -- tests/composables/useSlippage.test.ts tests/utils/swap-slippage-validation.test.ts tests/composables/useSavingsRepay.test.ts` — passed, 30 tests
- `npm run lint` — passed
- `npm run build` — passed
- Local headed Playwright smoke against an isolated route-level `useSlippage({ fromSymbol, toSymbol })` + real `SlippageSettings` save flow — failed on the custom override invariant

Smoke coverage: browser interaction smoke, desktop/mobile harness over the slippage settings lifecycle. No screenshots posted because the finding is state/effective-value based rather than visual; the useful evidence is the failed assertion and state dump.

Scalability / maintainability hygiene: the new shared helper direction is good, and the route call sites now consistently pass swap context across swap/supply/withdraw/borrow/multiply/repay-style surfaces. The missed case is still a composition/lifecycle gap: tests cover isolated helper/composable behavior, but not the common mounted combination where a contextual route instance and the settings component's non-contextual instance both observe the shared override.

Bot feedback: CodeRabbit's preset-equivalence comment is a separate minor UI-state nit; I did not treat it as blocking. The blocker below is independently reproduced against the current head.

composables/useSlippage.ts
```diff
@@ -79,52 +229,76 @@ export const useSlippage = (options?: UseSlippageOptions) => {
...
     }
-    if (persistedValue.value !== effective) {
-      persistedValue.value = effective
+
+    if (
+      override.value.defaultSlippageAtSet !== DEFAULT_SLIPPAGE
+      && !hasCompleteSwapContext.value
+    ) return
+
+    if (!isOverrideActive.value) {
```
Seranged, line 268: Fixed in `875d690a`.

Root cause was a per-instance `now` ref: the settings instance advanced its clock when saving the override, but the contextual route instance could still see the new `setAt` as future-dated and clear the shared override. `now` is now shared via `useState("slippage-now")`, and `tests/composables/useSlippage.test.ts` includes `keeps a contextual override saved from a second settings instance` to pin the route+settings lifecycle.

Also handled CodeRabbit’s preset-equivalence note by centralizing `isPresetOrDefault()` in `SlippageSettings.vue` and adding browser coverage for custom input matching preset `0.1%`.

Validation after the fix:
- `npm run test:run` — 113 files / 1094 tests passed
- `npm run build` — passed
- browser slippage modal matrix — 19/19 passed

LeonardEulerXYZ (changes requested):
Reviewed head `875d690ac0555771701b509a66d9a1c21526c543`.

Verdict: request changes — the original `useSlippage` route/settings lifecycle blocker looks fixed, but the new pair-sensitive default exposes a remaining route-integration bug in the refinance page.

Validation run:
- `npm run test:run -- tests/composables/useSlippage.test.ts tests/composables/useEulerAccount.test.ts tests/composables/useEulerSdk.test.ts` — passed, 30 tests.
- Direct code trace of the refinance slippage context against the actual collateral/debt quote request assets.

Prior Leonard blocker status: I rechecked the current `useSlippage` implementation. The shared `useState("slippage-now")`, the incomplete-context guard, and the new `keeps a contextual override saved from a second settings instance` test cover the earlier route+settings override-erasure failure, so I am not reposting that issue.

Blocking finding:

### Refinance debt quotes can choose the wrong default slippage pair

`pages/position/[number]/borrow/swap.vue` still provides one route-level slippage context as:

```ts
fromSymbol: () => sourceCollateralVault.value?.asset.symbol || sourceDebtVault.value?.asset.symbol,
toSymbol: () => targetCollateralVault.value?.asset.symbol || targetDebtVault.value?.asset.symbol,
```

but the debt quote request actually swaps the target debt asset into the source debt asset:

```ts
tokenIn: targetDebtVault.value.asset.address,
tokenOut: sourceDebtVault.value.asset.address,
slippage: slippage.value,
```

Now that this PR makes the default depend on the pair, that mismatch is user-visible. For a debt-only refinance:

- current collateral `USDC`, current debt `WETH`, target debt `USDC` ⇒ context becomes `USDC -> USDC`, so the default is `0.05%`, while the actual debt quote is `USDC -> WETH` and should use the normal `0.3%` default.
- current collateral `WETH`, current debt `USDC`, target debt `USDT` ⇒ context becomes `WETH -> USDT`, so the default is `0.3%`, while the actual debt quote is stable-to-stable `USDT -> USDC` and should use `0.05%`.

The same single-context problem appears when both collateral and debt legs are active: one displayed/default slippage value is reused for both quote requests even if one leg is stable-stable and the other is not.

Please derive the default from the actual active swap leg(s) on this route. A safe rule would be: use the stablecoin default only when every active quote leg that needs a swap is stable-to-stable; otherwise keep the normal `0.3%` default, or split the route into leg-specific slippage values. The invariant is that the slippage default shown/sent for a quote must be based on the assets in that quote, not on whichever collateral/debt symbol happens to win the current fallback order.

Browser smoke: not run — this finding is from code-path validation of the quote request wiring; focused composable tests and typecheck pass.

Seranged:
Addressed Leonard’s refinance slippage blocker in `5e939c81`.

What changed:
- The refinance page now builds slippage context from the actual active quote legs.
- Collateral leg uses source collateral asset -> target collateral asset.
- Debt leg uses target debt asset -> source debt asset, matching the quote request.
- Mixed collateral+debt flows use the stablecoin default only when every active swap leg is stable-to-stable; otherwise they use the normal 0.3% default.

Validation:
- `npm run test:run -- tests/utils/refinance-slippage.test.ts tests/composables/useSlippage.test.ts`
- `npx eslint pages/position/[number]/borrow/swap.vue utils/refinance-slippage.ts tests/utils/refinance-slippage.test.ts`
- `npm run test:run -- tests/composables/useSlippage.test.ts tests/composables/useEulerAccount.test.ts tests/composables/useEulerSdk.test.ts tests/utils/refinance-slippage.test.ts`
- `npm run test:run` (114 files / 1097 tests)
- `npm run typecheck`
- `npm run build`
- Slippage modal browser smoke matrix: 19/19 passed

LeonardEulerXYZ (commented):
Reviewed head `5e939c8108e5c823187121693d0f5da6c44aae5a`.

Verdict: comment — no blocking issues found in this pass. I am not approving automatically.

What I checked:
- Re-read the changed slippage persistence/defaulting path, `SlippageSettings`, borrow/multiply/collateral swap call sites, and the refinance-page slippage context fix.
- Reconstructed the prior refinance invariant: the slippage default must be derived from the actual active quote leg(s), using the stable default only when every active swap leg is stable-to-stable. The new `getRefinanceSlippageContext()` helper and its tests now cover the debt-quote direction and mixed-leg fallback correctly.
- Scalability / maintainability hygiene pass: searched the sibling `useSlippage()` consumers across lend, withdraw, repay, multiply, collateral swap, generic swap-page logic, and refinance. The patch is no longer a one-off page fallback; the refinance multi-leg decision is isolated in a small tested utility, and borrow/multiply tab coexistence is guarded with `enabled`. I did not find a sibling surface still carrying the same pair-default bug.
- Bot/reviewer feedback pass: CodeRabbit’s older preset-equivalence nit is addressed by the shared `isPresetOrDefault()` checks in the modal save paths. I did not find a material active bot finding to amplify.

Validation run:
- `npm run test:run -- tests/composables/useSlippage.test.ts tests/utils/refinance-slippage.test.ts tests/composables/useEulerAccount.test.ts tests/composables/useEulerSdk.test.ts` — passed, 33 tests.
- `npx eslint <changed files>` — passed.
- `npm run build` — passed.
- `npm run lint` repo-wide is still blocked by existing errors outside the PR diff; I did not treat those as PR-specific.
- Light headed Playwright route smoke against the Railway preview: `/` and `/borrow?network=1` loaded as the real app on desktop and mobile viewports, with no page/console errors observed.

Smoke coverage: browser route smoke + mobile route smoke only. I did not run wallet/signing coverage, and I did not capture/post screenshots because this pass did not produce a meaningful visual before/after artifact for the slippage-default logic.

Previous Leonard blockers: resolved in the current head based on code trace plus focused regression coverage.

LeonardEulerXYZ (commented):
Reviewed head `b23a45fa052ffcd285b1f67038d1540318caf372`.

Verdict: comment — no blocking issues found in this pass. I am not approving automatically.

What changed / risk areas reviewed:
- `useSlippage` now derives the default from active swap context: `0.05%` for stablecoin-to-stablecoin contexts, otherwise the normal `0.3%` default.
- High overrides remain pair-context scoped and expire/reset instead of being silently carried into a lower-default pair.
- Borrow, multiply, collateral supply/withdraw, lend/withdraw, generic swap, and refinance call sites now pass context intended to match their actual quote `tokenIn -> tokenOut` direction.
- Refinance context is centralized in `getRefinanceSlippageContext`, including the prior blocker: debt-only refinance now uses target-debt -> source-debt direction, and mixed legs only use the stable default when every active leg is stable-to-stable.

Validation run:
- `npm run test:run -- tests/composables/useSlippage.test.ts tests/utils/refinance-slippage.test.ts tests/composables/useEulerAccount.test.ts tests/composables/useEulerSdk.test.ts` — passed, 34 tests.
- Focused ESLint over the changed slippage files, related call sites, and focused tests — passed. Only a Node experimental CommonJS/ESM warning was emitted.
- `npm run build` — passed. Build emitted existing chunk/dynamic-import size warnings, but no build failure.
- Headed Chromium/Xvfb preview smoke against the Railway PR preview: `/` and `/borrow?network=1` loaded real app content on desktop `1280x900` and mobile `390x844`. The only console error observed was an analytics fetch failure on `/borrow?network=1`, which is unrelated to this slippage diff and did not block app rendering.

- The slippage defaulting logic is centralized in `useSlippage`, and the refinance-specific multi-leg rule is extracted to `utils/refinance-slippage.ts` with focused tests. That is the right reusable shape; this is not an inline one-component fix.
- I searched the sibling slippage consumers and traced the quote directions for borrow, multiply, collateral supply/withdraw, lend deposit/withdraw, generic swap, and refinance. I did not find another sibling flow still using the old generic-default behavior for a stable-to-stable quote path.
- The `SlippageSettings` preset/default predicate has been consolidated into `isPresetOrDefault`, addressing the prior drift risk around custom input values that match preset/default values.

- CodeRabbit's earlier preset-equivalence comment is fixed by the current `isPresetOrDefault` path.
- The older numeric-constant test nit is non-blocking here; the behavior is now covered by focused stable/default tests and the exported constants remain the source of product defaults.

Smoke coverage: browser visual smoke + mobile smoke for route load/rendering only; no wallet/signing flow. Screenshots were not posted because this PR is a behavioral slippage-default refactor with no intended visual UI delta, and the smoke did not capture a meaningful before/after visual state.

dglowinski (approved):

PR #505, opened 2026-05-27
feat: recognize access-control hook vaults with a distinct chip

kasperpawlowski:
## Summary
Adds first-class recognition of **access-control hook** vaults (e.g. `HookTargetAccessControl` allowlists), handled like the existing keyring/"Private" chip. Driven by the labels `tags` field — a product tagged `access control` surfaces a distinct chip and explanatory copy.

> **Stacked on #504** (base = `refactor/labels-tags-keyring`). Review/merge #504 first; GitHub will retarget this to `development` once #504 merges. The diff here is PR-3-only.

## Changes
- **`utils/eulerLabelsUtils.ts`** — `isVaultAccessControlled()` (tag check for `'access control'`), reusing the `productHasTag` helper from #504.
- **`composables/useVaultTypeBadges.ts`** — new `'accessControl'` badge in the `VaultTypeBadge`/summary unions; pushed for verified access-controlled vaults.
- **`components/entities/vault/AccessControlBadge.vue`** + **`AccessControlInfoModal.vue`** — new chip + info modal, mirroring `KeyringBadge`/`KeyringInfoModal` (icon `search-user`, tone `accent`).
- **`VaultTypeBadges.vue`** — renders the chip alongside the others.
- **`VaultHooksInfoModal.vue`** — labels the hook target as **"Access control (allowlist)"** and adds a note explaining that gated operations require an on-chain allowlist (access granted by the manager; no self-service verification).
- **tests** — `useVaultTypeBadges` test for the `accessControl` summary badge.

## Deferred (follow-up)
The hook target is `AccessControlEnumerable`, so we can RPC `getRoleMemberCount`/`getRoleMember` per hooked-operation selector (and `WILD_CARD`) to show *which* addresses are whitelisted for each operation. Intentionally left out of this PR to keep it focused on recognition + UI.

## Verification
- `npm run typecheck` passes.
- `vitest run useVaultTypeBadges` passes (9 tests, incl. the new access-control case).
- `eslint` clean on all changed files.

PR #504, opened 2026-05-27
refactor: read keyring classification from labels 'tags' field

kasperpawlowski:
## Summary
Follows euler-xyz/euler-labels#623, which replaces the product `keyring` boolean with a freeform `tags` string array. This switches euler-lite to recognize keyring vaults via `tags.includes('keyring')`.

## Changes
- **`entities/euler/labels.ts`** — drop `keyring?: boolean` from `EulerLabelProduct` and `EulerLabelVaultOverride`; add `tags?: string[]` to `EulerLabelProduct`.
- **`utils/eulerLabelsUtils.ts`** — `isVaultKeyring` / `isProductKeyring` now read product-level `tags` instead of delegating to the SDK's `isEulerLabelVaultKeyring` / `isEulerLabelProductKeyring` (keyring-boolean) helpers. Adds a small `productHasTag` helper.

The `isVaultKeyring` / `isProductKeyring` signatures are unchanged, so all consumers (badges, operation guard, hook modal, discovery graph, item rows) are untouched.

## Why a shim, not an SDK change
The labels parsing/detection now lives in the published `@eulerxyz/euler-v2-sdk`. Its product normalization preserves unknown **product-level** fields (so `tags` survive), but its `vaultOverrides` parser allowlists keys and **drops** unknown override fields. Reading product-level tags here unblocks the migration without an SDK release. Current keyring data is product-level only, so behavior is preserved.

## Follow-up
- Update `@eulerxyz/euler-v2-sdk` to parse `tags` natively (incl. override-level tags) and expose tag helpers; then this shim can delegate to the SDK again.

## Deploy ordering
⚠️ Depends on the euler-labels `tags` data (euler-xyz/euler-labels#623) being live in the served labels before this deploys, otherwise keyring vaults temporarily lose their "Private" classification.

## Verification
- `npm run typecheck` passes.
- `vitest run` for `useVaultTypeBadges` / `useBorrowForm` passes (10 tests).

## Summary by CodeRabbit

* **New Features**
  * Introduced freeform tags system for product labeling, replacing legacy boolean-based approach.

* **Refactor**
  * Updated vault and product metadata structure to support enhanced tagging and explorability controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/euler-xyz/euler-lite/pull/504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #506, opened 2026-05-28
refactor: read keyring classification from labels 'tags' field (master)

kasperpawlowski:
## Summary
Production-bound (`master`) variant of #504. Follows euler-xyz/euler-labels#623, which replaces the product `keyring` boolean with a freeform `tags` string array (also valid on vault overrides). Switches euler-lite to recognize keyring vaults via `tags.includes('keyring')`.

This is the master-targeted version: the SDK refactor that gutted `eulerLabelsUtils.ts` is on `development` but hasn't landed on `master` yet, so the parser is still local here. That means **override-level tags work natively** — no SDK shim required, no follow-up.

## Why a separate PR from #504
`utils/eulerLabelsUtils.ts` is ~435 LOC on `master` vs ~170 LOC on `development` (the SDK absorbed most of it on development). The two branches' diffs against the keyring-boolean migration are structurally different and can't share commits.

## Changes
- **`entities/euler/labels.ts`** — drop `keyring?: boolean` from `EulerLabelProduct` and `EulerLabelVaultOverride`; add `tags?: string[]` to both.
- **`utils/eulerLabelsUtils.ts`** — `extractVaultOverrides` now reads `entry.tags` (string-filtered) instead of `entry.keyring`. `isVaultKeyring` / `isProductKeyring` read tags via small `productHasTag` / `vaultOverrideHasTag` helpers. Supports product-level **and** override-level tags.

The `isVaultKeyring` / `isProductKeyring` signatures are unchanged, so all consumers (badges, operation guard, hook modal, discovery graph, item rows) are untouched.

## Deploy ordering
⚠️ Depends on the euler-labels `tags` data (euler-xyz/euler-labels#623) being live in the served labels before this deploys, otherwise keyring vaults temporarily lose their "Private" classification.

## Verification
- `npm run typecheck` passes.
- `vitest run useVaultTypeBadges` passes (8 tests).

## Companion
- #505 (the access-control chip) has a master-targeted variant stacked on this PR — see the next PR in this branch.

### DIFF
--- a/utils/eulerLabelsUtils.ts
+++ b/utils/eulerLabelsUtils.ts
@@ -344,6 +344,11 @@ export const isVaultKeyring = (vaultAddress: string): boolean => {
 export const isProductKeyring = (productKey: string): boolean =>
   productHasTag(products[productKey], 'keyring')
 
+export const isVaultAccessControlled = (vaultAddress: string): boolean => {
+  const product = getProductByVault(vaultAddress)
+  return productHasTag(product, 'access control') || vaultOverrideHasTag(product, vaultAddress, 'access control')
+}
+
 // Widened to Vault | SecuritizeVault — both expose governorAdmin: string and
 // the helper only reads that one field. Callers from the discovery matrix
 // pass either type without casting.
--- a/utils/eulerLabelsUtils.ts
+++ b/utils/eulerLabelsUtils.ts
@@ -105,7 +105,7 @@ export const extractVaultOverrides = (raw: Record<string, unknown>): Record<stri
     if (Array.isArray(entry.restricted)) override.restricted = entry.restricted.filter((v): v is string => typeof v === 'string')
     if (typeof entry.notExplorableLend === 'boolean') override.notExplorableLend = entry.notExplorableLend
     if (typeof entry.notExplorableBorrow === 'boolean') override.notExplorableBorrow = entry.notExplorableBorrow
-    if (typeof entry.keyring === 'boolean') override.keyring = entry.keyring
+    if (Array.isArray(entry.tags)) override.tags = entry.tags.filter((v): v is string => typeof v === 'string')
     if (Object.keys(override).length > 0) {
       overrides[normalizeAddress(key)] = override
     }
@@ -330,16 +330,19 @@ export const isVaultNotExplorableBorrow = (vaultAddress: string): boolean => {
   return override?.notExplorableBorrow === true
 }
 
+const productHasTag = (product: EulerLabelProduct | undefined, tag: string): boolean =>
+  product?.tags?.includes(tag) ?? false
+
+const vaultOverrideHasTag = (product: EulerLabelProduct | undefined, vaultAddress: string, tag: string): boolean =>
+  product?.vaultOverrides?.[normalizeAddress(vaultAddress)]?.tags?.includes(tag) ?? false
+
 export const isVaultKeyring = (vaultAddress: string): boolean => {
   const product = getProductByVault(vaultAddress)
-  if (product.keyring === true) return true
-  const override = product.vaultOverrides?.[normalizeAddress(vaultAddress)]
-  return override?.keyring === true
+  return productHasTag(product, 'keyring') || vaultOverrideHasTag(product, vaultAddress, 'keyring')
 }
 
-export const isProductKeyring = (productKey: string): boolean => {
-  return products[productKey]?.keyring === true
-}
+export const isProductKeyring = (productKey: string): boolean =>
+  productHasTag(products[productKey], 'keyring')
 
 // Widened to Vault | SecuritizeVault — both expose governorAdmin: string and
 // the helper only reads that one field. Callers from the discovery matrix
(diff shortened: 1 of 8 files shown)


utils/eulerLabelsUtils.ts
```diff
@@ -330,16 +330,19 @@ export const isVaultNotExplorableBorrow = (vaultAddress: string): boolean => {
...
+
+const vaultOverrideHasTag = (product: EulerLabelProduct | undefined, vaultAddress: string, tag: string): boolean =>
+  product?.vaultOverrides?.[normalizeAddress(vaultAddress)]?.tags?.includes(tag) ?? false
+
 export const isVaultKeyring = (vaultAddress: string): boolean => {
   const product = getProductByVault(vaultAddress)
-  if (product.keyring === true) return true
-  const override = product.vaultOverrides?.[normalizeAddress(vaultAddress)]
-  return override?.keyring === true
+  return productHasTag(product, 'keyring') || vaultOverrideHasTag(product, vaultAddress, 'keyring')
```
LeonardEulerXYZ, line 341: This now only recognizes `tags.includes('keyring')`. I checked the currently served `euler-labels` master data and it still has legacy `keyring: true` entries for the active Keyring products (`1/products.json` has `keyring`, `43114/products.json` has `keyring-avalanche`), while `euler-xyz/euler-labels#623` is still open/not merged. If this Lite master PR deploys before that data is live, `isVaultKeyring()` will return false for those vaults, which drops the Private badge/hook labeling and, more importantly, the Keyring operation guard. Please either keep a temporary legacy `keyring === true` fallback here or hold this PR until the labels master data has actually migrated.

LeonardEulerXYZ (dismissed):
Reviewed head `8952d8120e7828f67a40ecf559435c7407356185`.

Verdict: changes requested.

The code change itself is small and the local validation is green, but the production deploy invariant is not currently satisfied: `euler-labels` master still serves the legacy `keyring: true` fields, while this PR removes that reader. If merged/deployed now, existing Keyring/private vaults would temporarily lose their Private classification and the Keyring operation guard until labels master is migrated or the app remains backward-compatible during the rollout.

Validation performed:
- Inspected `origin/master...HEAD` diff for `entities/euler/labels.ts` and `utils/eulerLabelsUtils.ts`
- Checked current raw `euler-labels` master data for chains 1/43114 and confirmed legacy `keyring: true` remains present with no `tags: ['keyring']`
- Checked `euler-xyz/euler-labels#623`: still open, not merged
- `npm run typecheck` passed
- `npx vitest run tests/composables/useVaultTypeBadges.test.ts` passed (8 tests)
- `npm run build` passed, with existing/non-blocking build warnings only

Once the data migration is actually live on the labels branch served by production, or this PR keeps a temporary fallback for legacy `keyring`, I would expect this to be straightforward to re-review.

Seranged (approved):

dglowinski (approved):

utils/eulerLabelsUtils.ts
```diff
@@ -330,15 +330,23 @@ export const isVaultNotExplorableBorrow = (vaultAddress: string): boolean => {
...
```
LeonardEulerXYZ, line 341: This now only recognizes `tags.includes('keyring')`. That is fine after the labels branch served by production has migrated, but `/api/labels/*` defaults to `euler-labels` `master`, and current raw labels `master` still contains legacy `keyring: true` entries for the existing Keyring products without `tags: ['keyring']`. If this app deploys first, those vaults lose their Private classification / Keyring guard. Please either keep a temporary `product.keyring === true` / override fallback here, or merge the labels `master` migration before this app deploys.

LeonardEulerXYZ (changes requested):
Reviewed head `4c07b655c189a647a6a5d6837cb1d576996b991e`.

Verdict: changes requested — the code path is clean, but the production rollout invariant is still not satisfied for `master`.

The app defaults `/api/labels/*` to `euler-xyz/euler-labels` branch `master` (`server/api/labels/[file].get.ts`), and current raw labels `master` still has legacy `keyring: true` product fields on chains 1 and 43114 with no `tags: ['keyring']` replacement there. Since this PR removes the legacy `keyring` reader entirely, deploying it before labels `master` is migrated would make existing Keyring/private vaults lose their Private classification and operation guard. `euler-labels#623` is merged, but into `development`, not the labels branch this production app serves by default.

Validation performed:
- Inspected `origin/master...HEAD` diff at current PR head.
- Checked Lite labels proxy default branch (`NUXT_PUBLIC_CONFIG_LABELS_REPO_BRANCH || 'master'`).
- Checked current raw `euler-labels` master for chains 1 and 43114: legacy keyring entries remain, tag-based keyring entries absent.
- Checked `euler-xyz/euler-labels#623`: merged to `development`.
- `npm run typecheck` passed.
- `npm run test:run -- useVaultTypeBadges eulerLabelsUtils` passed (1 file, 9 tests).
- `npm run test:run` passed (75 files, 845 tests).
- `npm run build` passed with existing/non-blocking warnings.

Once labels `master` is migrated before this deploy, or the app keeps a temporary fallback for legacy `keyring`, this should be straightforward.

PR #508, opened 2026-05-28
Sdk fixes

dglowinski:
## Summary
  - Bring the SDK-backed branch back into parity with `development` for discovery cards, vault detail pages,
  position pages, modal previews, and share-link UX.
  - Fix post-transaction freshness so portfolio and wallet balances invalidate through the SDK query cache
  and refresh against the same subgraph proxy used by account reads.
  - Restore SDK migration regressions around borrow collateral visibility, CoW multiply quotes in spy mode,
  selectable collateral display, friendly simulation errors, reward snapshots, intrinsic APY, and vault
  warnings/badges.

  ## Changes
  - Bumps `@eulerxyz/euler-v2-sdk` to `0.2.10-beta`.
  - Replaces the partial `Featured` vault concept with the existing `Recently added` behavior across lend,
  borrow, and earn.
  - Sorts borrow collateral visibility by highest oracle value and adds regression coverage.
  - Routes post-tx subgraph catch-up through `/api/proxy/subgraph/{chainId}`, adds a delayed refresh retry,
  and marks balance queries for post-tx invalidation.
  - Restores vault overview UI for IRM, cyclical IRM, oracle routes, risk parameters, preview modals, exact
  swap displays, stale-quote warnings, and warning badges.
  - Gates Keyring/Cyclical/Securitize badges on governor verification, quiets browser debug logs unless opted
  in, and adds env control for rate limiting.
  - Switches yzPrime intrinsic APY to the DefiLlama pool source and rehydrates snapshot rewards into
  `VaultRewardInfo`.

  ## Test plan
  - [ ] `npm run test:run -- tests/utils/portfolio-visibility.test.ts tests/utils/tx-errors.test.ts tests/
  utils/yuzu-intrinsic-apy.test.ts`
  - [ ] `npm run typecheck`
  - [ ] Manually verify lend/borrow/earn discovery cards show consistent `Recently added` badges and
  ordering.
  - [ ] Manually verify post-borrow/post-deposit portfolio and repay wallet balances refresh without waiting
  for the 60s cache window.
  - [ ] Manually verify vault detail pages show the restored preview/share-link UX, oracle-route rows, IRM/
  risk cards, and warning badges.

## Summary by CodeRabbit

* **New Features**
  * Recently added vault badging system to highlight new vaults across markets
  * Debug opt-in flags for performance profiling and verbose logging

* **Improvements**
  * Enhanced token address display with truncation and tooltips for better readability
  * Refactored modal interactions for smoother preview and opening workflows
  * Improved collateral selection UI with better labeling and layout
  * Better error messaging for transaction simulation failures
  * Optimized balance refresh after withdrawals and transactions

* **Bug Fixes**
  * Quote timestamp tracking for swap-based operations
  * Governance verification badge visibility rules


server/utils/rate-limit.ts
```diff
@@ -80,6 +80,9 @@ export function createRateLimiter(config: RateLimiterConfig) {
      * In production, throws 403 if the request did not arrive via Cloudflare.
      */
     consume(event: H3Event, cost = 1): void {
+      // Escape hatch for local tooling (e.g. parity capture) that hammers the
+      // app from a single IP. Never set in deployed environments.
+      if (process.env.DISABLE_RATE_LIMIT === 'true') return
```
LeonardEulerXYZ, line 85: Blocking: this escape hatch runs before the production `CF-Connecting-IP` fail-closed check, so `DISABLE_RATE_LIMIT=true` in a deployed env would disable both per-IP limiting and the Cloudflare-origin guard. Since this endpoint protects public proxies, please make the bypass fail-closed in `prd` (or refuse startup when both `DOPPLER_ENVIRONMENT=prd` and `DISABLE_RATE_LIMIT=true` are set). A single env typo should not remove the invariant.

server/api/proxy/intrinsic-apy-overrides.get.ts
```diff
@@ -5,13 +5,12 @@ import { createRateLimiter } from '~/server/utils/rate-limit'
...
 import {
   extractHyperbeatWeightedApr,
-  extractLatestYuzuApy,
   extractValantisApy,
   type IntrinsicApyOverrideRow,
 } from '~/utils/yuzu-intrinsic-apy'
 
 const ALLOWED_METHODS = new Set(['GET', 'HEAD'])
-const PENDLE_STALE_MS = 7 * 24 * 60 * 60 * 1000
+const PENDLE_STALE_MS = 2 * 60 * 60 * 1000
```
LeonardEulerXYZ, line 13: Blocking/data correctness: this reduces Pendle freshness from a broad window to 2h, but the live Pendle market endpoint currently returns `timestamp: 2026-03-18T23:00:00.000Z` with a non-zero `impliedApy` (~0.1095). With this threshold `pendle-khype` is always treated stale and dropped by `row(..., stale ? 0 : ...)`. Please confirm Pendle's timestamp semantics or restore/use a freshness signal that does not silently remove this APY override in production.

utils/logger.ts
```diff
@@ -158,7 +159,18 @@ const emitNode = (level: Level, fields: Fields, msg: string | undefined): void =
...
+// `logger.warn` — are dropped to keep the console quiet. Opt back in to the full
+// stream with `?verbose` / `localStorage.euler_verbose=1`. The Node path is
+// unaffected: server logs keep every level for BetterStack queries.
+const LEVEL_PRIORITY: Record<Level, number> = {
+  trace: 10, debug: 20, info: 30, warn: 40, error: 50, fatal: 60,
+}
+const BROWSER_MIN_LEVEL = verboseLogsOptIn ? LEVEL_PRIORITY.trace : LEVEL_PRIORITY.error
+
 const emitBrowser = (level: Level, fields: Fields, msg: string | undefined): void => {
+  if (LEVEL_PRIORITY[level] < BROWSER_MIN_LEVEL) return
```
LeonardEulerXYZ, line 173: Blocking validation: this behavior change is intentional, but the existing browser logger test still expects `logger.warn(...)` to call `console.warn`. `npm run test:run` now fails in `tests/utils/logger.test.ts:176` because no call is emitted. Please update the test to cover the new default suppression and the `?verbose`/`localStorage` opt-in path.

composables/useSwapQuotesParallel.ts
```diff
@@ -245,23 +245,21 @@ export const useSwapQuotesParallel = (options: SwapQuotesParallelOptions) => {
...
   ): Promise<SwapQuoteCard | null> => {
-    const amountUsdPromise = getAmountUsd(quote).catch(() => undefined)
-
     // CoW intents settle off-chain — gas cost is genuinely zero from the
     // user's perspective. Mark the card so the route selector renders
-    // "Gasless" and the ranking treats it as a real zero rather than a
-    // missing estimate.
+    // "Gasless". Do not block route visibility on auxiliary USD pricing.
     if (isCowProviderOrQuote(provider, quote)) {
       return {
```
LeonardEulerXYZ, line 252: Non-blocking UX concern: the CoW/gasless early return skips `amountUsd`, while quote sorting ranks every card with a non-null USD score ahead of cards without one. That can push a gasless CoW route below worse priced routes simply because the others priced successfully. Consider still fetching `amountUsd` opportunistically for CoW and only treating USD pricing as non-blocking for visibility.

LeonardEulerXYZ (changes requested):
Reviewed head 83fcda813eb187d10d689b87c8027aeb33c8d664.

Verdict: Request changes.

Validation performed:
- `npm run test:run` — fails: `tests/utils/logger.test.ts` browser warn expectation no longer receives a console call.
- `npm test -- --run tests/utils/tx-errors.test.ts tests/utils/portfolio-visibility.test.ts tests/utils/yuzu-intrinsic-apy.test.ts tests/utils/logger.test.ts` — same logger failure; other targeted tests passed.
- `npm run build` — passed, with existing chunk/Sentry-token warnings.
- Diff/security sweep over changed files and package bump.

Main blockers are the failing test suite and two production/data-safety concerns called out inline. No browser smoke run; this pass was code review + local validation for a broad SDK-fixes PR.

server/utils/rate-limit.ts
LeonardEulerXYZ, line 85: This escape hatch now returns before the production `CF-Connecting-IP` fail-closed check below. If `DISABLE_RATE_LIMIT=true` is ever present in `DOPPLER_ENVIRONMENT=prd`, every route using this limiter loses both throttling and the origin-bypass protection. Please either run the CF guard before this return, or make the escape hatch explicitly non-prod (for example, only when `DISABLE_RATE_LIMIT === 'true' && DOPPLER_ENVIRONMENT !== 'prd'`).

composables/useEulerTx.ts
```diff
@@ -1014,14 +1031,9 @@ export const useEulerTx = () => {
...
     triggerPortfolioRefresh()
-    if (lastReceipt && SUBGRAPH_URL) {
-      void waitForSubgraphBlock(SUBGRAPH_URL, lastReceipt.blockNumber)
-        .then((caughtUp) => {
-          if (caughtUp) {
-            void invalidateSdkQueries([...INVALIDATE_AFTER_TX])
-            triggerPortfolioRefresh()
-          }
-        })
+    const cid = chainId.value
```
LeonardEulerXYZ, line 1034: Non-blocking, but this reads the reactive chain at finalization time rather than the chain captured for the transaction execution. If the user switches network/view while the wallet is confirming, post-tx catch-up can poll `/api/proxy/subgraph/{newChain}` for a receipt block from the old chain. Consider passing the execution `cid` into `finalizeExecution` so the subgraph wait matches the chain used by `executeTransactionPlan`.

utils/tx-errors.ts
```diff
@@ -175,6 +175,21 @@ const formatDecodedChain = (chain: readonly DecodedSmartContractError[]): string
   return inner === outer ? inner : `${inner}  (wrapped by ${outer})`
 }
 
+// If any error in the decoded chain maps to a known user-facing message,
+// prefer that friendly copy (e.g. "Account liquidity too low for this
+// action.") over the raw signature dump the SDK gives us. This keeps the
+// SDK simulation path in parity with the thrown-error path (getTxErrorMessage),
+// which already resolves names through ERROR_MESSAGE_MAP.
+const friendlyMessageFromChain = (
+  chain: readonly DecodedSmartContractError[],
```
LeonardEulerXYZ, line 184: This walks the SDK decoded chain outer-first. The formatter above documents that the SDK returns outer wrapper first and inner cause last, so if both wrapper and inner error names are mapped, this will choose the less-specific outer friendly message. If that is intentional, a regression test would be useful; otherwise iterate from the end to preserve the “most specific cause” behavior.

LeonardEulerXYZ (changes requested):
## Code Review Summary

**Verdict: Changes requested**

I found one production hardening issue that I would fix before merge. The PR otherwise looks directionally sound: targeted tests, full Vitest suite, typecheck, and production build passed locally; a local built-server API smoke also passed for the core app/API endpoints I sampled.

### Blocking
- `server/utils/rate-limit.ts` now lets `DISABLE_RATE_LIMIT=true` return before the production `CF-Connecting-IP` fail-closed guard. That escape hatch is useful for local tooling, but in `DOPPLER_ENVIRONMENT=prd` it would disable both throttling and the origin-bypass protection for every route using this limiter.

### Non-blocking notes
- Post-tx subgraph catch-up currently reads the reactive chain at finalization time rather than the chain captured for execution; worth tightening so network/view switches during wallet confirmation cannot poll the wrong subgraph.
- The friendly simulation-error mapping now prefers the first mapped decoded error in an outer-first chain; if both wrapper and inner cause are mapped, this may surface the less-specific message. A small regression test would remove the ambiguity.
- The yzPrime DefiLlama migration works in local API smoke, but the test coverage mostly exercises helper parsers rather than the chain-143 route/source behavior.

### Validation
- Worktree: `/home/leonard/euler-lite-worktrees/pr-508`
- Reviewed head: `02ee48265c5a3862f770a1b4ecb2038d86972515`
- `npm run test:run -- tests/utils/portfolio-visibility.test.ts tests/utils/tx-errors.test.ts tests/utils/yuzu-intrinsic-apy.test.ts` — passed, 20 tests
- `npm test -- --silent` — passed, 73 files / 813 tests
- `npm run build` — passed
- Local built-server API smoke on `127.0.0.1:3378`: `/`, `/api/euler-chains`, `/api/token-list?chainId=1`, `/api/public/metadata?chainId=1`, `/api/proxy/intrinsic-apy-overrides?chainId=143` all returned 200. The chain-143 intrinsic APY endpoint returned the yzPrime DefiLlama row.

Browser smoke: incomplete — local Playwright browser binaries were missing in this worktree environment; I fell back to built-server HTTP/API smoke rather than claim route-level visual coverage.

A small guard before the escape hatch should make the main issue vanish. Elegant fix, small diff — my favorite species.

Seranged (approved):

PR #574, opened 2026-06-15
feat: collect additional ToU for RLP redemption vaults

kasperpawlowski:
## Summary

Adds support for collecting an **additional** Terms of Use signature on top of the global ToS, gated per-vault via a new `"RLP redemption ToU"` label tag. Acceptance is recorded on-chain by reusing the existing `termsOfUseSigner` contract with a different message hash, so it lands atomically in the same EVC batch as the user's operation — no separate signing UX, no gas-only tx.

## How it works

- **Tag detection** — new helper `isVaultRlpRedemptionTou()` checks the `"RLP redemption ToU"` tag at product, vault-override, and earn-entry layers in `utils/eulerLabelsUtils.ts`.
- **Config** — new `NUXT_PUBLIC_CONFIG_RLP_TOU_MD_URL` env var. When set, the markdown is fetched directly from the URL (no `/api/...` proxy), keccak256'd, and the resulting hash is checked against `lastTermsOfUseSignatureTimestamp` on-chain. Intentionally **not** added to `.env.example` since it is tied to a one-off process.
- **Per-vault guard** — `useRlpRedemptionTouGuard(addresses)` mirrors `useTosGuard` but is inert unless an in-scope vault carries the tag. Wired into `useOperationGuard` alongside the other guards.
- **SDK plugin** — `createLiteRlpTouPlugin()` is a sibling of `createLiteTosPlugin()`, registered immediately after it. Each plugin owns its own signature map, so global ToS and RLP ToU never clobber each other. The existing `utils/sdk-tos.ts` is unchanged.
- **UI** — new `AcknowledgeRlpRedemptionTermsModal` and a `showRlpTouFlow` branch in `VaultFormSubmit`. Precedence: Keyring → global ToS → **RLP ToU** → unverified vault.

When the redemption process is finished, the feature can be removed by deleting the four new files (`useRlpTouData.ts`, `sdk-rlp-tou.ts`, `useRlpRedemptionTouGuard.ts`, `AcknowledgeRlpRedemptionTermsModal.vue`) and reverting the six small wire-up edits. The global ToS path is unaffected.

## Test plan

- [ ] With `NUXT_PUBLIC_CONFIG_RLP_TOU_MD_URL` set to a CORS-permissive markdown URL and a local labels override adding `"RLP redemption ToU"` to a vault's `tags`, visit `/lend/<tagged-vault>` and confirm the submit button is replaced by "Accept RLP Redemption Terms" once the global ToS is accepted.
- [ ] Trigger an operation and decode the EVC batch in `OperationReviewModal` — expect two `signTermsOfUse` items prepended (global + RLP), both targeting `termsOfUseSigner`.
- [ ] After the tx confirms, refresh the page and verify `checkHasSigned()` resolves `true` for both hashes (no further prompts).
- [ ] Visit an untagged vault — confirm the RLP guard is inert (no fetch, no blocker, no modal) and global ToS behavior is unchanged.
- [ ] Switch wallets mid-session — the RLP prompt should reappear for the new account if it has not signed the RLP hash.
- [ ] Leave `NUXT_PUBLIC_CONFIG_RLP_TOU_MD_URL` empty — confirm the feature is fully disabled (no fetch, no blocker, no UI) even on tagged vaults.

## Release Notes

* **New Features**
  * Introduced RLP redemption Terms-of-Use acceptance requirement. Users performing operations on RLP-eligible vaults must review and accept redemption terms via a new modal. The modal displays the terms and provides a link to the complete terms document.


composables/guards/useRlpRedemptionTouGuard.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    try {
+      const client = rpcClient.value!
+      const lastSignTimestamp = await client.readContract({
+        address: tosSignerAddress.value,
+        abi: tosSignerReadAbi,
+        functionName: 'lastTermsOfUseSignatureTimestamp',
+        args: [address.value as Address, data.tosMessageHash],
+      })
+      hasSigned.value = (lastSignTimestamp as bigint) > 0
```
LeonardEulerXYZ, line 80: This commits an async result without proving it still belongs to the current wallet/chain/tagged-vault context. A stale `checkHasSigned()` from the previous account can resolve after the address/chain watchers reset state and set `hasSigned = true` for the new account. In the RLP ToU path that is a fail-open: the operation blocker is cleared, but `updateSdkSignature()` will not store an RLP signature for the new account, so the subsequent plan can proceed without prepending the additional `signTermsOfUse` item.

Please add a generation/snapshot guard around the awaited ToU fetch and `readContract()` result before mutating `tosData`, `tosLoadFailed`, or `hasSigned`.

LeonardEulerXYZ (changes requested):
## Leonard review — PR #574

Verdict: Request changes

I reviewed head `61a189362923644dc77c3dced81acf93beb36f68` for the new RLP redemption Terms-of-Use gate.

Scope checked:
- Diff and surrounding code for the new RLP ToU guard, modal, SDK plugin, runtime config, label tag helper, and `VaultFormSubmit` wiring.
- Existing ToS/keyring/operation-guard patterns for comparison.
- Prior PR artifacts: no prior Leonard comments/reviews were present; CodeRabbit had two inline findings, and I verified the async-race one as a real blocker.
- Supply-chain/malicious-code sweep over the PR diff: only the intended ToU markdown `fetch()` surface appeared; no dependency, workflow, install-script, credential, or obfuscated-code changes.

Validation:
- `npm run test:run -- tests/utils/euler-labels-utils.test.ts tests/composables/useEulerSdk.test.ts tests/utils/transactionPlanApprovals.test.ts` — passed, 20 tests.
- `npm run build` — passed; only existing-style Rollup/Vite chunk/dynamic-import warnings observed.
- Railway preview route smoke, headed Chromium under Xvfb:
  - desktop: `/lend/0xD8b27CF359b7D15710a5BE299AF6e7Bf904984C2?network=1` returned 200 and rendered the WETH vault form/content.
  - mobile: same route at iPhone viewport returned 200 and rendered the supply form plus global ToS CTA.
  - Caveat: the preview has the RLP ToU URL unset, so this was route/mobile regression smoke, not an exercised RLP acceptance flow. No wallet/signing smoke was run.

Blocking finding:
- The new RLP guard can apply stale async signature-check results after a wallet or chain switch. For this feature, that can fail open: an old account’s positive `lastTermsOfUseSignatureTimestamp` read may set `hasSigned = true` for the new account, unregistering the RLP blocker while no RLP `signTermsOfUse` item is stored for that new account. The user could then continue without the additional RLP ToU being recorded. See inline comment.

Non-blocking but worth tightening:
- The signed RLP message includes only a short prefix of the markdown content hash. For an acceptance artifact meant to bind a user to a specific legal text, the full content hash is the cleaner invariant. CodeRabbit already left an inline comment on this; I agree with the direction.

Smoke coverage: browser visual route smoke + mobile smoke only; no wallet/signing smoke; RLP modal/acceptance flow not exercised because the deployed preview did not enable the RLP ToU URL.

Screenshots: captured for local inspection but not posted, because they only show ordinary WETH vault route rendering while the gated RLP flow is disabled on the preview. Posting them would add noise rather than useful review evidence.

Comment lifecycle: no prior Leonard comments/reviews/inline comments found, so no cleanup was needed.

PR #577, opened 2026-06-17
fix: guard vault loads against chain switches

Seranged:
## Summary
- Prevent the shared vault-loading pipeline from mixing vault addresses from one chain with a newer active chain id after a network switch.
- Covers the app-wide EVault, Earn, Securitize, escrow, unresolved-collateral, and background refresh paths that feed Earn, Lend, Borrow, Explore, Portfolio, and detail pages.
- Fixes invalid V3 detail requests such as Base Earn vaults being requested under chain 1; Earn was the observed symptom, not the only protected loader.

## Changes
- Carry the captured load chain id through EVault, Earn, Securitize, escrow, and unresolved-collateral loaders.
- Abort stale load generations when the active chain no longer matches the chain that started the load.
- Capture the refresh chain id in `refreshVaults()` so background refreshes cannot reuse registry addresses against a new active chain.
- Add regression coverage for an Earn vault load invalidated by a chain switch before detail fetches run.

## Test plan
- [x] npm run test:run -- tests/composables/useVaults.test.ts
- [x] npm run typecheck

### DIFF
--- a/composables/useVaults.ts
+++ b/composables/useVaults.ts
@@ -72,6 +72,7 @@ const loadGeneration = ref(0)
 
 interface UpdateEVaultsOptions {
   verifiedAddresses?: ReadonlySet<string>
+  chainId?: number
 }
 
 const getSdkVaults = async () => {
@@ -169,9 +170,20 @@ const resetVaultsState = () => {
   resetVaultCategoryCache()
 }
 
+const isCurrentVaultLoad = (generation: number, targetChainId: number): boolean => {
+  const { chainId } = useEulerAddresses()
+  return loadGeneration.value === generation && chainId.value === targetChainId
+}
+
+const resolveTargetChainId = (targetChainId?: number): number => {
+  const { chainId } = useEulerAddresses()
+  return targetChainId ?? chainId.value
+}
+
 const updateEVaults = async (vaultAddresses: string[], generation?: number, silent = false, options: UpdateEVaultsOptions = {}) => {
   const { setMany: registrySetMany, get: registryGet, isKnownEscrowAddress } = useVaultRegistry()
   const gen = generation ?? loadGeneration.value
+  const targetChainId = resolveTargetChainId(options.chainId)
 
   try {
     if (!silent) {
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed head `5ea5addada44360b2bc183673e5fe97d6ef2cc96`.

Verdict: no blocking findings from this pass.

What changed:
- `useVaults` now captures the intended chain id at the start of vault-load/refresh flows and re-checks both generation and chain before registering EVault, Earn, Securitize, escrow, and unresolved-collateral results.
- The load pipeline passes that captured chain through the nested loaders, so late results from an invalidated chain switch should not write stale vaults into the active registry.
- A regression test covers an invalidated load where the chain changes before Earn vault fetching.

Risk areas checked:
- Stale async writes after chain switch: the meaningful registry-write paths added chain-aware guards before fetches and before `setMany` writes.
- Nested collateral resolution: the captured `startChainId` is carried into the unresolved-collateral and escrow fetch path, avoiding a silent fall-back to the current chain midway through a stale load.
- Public-facing surface: no new secrets, external endpoints, eval/dynamic execution, or privileged client behavior introduced by the diff.

Validation performed:
- `npm run test:run -- tests/composables/useVaults.test.ts` — passed, 8 tests.
- `npm run build` — passed, with existing bundle/chunk warnings only.
- Browser smoke on a local PR build using headed Chromium under Xvfb: desktop 1440x1000 and mobile 390x844 navigated `/borrow?network=1` then `/borrow?network=8453`; final route stayed on network 8453, Base borrow rows rendered, `/api/vaults?chainId=8453` and same-chain RPC calls returned 200, and no page errors occurred. Observed Merkl allowlist 404 console noise; not PR-specific to this diff.

Smoke coverage: browser route/API smoke + mobile smoke. Wallet/signing smoke not run; this PR changes public vault-load invalidation, not a transaction/signing path.

Screenshots: not posted — this is a state/race guard change rather than a visual UI change, and the smoke evidence was request/route/body assertions rather than screenshot-diff evidence.

dglowinski (approved):

PR #564, opened 2026-06-09
Add transaction batching workflow

dglowinski:
## Summary
- add batch transaction state, drawer, review modal, labels, and batch page
- wire borrow, lend, withdraw, repay, swap, and multiply flows into the batch queue
- add Docker/Railway support for previewing Lite against an unreleased SDK branch and document local SDK development

## Test plan
- npm run typecheck
- commit hooks: eslint --fix via lint-staged

## Summary by CodeRabbit

* **New Features**
  * App-wide transaction batching: "Add to batch" across flows, laptop batch drawer, dedicated Batch page, per-entry review, and full Batch Review modal with approvals, execute, copy-calldata, and Tenderly simulate.
  * Simulation overlays: simulated wallet/position deltas, revert/warning alerts, and visual highlights for modified/removed items.

* **Settings**
  * New "Enable advanced mode" toggle to access batch features.

* **Documentation**
  * Docker build can opt into a preview SDK branch; added local SDK development instructions.

* **Style**
  * Added an intermediate error color shade for UI consistency.

PR #591, opened 2026-06-19
Codex/fix batch collateral swap

dglowinski:


dglowinski:
@coderabbit review

LeonardEulerXYZ (approved):
Reviewed as the stacked child delta against `origin/euler-lite-batch` / parent PR #564, not against `development`.

I focused on the changed batch/swap surfaces:

- `BatchReviewModal.vue`
- `AssetInput.vue`
- `useTxBatch.ts`
- `batchHealthSummary.ts`
- the related composable/utility tests

No blocking findings from the child delta.

Validation run:

- Focused tests: `npm run test:run -- tests/composables/useTxBatch.test.ts tests/composables/useTxBatchModifiedKeys.test.ts tests/composables/useSwapPageLogic.test.ts tests/utils/batchHealthSummary.test.ts` — 32/32 passing
- Typecheck: `npm run typecheck` — passing
- Lint on touched files command — exit 0; only existing warning-class `no-explicit-any` noise in unrelated files
- Production build: `npm run build` — passing
- PR deploy smoke: `https://dev-build-euler-lite-pr-591.up.railway.app` `/` and `/batch` both returned 200, title `Euler Finance`, with no captured console/page errors

Head reviewed: `682c4903e90f2db69dca1c75bbf40d621e9678c4`.

Clean work. The useful invariant here is that the batch review now scopes health/wallet/position changes to the actually touched entries rather than making the user solve a full portfolio diff in their head — a modest mercy for humans, who remain sadly non-deterministic machines.

composables/useTxBatch.ts
```diff
@@ -619,35 +744,67 @@ const hydrateBorrowLiquidity = (
...
+      totalCollateralValue.oracleMid = (totalCollateralValue.oracleMid ?? 0n) + (value.oracleMid ?? 0n)
     }
-  }
-
-  for (const collateralAddress of collateralAddresses) {
-    const valueUsd = getPositionUsdValue(positionsByVault.get(collateralAddress))
     if (valueUsd === undefined) {
       totalCollateralValueUsd = undefined
-      break
+      continue
```
LeonardEulerXYZ, line 801: This makes the aggregate USD value order-dependent when only some collateral values are known. If an earlier collateral has no `valueUsd`, `totalCollateralValueUsd` is set to `undefined`, but a later known collateral then turns it back into a numeric partial total via `(totalCollateralValueUsd ?? 0) + valueUsd`. The review modal can then present an incomplete collateral total as if it were complete. Please preserve the invariant that the aggregate is unknown if any included component is unknown, e.g. track a `hasMissingCollateralUsd` flag and set the final aggregate to `undefined`, and add a focused mixed-known/unknown collateral test.

composables/useTxBatch.ts
```diff
@@ -722,7 +879,7 @@ export const buildModifiedPositionKeySets = (
   for (const [addr, sa] of Object.entries(current.subAccounts)) {
     if (!sa) continue
     const subAccount = subAccountMapKey(addr)
-    const baseSa = base.subAccounts[subAccount]
+    const baseSa = getAccountSubAccount(base, subAccount)
```
LeonardEulerXYZ, line 882: This touched-position derivation still only marks shares/debt deltas below, but this PR now uses those keys to decide which borrow health rows to show. Pure collateral/controller enablement changes can change health without changing a position's shares or borrowed amount, so the health summary can omit exactly the row the user needs to review before executing the batch. Please include enablement-state changes in the touched-key derivation, or derive health-summary touched positions from the batch plan's affected subaccounts/controllers, and cover pure enable/disable collateral/controller cases in tests.

LeonardEulerXYZ (changes requested):
Reviewed the current head `682c4903e90f2db69dca1c75bbf40d621e9678c4` as the stacked child delta against `origin/euler-lite-batch` / parent PR #564, not against `development`.

Verdict: requesting changes.

I found two correctness gaps in the changed batch health/liquidity projection. They are both in the “review what the batch will do before signing” path, so I would fix them before merging: the UI can otherwise show a partially known collateral USD total as if it were complete, and it can omit health changes that are caused only by collateral/controller enablement state.

Validation performed:

- Focused tests: `npm run test:run -- tests/composables/useTxBatch.test.ts tests/composables/useTxBatchModifiedKeys.test.ts tests/utils/batchHealthSummary.test.ts tests/composables/useSwapPageLogic.test.ts` — 32/32 passing
- Typecheck: `npm run typecheck` — passing
- Targeted ESLint on touched source/test files — passing
- Production build: `npm run build` — passing
- GitHub checks observed: CodeRabbit pass; Railway dev-build pass
- Light deployed route smoke against parent PR #564 and this PR #591 on `/position/1/collateral/swap?network=1&spy=0x68e7E72938db36a5CBbCa7b52c71DBBaaDfB8264`, desktop and mobile widths — route returned 200 with real app content and no captured console/page errors. Coverage was limited: the chosen spy route did not expose a populated swap form, so this is route/mobile smoke, not wallet/signing or interaction coverage.

- The new helper extraction into `utils/batchHealthSummary.ts` is the right direction versus burying this in one modal.
- The missing cases are reusable invariant gaps, not local copy issues: changed-position key derivation should represent all state that can alter health, and the partial-USD aggregation rule should be tested directly. Please add focused tests for mixed known/unknown collateral USD values and pure enable/disable collateral/controller changes.

No screenshots were posted: the smoke route only reached the route shell / onboarding copy and did not produce a meaningful changed-surface visual artifact.

Comment lifecycle: no existing Leonard inline comments were present to clean up; an earlier Leonard approval exists on this same head and cannot be edited as a formal review, so this review supersedes it with the blocking findings below.

PR #594, opened 2026-06-19
Fix batch repay simulated balances

dglowinski:
## Summary
- Fix repay-with-collateral Max so target-debt quotes are requested even when the same displayed amount was previously emitted.
- Preserve owner-account simulated changes from full repay flows so moved collateral keeps simulated-state highlighting.
- Include wallet tokens that first appear in later simulation layers so downstream supply forms see withdrawn collateral as simulated wallet balance.

## Changes
- Force AssetInput Max to emit input because Max is an explicit action and parent state may have reset between clicks.
- Add affected sub-account metadata for full repay batch entries and use it when filtering simulated modified keys.
- Build wallet balance overlays from the union of all simulated wallet layers, not only the first layer.
- Add focused regression coverage for target-debt quote requests, affected owner keys, and later-layer wallet token overlays.

## Test plan
- [x] npm run test:run -- tests/composables/useTxBatch.test.ts tests/composables/useTxBatchModifiedKeys.test.ts tests/composables/useRepaySwapCore.test.ts tests/composables/useSavingsRepay.test.ts tests/composables/useWalletSwapRepay.test.ts
- [x] lint-staged eslint --fix via commit hook
- [ ] Manually verify repay collateral Max quote loading, full-repay moved-collateral highlighting, and subsequent supply wallet balance in the browser

## Release Updates

* **Bug Fixes**
  * The "Max" action in asset input now consistently applies the maximum available amount, regardless of previous transaction state.

* **Improvements**
  * Enhanced multi-layer transaction handling for improved accuracy in wallet balance calculations.
  * Improved tracking of affected sub-accounts across batch operations to ensure more reliable transaction execution.

* **Tests**
  * Expanded test coverage for complex repay scenarios and wallet balance layer computations.

### DIFF
--- a/components/entities/asset/AssetInput.vue
+++ b/components/entities/asset/AssetInput.vue
@@ -37,8 +37,8 @@ const isFocused = ref(false)
 const emitInputTimeout = ref<ReturnType<typeof setTimeout> | null>(null)
 const lastEmittedInputValue = ref(model.value)
 
-const emitInputIfChanged = () => {
-  if (lastEmittedInputValue.value === model.value) return
+const emitInputIfChanged = (force = false) => {
+  if (!force && lastEmittedInputValue.value === model.value) return
   lastEmittedInputValue.value = model.value
   emits('input')
 }
@@ -53,12 +53,12 @@ const emitInputDebounced = () => {
   }, 250)
 }
 
-const emitInputNow = () => {
+const emitInputNow = (force = false) => {
   if (emitInputTimeout.value) {
     clearTimeout(emitInputTimeout.value)
     emitInputTimeout.value = null
   }
-  emitInputIfChanged()
+  emitInputIfChanged(force)
 }
 
 const matchesSelectedSubAccount = (a?: string, b?: string) => {
@@ -136,7 +136,9 @@ const setMax = () => {
     return
   }
   model.value = trimTrailingZeros(formatUnits(props.balance ?? 0n, Number(props.asset.decimals)))
-  emitInputNow()
+  // Max is an explicit command. The parent may have reset this model since
+  // the last emitted input, so do not suppress the handler by cached value.
+  emitInputNow(true)
   if (inputEl.value) {
     inputEl.value.value = model.value || ''
   }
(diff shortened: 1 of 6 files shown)


Seranged (approved):

LeonardEulerXYZ (commented):
Reviewed current head be0b6f8618db489dbfbe315a960b9f3324b7f201.

Verdict: COMMENT — no blocking issues found in the diff.

What changed / risk area:
- `AssetInput` now force-emits the default Max action even when the model value matches the last emitted input, which addresses the repeated-max / parent-reset interaction path.
- Batch modified-position scoping now accepts explicitly affected sub-accounts, so full repay cleanup that moves collateral to the owner account does not get filtered out as an unrelated owner-key change.
- Batch wallet balance stitching now collects wallet tokens across all simulated layers, not just layer 0, so later-layer wallet movements are represented in simulated balances.
- The repay page marks full-repay cleanup targets, including owner account and savings source account where relevant.

Validation run:
- Focused tests passed: `npm run test:run -- tests/composables/useTxBatch.test.ts tests/composables/useTxBatchModifiedKeys.test.ts tests/composables/useRepaySwapCore.test.ts tests/composables/useSavingsRepay.test.ts tests/composables/useWalletSwapRepay.test.ts` — 5 files, 28 tests.
- Production build passed: `npm run build`.
- Typecheck was attempted but is not a clean gate here: `npm run typecheck` fails with a Vue Router Volar package export error and broad existing strictness errors outside this PR’s changed surface.
- Preview route smoke, desktop and mobile, loaded the repay page at `/position/1/repay?network=1&spy=0x000001ac4e512d670c34feDf6c71cE2F49fb160a`; the route rendered the USDC/EUL repay form rather than `Position not found`.

- The batch changes extract reusable helpers (`buildScopedEntrySubAccounts`, `buildWalletBalanceLayers`) and add focused tests around both. This is the right shape for a shared batch invariant rather than a one-off repay-only patch.
- The new `getAffectedSubAccounts` helper is local to the repay page. That looks acceptable for this PR because only repay cleanup currently needs to declare extra touched accounts, but if borrow/withdraw/multiply start needing similar cross-account cleanup metadata, this should move into a shared batch-entry helper to avoid sibling drift.
- I did not find sibling repay tabs left with the old owner-key filtering bug: wallet direct, wallet swap, collateral, and savings repay all pass `affectedSubAccounts` for full repay.

Smoke coverage: route/browser smoke on the deployed PR preview, desktop + mobile viewport. No wallet signing coverage. No screenshots posted: this is state-wiring behavior with no meaningful visual delta to preserve, and the text/route evidence was sufficient for the reviewed surface.

PR #608, opened 2026-06-23
Lite 145 min borrow liquidity filter

kanvgupta:
## Summary by CodeRabbit

* **New Features**
  * Added a default minimum borrow liquidity custom filter ($1,000) on the borrow page (including behavior for unavailable metric values).

* **Bug Fixes**
  * Improved borrow page readiness to require cached per-pair USD liquidity and total-borrowed data (not a single aggregate flag).
  * Prevented stale debounced price fetch results from overwriting newer cache updates.
  * Enhanced custom filter evaluation to correctly handle empty filter sets and treat unavailable/non-finite metric values based on filter configuration.


LeonardEulerXYZ (commented):
Reviewed head `2352496069ccedab7496199d04006045a9c4c812`.

Verdict: **COMMENT — no blocking findings found.** I’m not approving automatically for this repo.

What changed:
- `useCustomFilters` can now receive initial filters and `clearCustomFilters()` restores those defaults.
- The Borrow/Multiply index seeds a default `Avail. liquidity > $1K` filter.
- Borrow-page price loading now waits for liquidity/borrowed USD map entries for active pairs, guards stale async loads, and treats individual price failures as zero instead of failing the whole price pass.

Risk assessment:
- User-visible behavior is intentionally narrower: low-liquidity borrow pairs are hidden by default, but the chip is visible and removable.
- The readiness gate is important here: without it, the default filter could evaluate missing liquidity as `0` during load. The new `areBorrowPriceValuesReady` check avoids that transient empty/underfiltered state for active borrow pairs.
- The per-pair `catch(() => 0)` means unresolved price data remains conservative and hides the affected pair under the default filter. That matches the “avoid tiny/unpriced liquidity by default” intent, but it is worth remembering if support reports a missing borrow pair after a price/RPC incident.

- I checked sibling custom-filter consumers on Explore, Lend, and Earn. They still call `useCustomFilters` without defaults, so the shared composable extension preserves existing behavior there.
- The liquidity threshold is borrow-specific, so keeping the default filter declaration in `pages/borrow/index.vue` is reasonable for this patch. I did not find another sibling route that should automatically inherit the same `$1K` borrow-liquidity default.
- Non-blocking maintainability note: since `useCustomFilters` now has default-state semantics, a small focused composable test for “initial filters are present, removable, and restored by clear” would be a useful future guard. The current full test suite does not appear to exercise that reusable behavior directly.

Bot feedback check:
- CodeRabbit had only an in-progress/status comment and no actionable inline review comments at the time I reviewed this head, so there were no bot findings to adopt or refute.

Validation performed:
- `npm run test:run` — passed: 954 tests passed, 1 skipped.
- `npm run build` — passed; only existing/general bundler warnings observed.
- Browser visual smoke on the PR preview `/borrow?network=1`:
  - desktop viewport: Borrow/Multiply rendered real content, default `Avail. liquidity > $1K` chip visible, list populated, no empty-state regression.
  - mobile viewport: same default chip and populated borrow list visible.
  - interaction check: dismissed the intro modal, clicked the default chip, and verified the chip disappeared without causing an empty-state regression.
  - Wallet/signing coverage: not run; this PR changes list filtering only.
  - Exact base-preview comparator was not available during this pass, so screenshots are PR-preview evidence rather than base-vs-PR composites.

Screenshots:

Desktop PR preview, default borrow-liquidity filter visible:

Desktop PR preview, default borrow-liquidity filter visible

Mobile PR preview, default borrow-liquidity filter visible:

Mobile PR preview, default borrow-liquidity filter visible

LeonardEulerXYZ (commented):
Reviewed head `36c2076e5c10cf3d586ec0673d23e1bae6f62836`.

Verdict: **COMMENT — no new blocking Leonard findings.** I’m not approving automatically for this repo.

What changed:
- `useCustomFilters` now supports unavailable metric handling and an `initialFilters` argument.
- `/borrow` seeds a default `Avail. liquidity > $1K` filter and now keeps liquidity/borrowed USD values as `undefined` when price conversion fails.
- Borrow-page loading now waits for per-pair liquidity/borrowed map entries instead of only a global price-ready flag.

Validation:
- `npm run typecheck` ✅
- `npm run test:run` ✅ — 954 passed, 1 skipped
- `npm run build` ✅
- Browser visual smoke ✅ — headed Chromium under Xvfb on `https://dev-build-euler-lite-pr-608.up.railway.app/borrow?network=1`, desktop 1440×1000 and mobile 390×844, compared with the development Railway URL. Route rendered borrow cards, no Nuxt/error shell, no empty-filter state, and no PR-only material console/runtime failure. Wallet/signing smoke was not run.

Visual evidence:
PR #608 borrow page desktop/mobile comparison

- Sibling `useCustomFilters` consumers (`/lend`, `/earn`, `/explore`) keep their existing behavior because they do not pass `initialFilters` and their metric getters still return finite numbers.
- The default borrow-liquidity threshold is localized to `/borrow`, which matches the PR intent; I did not find a missed sibling page that should get the same default filter.
- There is still no focused unit coverage for the new reusable `useCustomFilters` semantics (`initialFilters` + `includeWhenValueUnavailable`). The full suite is green, but this behavior is currently only covered indirectly.

- CodeRabbit’s current comment that the default `initialFilters` participate in matching but are not exposed in returned `customFilters` is valid. That means the `$1K` default filter is active but invisible/non-removable in the `UiCustomFilterChips` UI. If that hidden default is intentional product behavior, I’d suggest making it explicit in copy; if it is meant to behave like a normal custom filter, expose it in the returned active filter state.
- CodeRabbit’s current stale-price-load comment is also directionally valid: `priceLoadId` increments when the debounced callback starts, not when a new borrow-list refresh schedules the next run, so an already in-flight fetch can still write during the debounce window before the next generation exists. This looks transient rather than exploitable, but it is exactly on the pricing/filtering surface this PR changes.

I did not duplicate those as fresh inline comments because they are already anchored on the current head and verified; adding parallel Leonard threads would make the review noisier without adding coordinates.

Seranged (approved):

PR #612, opened 2026-06-23
fix: support reward claim batching

Seranged:
## Summary
- Consume `@eulerxyz/euler-v2-sdk@1.0.9` so Lite gets the Merkl claim-data hydration fix needed for batched reward claims.
- Let standalone reward claim and rEUL unlock plans skip the simulated account snapshot they do not need while still participating in full batch simulation.
- Show queued reward claims in the portfolio with the same simulated removed-state styling used by other batch-modified portfolio cards.

## Changes
- Adds a typed `requiresPlanningAccount: false` batch entry mode for standalone plans and keeps the existing account-aware behavior as the default.
- Marks SDK reward claims and rEUL unlocks as standalone batch entries.
- Tracks queued reward claims with a stable `rewardClaimKey` so the same reward row disables `Claim` / `Add to batch`, displays `In batch`, and re-enables when the batch entry is removed or cleared.
- Surfaces formatted add-to-batch errors in the reward toast with a fallback if error formatting fails.
- Updates `package.json` and `package-lock.json` from `@eulerxyz/euler-v2-sdk@1.0.8` to `1.0.9`.

## Test plan
- [x] `npx eslint components/entities/portfolio/PortfolioSdkRewardItem.vue components/entities/reward/RewardUnlockItem.vue composables/useTxBatch.ts`
- [x] `npx eslint components/entities/portfolio/PortfolioSdkRewardItem.vue`
- [x] `npm run typecheck`


composables/useTxBatch.ts
```diff
@@ -83,6 +83,8 @@ export interface BatchEntry {
 export interface BatchEntryInput extends Omit<BatchEntry, 'id' | 'plan'> {
   /** Builds this entry once, at add-time, against the current batch end-state. */
   buildPlan: (account: Account<IHasVaultAddress>) => Promise<TransactionPlan>
+  /** Set false for standalone plans that do not need the current simulated account. */
+  requiresPlanningAccount?: boolean
```
LeonardEulerXYZ: Non-blocking maintainability suggestion: the new `requiresPlanningAccount: false` path would be safer with a small focused `useTxBatch` test proving `addEntry` skips `getEntryPlanningAccount()` and calls a standalone builder without requiring an account. A discriminated union for account-aware vs standalone entries would also remove the `account as Account<...>` cast below, so a future standalone builder cannot accidentally start depending on the planning account.

LeonardEulerXYZ (commented):
Reviewed current head `e1a1022dac97e0e89cb786db49f794b1defb734d`.

Verdict: COMMENT — no blocking issue found. The reward batch fix is narrow and the runtime path looks consistent: reward claims opt out of the batch planning-account snapshot only at the SDK reward row, while existing account-aware batch entries keep the default behavior.

Validation performed:
- Diff/context review of `PortfolioSdkRewardItem.vue` and `useTxBatch.ts`.
- Sibling-surface pass over other `addBatchEntry` call sites: deposits, withdraws, borrows, repays, savings, multiply, and swap/refinance entries still depend on the simulated planning account and remain covered by the default `requiresPlanningAccount !== false` path.
- `npm run test:run -- tests/composables/useSdkRewards.test.ts tests/composables/useTxBatch.test.ts` — passed.
- `npm run build` — passed, with existing chunk/dynamic-import warnings only.
- Railway preview route smoke, headed Chromium: `/` and `/portfolio` on desktop and mobile viewports returned the real app. This was a disconnected route smoke only; it did not exercise the reward-row Add to batch interaction because that needs a connected wallet/account with claimable rewards and advanced mode enabled.

- The abstraction is appropriately narrow for this PR: only standalone SDK reward claim entries opt out; account-dependent sibling flows remain unchanged.
- One maintainability gap remains: the new opt-out behavior is not directly covered by a focused `useTxBatch` unit test, and the current type uses a cast when `requiresPlanningAccount: false`. I left an inline suggestion for that. Not blocking, but worth tightening while this code is small.

Bot-review pass:
- CodeRabbit’s toast-resilience comment looks non-blocking / likely stale against `getTxErrorMessage`: the helper catches the SDK decoder fallback and returns default strings for unknown errors.
- CodeRabbit’s type-safety concern maps to the same maintainability gap above; I would treat it as a small type/test cleanup rather than a runtime blocker for the current call site.

Smoke coverage: route smoke + mobile smoke; no wallet/signing smoke. Screenshots were not posted because this PR has no standalone visual diff in the disconnected state, and the changed toast/reward action was not meaningfully screenshotable without a reward-bearing wallet fixture.

kanvgupta (dismissed):

kanvgupta (approved):

PR #619, opened 2026-06-24
Add multi-chain vault browsing

Seranged:
## Summary
- Adds multi-chain vault browsing so users can select several supported chains and scan lend, borrow, earn, and explore surfaces together.
- Keeps detail and transaction flows anchored to a current chain while list/catalog views can aggregate selected chains.
- Preserves chain-specific label, registry, category, and refresh state to avoid cross-chain address collisions.

## Changes
- Adds selected-chain state, URL/localStorage sync for `networks`, and a multi-select chain picker in the header.
- Loads labels and vault registries per chain, then merges catalogue data for browsing while retaining chain-scoped cache keys.
- Adds chain filters and chain indicators across vault lists, borrow pairs, and vault detail headers.
- Makes periodic refresh, escrow discovery, unresolved collateral lookup, and unknown-vault resolution chain-aware.

## Test plan
- [x] npm run typecheck
- [x] npm run test:run
- [x] git diff --check origin/development...HEAD

## Summary by CodeRabbit

* **New Features**
  * Added multi-chain selection across the app, including chain filtering in header, explore, lend, borrow, and earn views.
  * Vault and market cards now show the relevant chain logo and chain ID.
  * Search, sorting, and reward views now respect the selected chain(s).

* **Bug Fixes**
  * Improved correctness for labels, rewards, balances, and vault data when the same address exists on multiple networks.
  * Fixed chain-specific navigation and cache collisions so data stays accurate across chains.


composables/useVaultRegistry.ts
```diff
@@ -52,24 +54,44 @@ const pendingResolutions = new Map<string, Promise<VaultEntity | undefined>>()
...
+    .map(([, entry]) => entry)
+  return matches.length === 1 ? matches[0] : undefined
+}
+
 // Get vault entry from registry
-const get = (address: string): VaultEntry | undefined => {
-  return registry.value.get(normalizeAddress(address))
+const get = (address: string, chainId = getDefaultChainId()): VaultEntry | undefined => {
+  const keyed = chainId ? registry.value.get(normalizeRegistryKey(address, chainId)) : undefined
+  return keyed ?? findUniqueEntryByAddress(address)
```
LeonardEulerXYZ, line 79: This fallback is unsafe when the caller supplied an explicit `chainId`. If chain 2 asks for `get(address, 2)` before that chain's entry is loaded, and the same address exists only once in the registry on chain 1, `findUniqueEntryByAddress()` returns the chain-1 vault. The downstream typed helpers (`has`, `getType`, `isVerifiedVault`, escrow/category checks) then treat a wrong-chain entry as valid. For multi-chain browsing this is a data-integrity bug: address-only fallback should only apply to genuinely address-only legacy calls, not after an explicit chain miss. Please add a focused same-address/different-chain registry test as well.

composables/useVaults.ts
```diff
@@ -800,9 +814,32 @@ const loadVaults = async () => {
     }
   }
   finally {
-    if (loadGeneration.value === generation && chainId.value === startChainId) {
+    if (loadGeneration.value === generation) {
```
LeonardEulerXYZ, line 817: This makes the global `isReady` flag true as soon as the first selected chain finishes. `loadVaults()` starts all selected chains concurrently, but every `loadVaultsForChain()` publishes readiness in its own `finally`; consumers such as market groups treat that as the selected multi-chain dataset being ready. A fast chain can therefore render a partial catalogue while slower selected chains are still loading. Please keep readiness false until all target chain loads complete, or derive it from `selectedChainIds.every(id => loadedChainIds.includes(id))`.

composables/useMarketGroups.ts
```diff
@@ -506,11 +526,10 @@ export const useMarketGroups = () => {
     const memberVaults: EVault[] = []
 
     try {
-      const { chainId } = useEulerAddresses()
       const { getEulerSdk } = useEulerSdk()
       const sdk = await getEulerSdk()
       const result = await sdk.eVaultService.fetchVaults(
-        chainId.value,
+        targetChainId,
```
LeonardEulerXYZ, line 532: The on-demand direct-market path fetches from `targetChainId`, but the returned SDK vaults are pushed into `memberVaults` without tagging them with that `targetChainId`. The main vault-loading path now uses chain-tagged vaults, and the rest of this file keys labels/graph data off `vault.chainId`; direct routes for a non-current or `chainId:productKey` market can end up with `undefined` chain IDs and wrong/fallback label lookups. Please assign the target chain to each fetched vault before computing metrics/graph data, and cover the direct-route/on-demand case in a focused test.

pages/explore/index.vue
```diff
@@ -168,10 +191,16 @@ const matchesRiskManagerFilter = (group: MarketGroup): boolean => {
   return getUniqueEntitiesByVaults(group.vaults).some(e => selectedRiskManagers.value.includes(e.name))
 }
 
+const matchesChainFilter = (group: MarketGroup): boolean => {
```
LeonardEulerXYZ, line 194: The new Chain selector is a real filter, but the active-filter/clear-filter path does not include it. On Explore, Lend, Earn, and Borrow, `selectedChains` is applied to the list but omitted from `hasActiveFilters`/`hasClearableFilters` and from the corresponding clear function. If a chain filter hides all results, users can get a misleading `No markets/vaults yet` state with no useful clear action; if another filter exposes `Clear filters`, the chain filter remains applied. Please wire `selectedChains.length` and reset it consistently across all four list pages.

components/layout/TheHeader.vue
```diff
@@ -15,8 +15,13 @@ const { connect } = useWagmi()
...
 // Wagmi account info
 const { address, isConnected } = useWagmi()
-const { chainId, allowedChainIds } = useEulerAddresses()
-const chainLogoSrc = computed(() => getChainLogoUrl(chainId.value))
+const { chainId, selectedChainIds, allowedChainIds } = useEulerAddresses()
+const primarySelectedChainId = computed(() => selectedChainIds.value[0] ?? chainId.value)
+const chainLogoSrc = computed(() => getChainLogoUrl(primarySelectedChainId.value))
+const selectedChainLabel = computed(() => {
+  const count = selectedChainIds.value.length || 1
+  return count === 1 ? String(primarySelectedChainId.value) : `${count} chains`
```
LeonardEulerXYZ, line 23: For a single selected chain this renders the raw numeric chain id (`1`, `8453`, etc.) as the visible header label. The chain modal/options already use chain names, and the PR smoke confirmed the desktop Lend header displays `1`. This is a small but user-visible regression; use the chain registry name with numeric id only as fallback.

LeonardEulerXYZ (changes requested):
## Leonard review — PR #619

Verdict: Request changes.

I reviewed the current head `46a0160` as a multi-chain browsing/risk-assessment pass. The main architecture is moving in the right direction — chain-qualified keys are being threaded through labels, rewards, registry, list filters, and discovery metrics — but I found several correctness/UX issues that should be fixed before merge.

Scope covered:
- data/composables: labels, vault registry, vault loading, token/reward lookups, market groups
- user-facing routes: Explore, Lend, Earn, Borrow, header chain selector, vault cards/detail anchoring
- bot feedback: CodeRabbit currently only has the processing/walkthrough comment on this head; no actionable inline comments were present to verify
- scalability / maintainability hygiene: checked sibling list pages and shared chain-scoped registry/data paths, not only the first touched component

Validation performed:
- `npm test -- --run tests/composables/useEulerAccount.test.ts tests/utils/discovery-calculations.test.ts` — pass, 19 tests
- `npm run typecheck` — pass
- `npm run build` — pass, with existing chunk/module-preload style warnings
- `git diff --check origin/development...HEAD` on the reviewed diff — pass
- Browser visual smoke on the PR Railway preview, headed Chromium under Xvfb:
  - desktop `/lend?network=1&networks=1` — loaded real Lend content; header renders the selected chain as `1`
  - mobile `/lend?network=1&networks=1` — loaded real Lend content; chain chip is icon-only on mobile but text extraction still exposes `1`
  - desktop `/explore?network=1&networks=1,8453` — loaded real Explore content; header renders `2 chains`

Screenshot evidence:
- [Desktop Lend, single selected chain shows numeric `1`](https://gist.githubusercontent.com/LeonardEulerXYZ/3d4a2203a5b861863b2221c06499d1f4/raw/b12f6ce68cde9e2b1cea0cd02bc3d5f754ed5d26/euler-xyz-euler-lite-pr619-pr619-desktop-lend-network1.png)
- [Mobile Lend, same route at 390px](https://gist.githubusercontent.com/LeonardEulerXYZ/3d4a2203a5b861863b2221c06499d1f4/raw/b1b5b5ed8043a7e84aa55c323839753ec53fa68e/euler-xyz-euler-lite-pr619-pr619-mobile-lend-network1.png)
- [Desktop Explore, multi-chain selection shows `2 chains`](https://gist.githubusercontent.com/LeonardEulerXYZ/3d4a2203a5b861863b2221c06499d1f4/raw/914dc09150ac6f41caae49e5a3910b58c628005d/euler-xyz-euler-lite-pr619-pr619-desktop-explore-multichain.png)

Findings are inline. The two highest-risk ones are the explicit-chain registry fallback and the global readiness flag becoming true before all selected chains finish loading. Both can make a multi-chain page look settled while it is using wrong-chain or partial-chain data — a small symbol error with a rather large denominator.

Smoke coverage: browser visual smoke + mobile smoke, no wallet/signing smoke.

PR #622, opened 2026-06-24
feat: add PT asset category

Seranged:
## Summary
- Add PT as a supported token category so Principal Token assets can appear as their own quick filter when token-list data tags them.
- Keep category matching and display label behavior aligned with the existing asset category helpers.

## Changes
- Add the pt category label to the central correlated category allowlist.
- Extend token category tests to cover PT matching, filter option output, and correlation label display.

## Test plan
- [x] npm run test:run -- tests/utils/token-categories.test.ts
- [x] npm run typecheck
- [x] npx eslint utils/token-categories.ts tests/utils/token-categories.test.ts


Seranged:
Closing this draft; PT should likely be handled as an asset-filter-only category rather than part of correlated category logic.

PR #631, opened 2026-06-25
Fix/batch simulation not loaded

dglowinski:
### Problem

Some users hit `Error: Batch simulation not loaded` while building a batch (e.g. add repay → add withdraw on the same position). The batch itself was valid — nothing actually failed — and it couldn't be reproduced in spy mode.

### Root cause

The generic message is only reachable when a resimulation is **superseded mid-flight**. `getEntryPlanningAccount` awaited a single resimulation promise to get the post-cart "final layer" before planning the next op. But both the `entries` watcher and the plan-time call start runs that share `resimToken`, and the loser hits the `token !== resimToken` early-return — which bails **without** populating `layers` or setting `simError`. A single `await` could therefore resolve on the superseded run before the winning run produced the final layer, so planning fell straight through to:

```ts
throw new Error(simError.value ?? 'Batch simulation not loaded')
```

A second, latent bug made this worse: `runResimulate` compared `resimulatePromise === promise` against the raw `resimulate()` promise while storing the `.finally()`-wrapped one, so the guard never matched and `resimulatePromise` was never cleared.

The race window scales with `simulateTransactionPlan` latency and account size, which is why it was intermittent for some users and invisible in spy mode (stable `owner`, faster/lighter sims).

### Changes (`composables/useTxBatch.ts`)

- **`runResimulate`**: compare against the wrapped promise actually stored, so `resimulatePromise` is correctly nulled when a run settles.
- **`getEntryPlanningAccount`**: re-await the current run in a bounded loop, returning once the final layer settles or a real `simError` surfaces, instead of throwing on a transient superseded state.
- **`resimulate`**: add a temporary `logWarn` on the post-sim supersession path to confirm the trigger in the field (logs the `token → resimToken` transition and entry count). Marked for removal once confirmed.

### Testing

- Syntax-validated; full `typecheck`/`lint` to run in CI.
- Suggested manual repro: throttle the RPC/network (DevTools "Slow 3G") and add two+ entries in quick succession — should no longer throw, and the diagnostic log will confirm whether a supersession fired.

### Notes

- Based on `master`; fix regions are identical on `development` if you'd prefer to retarget.
- The added `logWarn` is intentionally temporary — drop it in a follow-up once field logs confirm the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batch resimulation reliability by detecting superseded runs earlier and stopping with clearer error diagnostics.
  * Reduced false “Batch simulation not loaded” failures by retrying until the final planning layer is available (or a terminal simulation error occurs).
  * Improved in-flight tracking to prevent stale state across overlapping resimulation attempts.
* **Tests**
  * Added coverage for the final-layer waiting helper, including retry behavior, in-flight reuse, correct error messaging, and max-attempt failure handling.


LeonardEulerXYZ (changes requested):
Reviewed head `ccc5bb8bae012a71ead75f687e8962eb0d022bc3`.

Verdict: Request changes — the main race fix is directionally right, but the retry loop still has a terminal-iteration race that can throw `Batch simulation not loaded` while a fresh resimulation has just been scheduled and not awaited.

What changed:
- `useTxBatch` now clears `resimulatePromise` against the wrapped promise, which fixes the stale-promise cleanup bug.
- `getEntryPlanningAccount()` now retries when a resimulation is superseded before layers are populated.
- A diagnostic warning was added for superseded simulations.

Blocking finding:
- `composables/useTxBatch.ts:1498` — verified CodeRabbit’s current-head comment as valid. On the final loop iteration, `if (!resimulatePromise) runResimulate()` can start a new resimulation and then the loop exits immediately, throwing `Batch simulation not loaded` without awaiting the run it just scheduled. That preserves a tail version of the same user-facing failure this PR is trying to remove. Bound the kick-off to iterations that have another pass left to await it, or restructure the loop so every scheduled retry is awaited before falling through. A focused regression test for the superseded-run path would be useful here.

Validation performed:
- `npm run test:run -- tests/composables/useTxBatch.test.ts` — passed, 13 tests.
- `npm run build` — passed; only existing bundle/chunk warnings observed.
- Railway preview smoke, headed Chromium under Xvfb: `/`, `/lend?network=1`, and `/borrow?network=1` loaded on desktop and mobile viewports with 200 responses and real app content.

- Sibling surfaces that consume `useTxBatch` include lend/earn withdraw, lend rebalance/swap, borrow/multiply, repay, collateral swap, reward unlock, batch drawer/review modal, and portfolio batch-aware components. Because the changed state gate is centralized in `useTxBatch`, this is the right abstraction point; I did not find a second sibling implementation of the same `resimulatePromise`/planning-account gate that would need a parallel fix.
- The remaining concern is test coverage: current `useTxBatch` tests cover stitching, snapshots, wallet layers, and execution-error clearing, but not the superseded resimulation/retry invariant being changed here. Given this is concurrency-sensitive cart plumbing shared across many action routes, a focused test is worth adding rather than relying on route smoke.

Bot/reviewer feedback:
- CodeRabbit’s actionable comment about the final-iteration retry is current and materially correct. I did not find a stale or misleading bot finding to dismiss.

Smoke coverage:
- Browser visual/route smoke: desktop + mobile route-load smoke on the preview URL.
- Wallet/signing smoke: not run.
- Batch-entry interaction smoke: not run; the remaining blocker is in the composable retry logic and is cheaper/more precise to validate with a focused regression test.
- Screenshots: not posted because this PR has no intended visual UI delta and I did not capture screenshots for evidence.

LeonardEulerXYZ (changes requested):
Reviewed head `52e7992fb8155870001525222f2b1a086358149b`.

Verdict: Request changes — the main fix is still directionally right, but the current retry loop preserves a tail case where `Batch simulation not loaded` can be thrown immediately after scheduling a fresh resimulation that is never awaited.

What changed:
- `runResimulate()` now stores and clears the wrapped promise, fixing the stale-promise cleanup bug.
- `getEntryPlanningAccount()` now accepts `layers.length >= entries.length + 1`, which correctly handles plans that expand to multiple SDK operations.
- It retries superseded simulations instead of relying on a single await.

Blocking finding:
- `composables/useTxBatch.ts:1514` — CodeRabbit’s current-head comment is valid. On the final retry iteration, `if (!resimulatePromise) runResimulate()` can start a new run and then the loop exits, throwing `Batch simulation not loaded` without awaiting the run it just scheduled. This leaves a smaller version of the same user-facing failure this PR is intended to remove. The retry should only schedule work that a later iteration will await, or the loop should be structured so every scheduled resimulation is awaited before falling through. A focused regression test for the superseded-run/no-layer/no-error path would be useful.

Validation performed:
- `npm run test:run -- tests/composables/useTxBatch.test.ts` — passed, 13 tests.
- `npm run build` — passed; only existing bundle/chunk warnings observed.
- Railway preview smoke with headed Chromium under Xvfb: `/batch?network=1` on desktop and mobile, plus `/borrow?network=1` on desktop, loaded with HTTP 200 and real app content. Wallet/signing flow was not exercised.

- This patch is in the shared `useTxBatch` composable, which is the right abstraction point for batch planning state. Sibling surfaces using it include batch review/drawer components and lend, earn, borrow, multiply, repay, withdraw, collateral-swap, borrow-swap, reward-unlock, and portfolio batch-aware flows. I did not find a sibling component carrying a separate copy of this readiness gate.
- The missing part is test coverage: the current `useTxBatch` tests pass but do not pin the final-iteration supersession case, so this race can survive a green suite.

Bot/reviewer feedback:
- Verified the active CodeRabbit inline comment against current code and agree it is actionable. No material stale/false-positive bot feedback found.

Smoke coverage: route/browser smoke + mobile smoke; no wallet/signing smoke. Screenshots were not posted because this PR changes async batch-planning behavior rather than visual layout/copy, and the useful evidence is code-path/test/smoke output rather than a UI screenshot.

composables/useTxBatch.ts
```diff
@@ -1464,20 +1604,50 @@ export const useTxBatch = () => {
...
     if (entries.value.length > 0) {
-      await (resimulatePromise ?? runResimulate())
-      const refreshedFinalLayer = getCurrentFinalLayer()
-      if (refreshedFinalLayer) return refreshedFinalLayer
-      throw new Error(simError.value ?? 'Batch simulation not loaded')
+      // Retry across superseded resimulations until the final layer for the
+      // present entries settles or a real error appears (see awaitFinalPlanningLayer).
+      try {
+        return await awaitFinalPlanningLayer<Account<IHasVaultAddress>>({
+          getFinalLayer: getCurrentFinalLayer,
```
LeonardEulerXYZ, line 1615: This retry helper still depends on `getCurrentFinalLayer()`, whose predicate above is `layers.length === entries.length + 1`. That exact-count check is too strict for valid entries whose fixed plan expands into multiple SDK operations/layers (approval/Permit2 + action, full-repay cleanup, etc.). In that case the final layer exists at the end of `layers`, but this callback returns `undefined`, so `awaitFinalPlanningLayer()` will retry and eventually rethrow the same generic `Batch simulation not loaded`.

Please restore the multi-operation-safe predicate (`layers.length >= entries.length + 1`, returning the last layer) and add a focused regression where one queued UI entry produces extra operation layers before a second planning-account entry is added.

LeonardEulerXYZ (dismissed):
Reviewed head `012bb4c40700887e8a072c64cae9ba628da9a91a`.

Verdict: Request changes. The previous terminal-iteration retry blocker is resolved by extracting `awaitFinalPlanningLayer()` and awaiting every run it starts, but the current head appears to have dropped the multi-operation layer-count fix that was present in the intermediate head.

Blocking finding:
- `getCurrentFinalLayer()` still requires `layers.length === entries.length + 1`. That is too strict for valid entries whose plan expands into multiple SDK operations/layers, e.g. approval/Permit2 + action or full-repay cleanup. In those cases the simulation can produce more than one layer for a single UI batch entry, so the final layer exists but this predicate returns `undefined`; the new retry helper will then keep awaiting/retrying and eventually throw `Batch simulation not loaded` again. The robust predicate should accept `layers.length >= entries.length + 1` and return the last layer, with a focused regression test for a first entry that yields extra operation layers before adding a second planning-account entry.

Validation performed:
- `npm run test:run -- tests/composables/useTxBatch.test.ts` — passed, 17 tests.
- GitHub status checks on current head: CodeRabbit pass; Railway dev-build pass.
- Security/supply-chain sweep: diff is limited to `composables/useTxBatch.ts` and `tests/composables/useTxBatch.test.ts`; no dependency, workflow, package script, credential, external fetch, or privileged-action surface changes. The temporary diagnostics log batch labels/subaccounts/plan item counts, not secrets/signatures.

Notes:
- The CodeRabbit/Leonard final-iteration finding on older heads is fixed in this head.
- Browser/wallet smoke was not rerun for this incremental review; the remaining issue is a pure composable predicate/test-coverage invariant and is cheaper to prove at unit level. Prior preview route smoke on older heads only proved generic load, not this multi-operation planning case.

Seranged (approved):

Seranged (approved):

PR #634, opened 2026-06-26
Fix supply APY tooltip zeroing lending APY on borrow page

kasperpawlowski:
## Summary
- On the Borrow page, a pair row's Supply APY figure (e.g. 4.08%) did not match its Supply APY tooltip, which showed "Lending APY 0.00%" and "Total supply APY 0.00%".
- The row cell computes its value with `getVaultSupplyApy(pair.collateral)`, but the tooltip's modal data gated the same call behind a `'interestRateInfo' in pair.collateral` check. `pair.collateral` is the SDK `EVault` entity, which exposes `interestRates` (a ready-to-use `number`); `interestRateInfo` is the raw VaultLens shape that the entity never carries. So the guard was always false and forced `lendingAPY` to 0.

## Changes
- `VaultBorrowItem.vue`: pass `lendingAPY: getVaultSupplyApy(pair.collateral)` to the supply APY modal — identical to the row cell — removing the mismatched `interestRateInfo` guard and the now-unused `EVault` import/cast.
- The intrinsic-APY input was already correct; this only restores the lending component, so the tooltip's Lending APY and Total now match the displayed figure.

## Test plan
- [x] On the Borrow page, hover the Supply APY tooltip for a pair whose collateral earns lending yield (e.g. USDC/wstETH): "Lending APY" and "Total supply APY" should equal the row's Supply APY figure (e.g. 4.08%), not 0.00%.
- [x] Verify intrinsic-only collateral rows (e.g. escrow wstETH, zero lending) still show the correct total.

## Summary by CodeRabbit

* **Refactor**
  * Simplified how supply APY values are prepared for vault borrowing items.
  * Improved consistency in the APY information shown to users by using a direct calculation path.

### DIFF
--- a/components/entities/vault/VaultBorrowItem.vue
+++ b/components/entities/vault/VaultBorrowItem.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import type { EVault } from '@eulerxyz/euler-v2-sdk'
 import { getUtilisationWarning, getBorrowCapWarning, getCollateralSupplyCapWarning } from '~/composables/useVaultWarnings'
 import { formatAssetValue } from '~/utils/sdk-prices'
 import { getMaxMultiplier, getMaxRoe } from '~/utils/leverage'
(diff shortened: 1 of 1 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `6544a5ae3d410bc65be6d2ad50d8647fcca85dfa` for PR #634.

Verdict: no blocking findings found. I am leaving this as a comment review rather than an approval, per normal Lite review policy.

What changed:
- `VaultBorrowItem.vue` now passes `getVaultSupplyApy(pair.collateral)` directly into the Supply APY modal data instead of guarding on `interestRateInfo` and falling back to `0`.
- This aligns the borrow-page Supply APY modal with the row value and with sibling supply-APY modal patterns that already read through the shared display helper.

Validation performed:
- `git diff --check origin/development...HEAD` passed.
- `npm run typecheck` passed.
- `npm run test:run` passed.
- `npm run build` passed.
- `npx eslint components/entities/vault/VaultBorrowItem.vue` passed.
- `npm run lint` still fails on existing repo-wide lint issues outside the changed file; I did not classify that as introduced by this PR.
- Railway preview smoke on `/borrow?network=1` opened the Supply APY modal from a borrow pair row on desktop and mobile, and verified the modal text includes `Lending APY`, `Intrinsic APY`, `Source`, and `Total supply APY` with the row's non-zero Supply APY.

Visual smoke evidence:

Desktop borrow-page Supply APY modal:
Desktop borrow-page Supply APY modal

Mobile borrow-page Supply APY modal:
Mobile borrow-page Supply APY modal

- Searched sibling Supply APY modal data sites across portfolio, earn, vault list, and overview components.
- Most EVK-backed surfaces already route through `getVaultSupplyApy(...)` or a local computed value derived from it; the old borrow-page `interestRateInfo` guard was the inconsistent one.
- Securitize-specific overview surfaces still intentionally pass `0` with comments because those vaults do not have interest-rate lending APY in that context; I treated those as out-of-scope, not missed copies of this bug.
- No one-off helper or duplicated business rule was added by this PR. A focused unit test for this display/modal mapping would be nice future coverage, but the tiny diff does not create a new abstraction that requires one.

Bot/reviewer feedback:
- CodeRabbit only posted its walkthrough/status comment; I did not find actionable inline bot findings to verify or carry forward.

Coverage caveat:
- This was disconnected visual smoke only; no wallet/signing flow was exercised.

Seranged (approved):

PR #639, opened 2026-06-26
fix(batch): don't surface dust/self collateral rows in simulated borrow

dglowinski:
## Problem

In the **simulated** batch account, a borrow position showed phantom collaterals the real/lens view never lists — e.g. repaying on the `…bfc` USDC/PYUSD position made the collateral read **"PYUSD & others"** instead of just `PYUSD`.

This only affects the simulated account: `hydrateBorrowLiquidity` runs solely from `stitchAccount`, never the base/real portfolio (which gets a clean `liquidity.collaterals` from the SDK lens).

## Root cause — two distinct sources

`hydrateBorrowLiquidity` built the collateral list partly from the EVC `enabledCollaterals` set (added in 50cce753 to surface a collateral-swap's newly-enabled collateral), but it added entries the lens deliberately omits:

1. **Enabled-but-empty dust collaterals.** An account can carry several leftover EVC enablements (here: `eUSDC-2/22/70`, PT-wstUSR, etc., all zero balance). They were added as zero-value rows.
2. **The borrow vault listing itself as collateral** — the actual cause of the visible "& others" on `…bfc`. The EVC has `eUSDC-80` (the borrow vault) enabled as a collateral, and its own borrow position carries debt, so it passed the `hasPositionValue` guard. The SDK lens never includes a vault in its own collateral list.

Confirmed live via instrumentation: `outCollaterals = [ePYUSD-6 (usd 18.5M), eUSDC-80 (usd 0)]`.

## Fix (`composables/useTxBatch.ts`)

- Thread the **pre-batch base enabled set** (`existing.enabledCollaterals`, still prevFull at that point in `stitchAccount`) into `hydrateBorrowLiquidity`. The enabled-collateral loop now adds unconditionally only when a collateral is **newly enabled by the batch** (not in the base set); pre-existing enablements go through the same balance/value guard as the lens collaterals, so empty dust is dropped.
- Skip the borrow vault itself: `if (key === borrowVaultKey) return` — a vault is never its own collateral.

The original collateral-swap behaviour (surfacing a genuinely newly-enabled collateral) is preserved.

## Tests

`tests/composables/useTxBatch.test.ts`:
- "omits a pre-existing enabled-but-empty collateral from simulated borrow liquidity"
- "never lists the borrow vault as its own collateral"

Both reproduce the bug (fail without the fix) and pass with it. Full file: **15/15 green**, eslint clean. The existing "hydrates newly enabled collateral" (collateral-swap) test still passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved simulated borrow-liquidity calculations so collaterals enabled before a batch are handled differently from collaterals enabled by that batch.
  * Prevented empty collateral rows from appearing for previously enabled assets with no balance.
  * Ensured the borrow vault is not shown as a collateral asset in simulated results, even when it’s enabled and the position has debt.

### DIFF
--- a/composables/useTxBatch.ts
+++ b/composables/useTxBatch.ts
@@ -798,6 +798,7 @@ const hydrateBorrowLiquidity = (
   borrow: StitchPosition,
   positions: StitchPosition[],
   enabledCollaterals: Address[] | undefined,
+  baseEnabledCollaterals: Address[] | undefined,
 ): void => {
   const liquidity = borrow.liquidity
   if (!liquidity || (borrow.borrowed ?? 0n) === 0n) return
@@ -805,6 +806,7 @@ const hydrateBorrowLiquidity = (
   const positionsByVault = positionByVault(positions)
   liquidity.vault = liquidity.vault ?? borrow.vault
   liquidity.liabilityValueUsd = borrow.borrowedValueUsd
+  const borrowVaultKey = getAddress(borrow.vaultAddress).toLowerCase()
 
   const existingCollaterals = new Map<string, StitchLiquidityCollateral>()
   for (const collateral of liquidity.collaterals) {
@@ -812,11 +814,23 @@ const hydrateBorrowLiquidity = (
   }
   const enabledCollateralKeys = enabledCollaterals?.map(address => getAddress(address).toLowerCase())
   const enabledCollateralSet = enabledCollateralKeys ? new Set(enabledCollateralKeys) : undefined
+  // Collaterals that were already enabled on the EVC *before* this batch ran.
+  // Used to tell a collateral newly enabled by the batch (e.g. a collateral
+  // swap) apart from a pre-existing, possibly empty, enablement.
+  const baseEnabledCollateralSet = baseEnabledCollaterals
+    ? new Set(baseEnabledCollaterals.map(address => getAddress(address).toLowerCase()))
+    : undefined
   const collateralAddresses: Address[] = []
   const seen = new Set<string>()
   const addCollateralAddress = (address: string, requireEnabledOrValue: boolean) => {
     const key = getAddress(address).toLowerCase()
     if (seen.has(key)) return
+    // A vault is never its own collateral. The EVC can list the borrow vault in
+    // enabledCollaterals, and its own position carries debt (so it passes the
+    // value guard below), but the controller grants it no LTV — the lens omits
+    // it from liquidity.collaterals and so must we, or it surfaces as a spurious
+    // zero-value collateral row.
+    if (key === borrowVaultKey) return
     const collateralPosition = positionsByVault.get(getAddress(address))
     const hasBalance = hasPositionValue(collateralPosition)
     const isEnabled = enabledCollateralSet?.has(key) ?? true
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed head `4d40c4a62e9223358c685119d11638aaa201a648`.

Verdict: no blocking findings. I’m leaving this as a COMMENT review, not an approval.

What changed:
- `hydrateBorrowLiquidity` now distinguishes pre-batch enabled collaterals from collaterals newly enabled by the batch.
- Pre-existing empty enablements must pass the same value/lens guard as ordinary collaterals, while newly enabled collaterals still surface for collateral-swap flows.
- The borrow vault is explicitly excluded from its own simulated collateral list.

Risk/behavior assessment:
- Affected users/routes: simulated batch state for borrow/repay/collateral-swap-style flows that render `PortfolioBorrowPosition.collateralVaults` after a transaction-plan simulation.
- Security/trust boundary: no new signing path, network call, dependency, workflow, or privilege surface was introduced. This is state projection/display logic inside the existing batch simulation path.
- Main regression risk was accidentally hiding genuinely newly enabled collateral from collateral-swap simulations. Existing coverage plus the new regression tests cover that branch.

Validation performed:
- Inspected PR metadata, current diff, surrounding `stitchAccount` / `hydrateBorrowLiquidity` code, tests, existing GitHub comments/reviews, and CodeRabbit feedback.
- `npm run test:run -- tests/composables/useTxBatch.test.ts` — 15/15 passed.
- `npx eslint composables/useTxBatch.ts tests/composables/useTxBatch.test.ts` — passed.
- `npm run build` — passed, with existing bundle/chunk warnings only.
- Browser smoke on the PR preview, desktop and mobile: opened `/position/1/repay?spy=0x045f6dcd770b1a8ca153cf821906e008de6c7400&network=1`; verified the repay route rendered real content with no app error shell. On desktop, switched to “From collateral”, entered `0.001`, and observed quotes plus the expected “Cannot submit / restore account health” guard rather than a route/runtime failure.

- I searched sibling consumers of the same simulated account/collateral data (`useTxBatch`, portfolio borrow display, repay/collateral-swap consumers, batch health summary, ROE helpers). The fix is placed in the central stitching/hydration path, so the visible borrow-position label and sibling batch consumers inherit the same corrected collateral set instead of each route needing a local filter.
- Tests cover both removed-regression cases and preserve the existing newly-enabled-collateral behavior, which is the important reusable invariant.
- I did not find a missed sibling surface carrying a separate copy of the old behavior.

Bot feedback:
- CodeRabbit’s nit about the redundant guard at `addCollateralAddress` line 837 is valid but non-blocking; line 838 fully subsumes it. I’m not duplicating that as a blocking Leonard finding.
- CodeRabbit’s ESLint install warning is not a PR code finding; local lint on the touched files passed.

Smoke coverage: browser route/interaction smoke + mobile route smoke. Wallet/signing smoke was not run. Screenshots were not posted because this PR changes simulated data projection rather than a layout/visual state, and I did not capture screenshot artifacts.

Seranged (approved):

PR #640, opened 2026-06-26
feat: use monadscan

kanvgupta:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Monad network’s block explorer information so it now points to MonadScan with the correct website.
  * Improved chain lookup so searching by the Monad name resolves to the correct network ID.

* **Tests**
  * Added coverage for the updated Monad explorer details.
  * Added a check for name-based chain resolution.

### DIFF
--- a/entities/chainRegistry.ts
+++ b/entities/chainRegistry.ts
@@ -44,6 +44,24 @@ const LEGACY_CHAIN_ALIASES: ReadonlyMap<string, number> = new Map([
   ['lineamainnet', 59144],
 ])
 
+const MONAD_CHAIN_ID = 143
+const MONAD_EXPLORER_URL = 'https://monadscan.com/'
+
+const monad = chainMap.get(MONAD_CHAIN_ID)
+if (monad) {
+  chainMap.set(MONAD_CHAIN_ID, {
+    ...monad,
+    blockExplorers: {
+      ...monad.blockExplorers,
+      default: {
+        ...monad.blockExplorers?.default,
+        name: 'MonadScan',
+        url: MONAD_EXPLORER_URL,
+      },
+    },
+  })
+}
+
 export const getChainIdByName = (name: string): number | undefined => {
   const normalized = name.trim().toLowerCase().replace(/\s+/g, '-')
   if (!normalized) return undefined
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed PR #640 at head b72566dac5f99ecb296c034ce01d91c147844ccf.

Scope reviewed:
- `entities/chainRegistry.ts`: Monad chain override sets the default block explorer to MonadScan.
- `tests/utils/chain-registry.test.ts`: coverage for the Monad explorer override and `monad` slug resolution.
- Consumer pass: explorer links flow through `utils/block-explorer.ts`, which reads `getChainById(chainId).blockExplorers.default.url`; affected UI surfaces are existing address/tx explorer links rather than a new layout component.

Validation performed:
- Focused test: `npm run test:run -- tests/utils/chain-registry.test.ts` — passed.
- Local throwaway regression test for `getExplorerLink(..., 143)` tx/address URL construction — passed, then removed.
- Typecheck: `npm run typecheck` — passed.
- Build: `npm run build` — passed, with existing-style bundler/chunk warnings only.
- Browser smoke: headed Chromium via Xvfb, desktop and mobile, `/explore?network=monad` normalized to `network=143`, persisted chain id `143`, and rendered the app without page errors. This was route/mobile smoke, not wallet/signing coverage.
- Domain check: the installed Reown/AppKit Monad definition already includes Monadscan as a secondary explorer, and Chainlist lists `https://monadscan.com` for chain 143. Direct MonadScan HTTP probes returned 403 from this environment, so I treated that as external site bot/access behavior rather than an app regression.

- The patch uses the shared chain registry, so existing explorer-link consumers inherit the new default without one-off per-component URL logic.
- I searched sibling explorer consumers (`SwapTokenSelector`, vault overview address blocks, hook info, discovery matrix, wallet disconnect modal) and found they route through `getExplorerLink`/`getChainById`, so this is not a single-component fix.
- No new reusable display/business helper appears necessary for this size of override. A permanent `utils/block-explorer` unit test for chain-specific tx/address URL generation would be a nice future hardening step, but I do not consider it blocking because the registry behavior itself is covered and the shared link path was locally regression-tested.

Bot/reviewer feedback:
- CodeRabbit reported no actionable comments. I found no material stale or misleading bot finding to call out.

Screenshots:
- Not posted. The changed behavior is the generated explorer URL, not a visible layout/copy delta; the browser pass verified route/mobile boot and network normalization, while the URL invariant was validated by tests.

Seranged (approved):

PR #662, opened 2026-07-01
fix: redact upstream URL paths in logs

Seranged:
## Summary
- Prevent upstream URL path segments from being emitted in structured log metadata, since provider URLs can embed secrets in paths.
- Make subgraph upstream failure logging tolerate malformed configured URLs and still return the normalized 502 response.

## Changes
- Reduced safeUrlLogFields output to upstreamHost and query-key metadata only.
- Kept caller-owned static path templating available through safePathTemplate.
- Switched the subgraph proxy catch handler to safeUrlLogFields and added route coverage for malformed configured upstream URLs.

## Test plan
- [x] npx vitest run tests/server/observability.test.ts tests/server/subgraph-proxy-route.test.ts
- [x] npx eslint server/utils/observability.ts 'server/api/proxy/subgraph/[chainId].post.ts' tests/server/observability.test.ts tests/server/subgraph-proxy-route.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subgraph proxy error handling so malformed upstream URLs return a sanitized 502 response.
  * Reduced sensitive URL details in warning logs while preserving useful host and query information.

* **Tests**
  * Added coverage for observability logging to ensure path tokens are not exposed.
  * Added route-level tests for malformed upstream configuration and sanitized failure behavior.

### DIFF
--- a/server/api/proxy/subgraph/[chainId].post.ts
+++ b/server/api/proxy/subgraph/[chainId].post.ts
@@ -35,7 +35,7 @@ import {
 } from 'h3'
 import { createRateLimiter } from '~/server/utils/rate-limit'
 import { logger } from '~/server/utils/logger'
-import { safePathTemplate, urlHost } from '~/server/utils/observability'
+import { safeUrlLogFields } from '~/server/utils/observability'
 import {
   createProxyCache,
   createProxyInFlight,
@@ -114,13 +114,11 @@ export default defineEventHandler(async (event) => {
     return res.body
   }
   catch (err) {
-    const targetUrl = new URL(target)
     logger.warn(
       {
         ctx: 'subgraph-proxy',
         chainId,
-        upstreamHost: urlHost(target),
-        pathTemplate: safePathTemplate(targetUrl.pathname),
+        ...safeUrlLogFields(target),
         err,
       },
       'upstream failed',
--- a/server/utils/observability.ts
+++ b/server/utils/observability.ts
@@ -6,7 +6,6 @@ const ADDRESS_RE = /^0x[a-fA-F0-9]{40}$/
 
 export interface SafeUrlLogFields {
   upstreamHost?: string
-  pathTemplate?: string
   searchKeys: string[]
 }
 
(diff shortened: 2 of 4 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `dd30b0ad6d0459e6bbb3b3c7f5a6619ed58d160d`.

Verdict: non-blocking comment — no issues found.

Scope reviewed:
- `server/utils/observability.ts`: `safeUrlLogFields()` now logs host + query-key names only, not upstream URL path templates.
- `server/api/proxy/subgraph/[chainId].post.ts`: upstream failure logging now uses the shared safe URL helper and no longer reparses `target` in the catch path.
- Tests covering path-embedded token redaction and malformed subgraph upstream handling.

Validation performed:
- Diff and surrounding server/proxy context reviewed.
- Focused tests: `npm run test:run -- tests/server/observability.test.ts tests/server/subgraph-proxy-route.test.ts` — passed.
- Typecheck: `npm run typecheck` — passed.
- Build: `npm run build` — passed; only existing bundle/chunk-style warnings observed.
- Route/API smoke: built server returned sanitized 502 for `POST /api/proxy/subgraph/1` with a synthetic failing upstream containing path/query sentinels.

Security / abuse review:
- The previous subgraph catch path could expose env-derived upstream URL path segments and could throw while trying to log malformed targets. This patch removes both behaviours for the subgraph proxy.
- `safeUrlLogFields()` still records `upstreamHost` and query-key names, not query values; that preserves useful operational signal without logging provider path segments.
- No new network surfaces, auth changes, dependency changes, workflow changes, or client-visible UI changes were introduced.

- Searched sibling proxy/logging surfaces that use `safeUrlLogFields`, `safePathTemplate`, `pathTemplate`, `upstreamHost`, and `searchKeys`.
- The reusable helper is the right abstraction for env-derived upstream URLs, and the subgraph route now uses it instead of one-off parsing.
- Fuul/Incentra upstream failure paths already route through `safeUrlLogFields()` via the shared proxy helper. Merkl/V3/body-limit path templates are request-route or static/public-template logging surfaces rather than env-derived secret-bearing upstream URLs, so I classify them as covered/out of scope for this fix rather than missed siblings.
- Focused regression coverage exists at both helper and route level; no extra shared utility extraction needed.

Bot feedback reviewed:
- CodeRabbit reported no actionable comments. I verified that against the current diff; no material stale or misleading bot finding to call out.

Smoke coverage: route/API smoke only. Browser/mobile smoke and screenshots were not applicable because this is a server-side observability/log-redaction change with no user-visible UI surface.

VSBDev (approved):

PR #663, opened 2026-07-01
fix: preserve client error fingerprints

Seranged:
## Summary
- Recompute sanitized client error fingerprints after the server preserves inbound message, name, and error metadata.
- Prevent distinct client-side failures from collapsing into the same server-side fingerprint or being sampled differently after upload.

## Changes
- Extracted the client observability fingerprint projection into one helper shared by browser-created and server-sanitized payloads.
- Recomputed the sanitized payload fingerprint after copying allowlisted inbound error details.
- Added regression coverage for two uploaded execution failures with the same client fingerprint but different sanitized error details.

## Test plan
- [x] npx vitest run tests/utils/client-observability.test.ts tests/server/client-error.test.ts
- [x] npx eslint utils/client-observability.ts tests/utils/client-observability.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how client observability fingerprints are generated after input sanitization, helping ensure consistent and distinct tracking for similar errors.
  * Fixed fingerprint handling so sanitized events now produce the expected short hexadecimal format.

* **Tests**
  * Added coverage to verify sanitized observability data recalculates fingerprints correctly and keeps different errors separated.

### DIFF
--- a/tests/utils/client-observability.test.ts
+++ b/tests/utils/client-observability.test.ts
@@ -70,6 +70,27 @@ describe('client observability payloads', () => {
     expect(JSON.stringify(payload)).not.toContain('private-rpc')
   })
 
+  it('fingerprints sanitized server input after preserving error details', () => {
+    const first = sanitizeClientObservabilityInput({
+      source: 'client',
+      event: 'tx_execute_failed',
+      fingerprint: 'client-sampled',
+      message: 'first rpc failure',
+      error: { kind: 'rpc-http', name: 'HttpRequestError', shortMessage: 'HTTP request failed', isTransport: true },
+    })
+    const second = sanitizeClientObservabilityInput({
+      source: 'client',
+      event: 'tx_execute_failed',
+      fingerprint: 'client-sampled',
+      message: 'second rpc failure',
+      error: { kind: 'rpc-timeout', name: 'TimeoutError', shortMessage: 'Timed out', isTransport: true },
+    })
+
+    expect(first?.fingerprint).toMatch(/^[0-9a-f]{8}$/)
+    expect(second?.fingerprint).toMatch(/^[0-9a-f]{8}$/)
+    expect(first?.fingerprint).not.toBe(second?.fingerprint)
+  })
+
   it('rejects forged error kind values from client input', () => {
     const payload = sanitizeClientObservabilityInput({
       source: 'client',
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `6b27c00994b35c803fd01d0e7835072da146e968`.

Verdict: COMMENT — no blocking issues found.

Scope reviewed:
- `utils/client-observability.ts`
- `tests/utils/client-observability.test.ts`
- Surrounding server ingest path in `server/api/client-error.post.ts`
- Current bot feedback on the PR

What changed:
- The fingerprint projection is now centralized in `clientPayloadFingerprint()`.
- Browser-created payloads and server-sanitized uploaded payloads now use the same projection after allowlisted `message`, `name`, and `error.kind` details are preserved.
- This avoids collapsing distinct sanitized client failures into the same server-side fingerprint and keeps sampling/deduplication tied to the sanitized observability payload rather than to the pre-preservation skeleton.

Validation performed:
- `npx vitest run tests/utils/client-observability.test.ts tests/server/client-error.test.ts` — passed, 7 tests.
- `npx eslint utils/client-observability.ts tests/utils/client-observability.test.ts` — passed.
- Local-only throwaway Vitest invariant check — passed, then removed. It verified:
  - a browser-created payload keeps the same fingerprint after server sanitization;
  - `error.kind` alone differentiates fingerprints when the other tested fields match.

- The fix is a shared helper inside the existing observability utility, not a one-off callsite patch.
- Sibling consumers found: browser emitters via `reportClientEvent()` and server ingest via `sanitizeClientObservabilityInput()` / `shouldSampleClientPayload()`. Both now route through the same fingerprint projection.
- No duplicated formatter/business logic or missed sibling observability route found in the reviewed surface.
- The helper remains private, which is reasonable because the public contract is the normalized/sanitized payload, not the hash primitive.

Bot feedback:
- CodeRabbit’s nitpick that the committed regression varies both `message` and `error.kind` is directionally fair as a test-isolation point, but I do not see it as blocking. The local throwaway invariant check verified that `error.kind` alone changes the sanitized fingerprint.

Smoke coverage:
- Browser/mobile smoke: not run — this PR changes utility/server-side observability fingerprinting only and has no user-visible route, layout, interaction, or screenshotable UI surface.
- Screenshots: not applicable.

No inline findings from me.

VSBDev (approved):

PR #676, opened 2026-07-03
fix: preserve vault cache on chunk failure

Seranged:
## Summary
- Fail the server vault snapshot refresh when a chunked EVault fetch throws instead of returning a partial result.
- Log chunk fetch failures with generic chunk metadata so the warm-cache failure is visible without listing vault addresses.
- Add a regression test proving a chunk failure rejects the refresh and preserves the previous cached snapshot.

## Validation
- `npm run test:run -- tests/server/vaults-cache.test.ts`
- `npm run typecheck`
- `npm run test:run` - 115 files passed, 1 skipped; 1083 tests passed, 1 skipped
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability during chunked cache refreshes by keeping the existing snapshot if a later fetch chunk fails.
  * Added clearer warning logs with structured context when a chunk fetch fails.

* **Tests**
  * Added coverage for failed chunk refreshes to verify the cache is preserved and the warning is emitted as expected.

### DIFF
--- a/server/utils/vaults-cache.ts
+++ b/server/utils/vaults-cache.ts
@@ -192,7 +192,14 @@ const fetchEVaultsForSnapshot = async (
       errors.push(...(fetched.errors as unknown[]))
     }
     catch (err) {
-      errors.push(err)
+      logger.warn({
+        ctx: 'vaults-cache',
+        chainId,
+        chunkIndex: i / EVAULT_FETCH_CHUNK_SIZE,
+        chunkSize: chunk.length,
+        err,
+      }, 'eVault chunk fetch failed')
+      throw err
     }
 
     if (i + EVAULT_FETCH_CHUNK_SIZE < addresses.length) {
--- a/tests/server/vaults-cache.test.ts
+++ b/tests/server/vaults-cache.test.ts
@@ -1,0 +1,120 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type { SerialisedSnapshot } from '~/server/utils/vaults-cache'
+
+const VAULTS = [
+  '0x0000000000000000000000000000000000000001',
+  '0x0000000000000000000000000000000000000002',
+  '0x0000000000000000000000000000000000000003',
+  '0x0000000000000000000000000000000000000004',
+  '0x0000000000000000000000000000000000000005',
+  '0x0000000000000000000000000000000000000006',
+  '0x0000000000000000000000000000000000000007',
+] as const
+
+const mocks = vi.hoisted(() => ({
+  fetchVaults: vi.fn(),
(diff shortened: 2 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `1267c08e0a5888451f43fc6ad3000580ee3c76ee`.

Verdict: non-blocking comment — I did not find a correctness or security blocker.

What changed:
- The server vault snapshot chunk loader now treats a thrown chunk fetch as a refresh failure instead of returning a partial EVault result.
- The failure path logs generic chunk metadata (`chainId`, `chunkIndex`, `chunkSize`) without listing vault addresses.
- A regression test covers the important invariant: a failed chunked refresh rejects and leaves the previous `vaultsCache` snapshot intact.

Validation run:
- `npm run test:run -- tests/server/vaults-cache.test.ts` — passed
- `npm run test:run` — 115 files passed, 1 skipped; 1083 tests passed, 1 skipped
- `npm run build` — passed, with existing Nuxt/Vite chunking/sourcemap warnings only

Risk / behavior assessment:
- This is the right failure mode for the server snapshot cache: partial EVault data should not overwrite the last-known-good snapshot, because downstream vault lists/market views would otherwise render a seemingly complete but silently truncated universe.
- The request handler already serves `vaultsCache.get(...) ?? vaultsCache.getStale(...)` before attempting a cold refresh, so warm-cycle failures preserve user-visible data within the existing staleness ceiling.
- Cold start with no usable snapshot will now fail closed with the existing upstream-load error rather than publish a partial snapshot. That is a safer invariant for money-facing market data.
- The log addition avoids address leakage while still leaving enough operational signal to identify the failing chain/chunk.

- Searched the sibling chunking path in `composables/useVaults.ts`. The browser-side loader is intentionally more resilient: it records per-vault `DataIssue`s and can continue rendering successful chunks, whereas the server snapshot cache needs all-or-nothing semantics to avoid poisoning the shared snapshot. I do not see a missed shared-abstraction issue here; the different behavior is domain-specific.
- The fix stays in the shared server snapshot loader used for both normal EVault and escrow snapshot fetches, so it is not a one-off component patch.
- The focused regression is attached to the reusable server refresh path, not an incidental UI surface.

- CodeRabbit had only its generated in-progress/walkthrough comment when I inspected the PR; no actionable bot finding needed verification or rebuttal.

Smoke coverage:
- Browser/mobile smoke: not run — this PR changes server-side cache refresh semantics and adds a focused server regression; no UI layout or interaction surface changed.
- Screenshots: not applicable for the same reason.

VSBDev (approved):

kanvgupta (approved):

PR #686, opened 2026-07-04
fix: escape "<" in inline __APP_CONFIG__ script payload

Seranged:
## Problem

The Nitro app-config plugin injects operator-controlled env vars into the HTML head as an inline script:

```
<script>window.__APP_CONFIG__=${JSON.stringify(appConfig)}</script>
```

`JSON.stringify` does not escape `<`, so a config value containing the literal string `</script>` would terminate the inline script element early and break out of the script context. Every value here comes from operator-controlled env vars, so the practical severity is low, but relying on well-formed env values is not a defense — the standard hardening is to make the payload structurally unable to escape the tag.

## Solution

Added a small `escapeScriptJson` helper (in the same helper-function style as the existing `escapeHtml`) that is applied to the stringified payload before it is embedded:

- Escapes `<` to `<` so no substring can form `</script>`.
- Also escapes U+2028 / U+2029 (line/paragraph separators), which are valid inside JSON strings but were invalid inside JS string literals before ES2019 — escaping them keeps the inline script robust.

These are unicode escape sequences, so JSON/JS semantics are unchanged: the values parse back to the exact original characters. The escaping is applied only to the inline-script path; the meta-patch path continues to use the existing `escapeHtml` helper.

## Verification

- Added regression tests in `tests/server/security.test.ts` covering the inline `__APP_CONFIG__` payload:
  - A value containing `</script><script>alert(1)</script>` no longer produces any `</script>` in the escaped payload, and the emitted tag contains exactly one (the real) closing tag.
  - U+2028 / U+2029 are escaped and no raw separators remain.
  - The escaped payload round-trips: `JSON.parse` yields the original object, proving semantics are preserved.
- `npx vitest run tests/server/security.test.ts` — 20 passed.
- eslint clean on both changed files.

### DIFF
--- a/server/plugins/app-config.ts
+++ b/server/plugins/app-config.ts
@@ -28,6 +28,18 @@ function escapeHtml(str: string): string {
     .replace(/'/g, '&#39;')
 }
 
+// JSON.stringify does not escape `<`, so a config value containing `</script>`
+// would break out of the inline script context. Escaping `<` (and the U+2028 /
+// U+2029 line separators, which are invalid in JS string literals) as unicode
+// escapes keeps the payload inside the script tag while preserving identical
+// JSON/JS semantics — `<` parses back to `<` inside string values.
+export function escapeScriptJson(json: string): string {
+  return json
+    .replace(/</g, '\\u003c')
+    .replace(/\u2028/g, '\\u2028')
+    .replace(/\u2029/g, '\\u2029')
+}
+
 function env(key: string, ...fallbackKeys: string[]): string {
   for (const k of [key, ...fallbackKeys]) {
     if (process.env[k]) return process.env[k]!
@@ -107,7 +119,7 @@ function injectSocialImage(html: { head: string[] }, socialImageUrl: string) {
 
 export default defineNitroPlugin((nitroApp) => {
   const appConfig = readAppConfig()
-  const scriptTag = `<script>window.__APP_CONFIG__=${JSON.stringify(appConfig)}</script>`
+  const scriptTag = `<script>window.__APP_CONFIG__=${escapeScriptJson(JSON.stringify(appConfig))}</script>`
 
   nitroApp.hooks.hook('render:html', (html) => {
     html.head.push(scriptTag)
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
## Leonard review — PR #686

Verdict: COMMENT — no blockers found for head `b430b054fbc40fe324745b3bd1d94a923bdc4f3a`.

Scope reviewed:
- `server/plugins/app-config.ts`: new `escapeScriptJson()` hardening before embedding `window.__APP_CONFIG__` in an inline script.
- `tests/server/security.test.ts`: regression coverage for `</script>` breakout, U+2028/U+2029 escaping, and JSON round-trip semantics.

Validation performed:
- `npx vitest run tests/server/security.test.ts` — passed, 20 tests.
- `npx eslint server/plugins/app-config.ts tests/server/security.test.ts` — passed.
- `npm run build` — passed; only existing-style bundler/chunk warnings observed.
- Built-server runtime smoke with an `APP_TITLE` containing `</script><script>alert(1)</script>` — the emitted `__APP_CONFIG__` script had exactly one closing `</script>`, contained escaped `\u003c` sequences, and parsed back to the original value.

Security assessment:
- The changed path addresses the right boundary: JSON is still valid JavaScript/JSON, but literal `<` can no longer form `</script>` inside the HTML parser.
- Existing HTML meta/title path still uses `escapeHtml`, so the PR does not conflate HTML escaping and script-JSON escaping.
- No new network, dependency, workflow, permission, credential, or telemetry surface was introduced.

- Searched sibling inline-render hooks and config consumers. `window.__CHAIN_CONFIG__` is also emitted via an inline `JSON.stringify(...)` script, but its current payload is numeric arrays derived from validated chain IDs, so I do not see a current equivalent breakout issue there.
- Non-blocking future hygiene: if another inline JSON payload begins carrying string values, prefer moving this helper into a shared server utility and reusing it rather than recreating a one-off local helper.

Bot/reviewer feedback:
- No active CodeRabbit/other bot inline comments found to verify or refute.

Smoke coverage: route/API runtime smoke only; browser/mobile visual screenshots were not applicable because this is server-side inline-script hardening with no user-visible UI state change.

No inline findings from me. Small surface, sharp invariant. :leonard:

kanvgupta (approved):

PR #698, opened 2026-07-06
fix: stop oracle adapter reloads from freezing explore matrix

Seranged:
## Problem

Selecting a cell in the explore matrix **Oracles** view froze the whole page — no error logged — for markets whose oracle adapters are not present in the oracle-checks dataset (e.g. Poppie Ondo Markets on BSC, whose pairs are priced by custom adapters).

Reproduced and traced with an instrumented dev build; it is an infinite reactive update loop cycling every ~7 ms:

1. The mounted oracle-details block calls `loadOracleAdapter` for each route step. For an adapter missing from the dataset the lookup misses, so it fell through to `loadAllOracleAdapters`, which re-fetched the chain's adapter map (served from the SDK's in-memory cache) and **replaced `oracleAdaptersRef` with a fresh object on every call**.
2. `useEulerOracleAdapters()` built its `toReactive(computed(...))` view **per call**. `toReactive` wraps the proxy in `reactive()`, whose `isReadonly` probe reads a property through the proxy — a reactive read of the underlying computed at construction time. That subscribes whatever effect is currently running. Market-group computation reaches `useEulerLabels()` (→ `useEulerOracleAdapters()`) mid-evaluation via governor verification, so `marketGroupsSync` became a subscriber of the adapter store.
3. Store rewrite → market groups recompute → new `markets` array → accordion re-renders → the oracle block's `routeSteps` watch refires (its `collateral-vaults` prop is a fresh array per render) → another lookup miss → back to 1.

Vue's "Maximum recursive updates" guard is dev-only, so production builds spin silently — the tab just hangs.

## Fix

Both legs of the loop are removed in `composables/useEulerOracleAdapters.ts`:

- `loadAllOracleAdapters` returns early when the chain's map is already loaded. The dataset is fetched whole per chain, so a lookup miss can never be resolved by reloading; the refetch only churned every subscriber. Chain switches still reload, and a failed load leaves the chain id unset so it retries.
- The `toReactive(computed(...))` view is now a module-level singleton, so calling the composable inside a running effect no longer registers that effect as a dependent of the store.

Either change alone breaks the cycle; together they remove both the redundant trigger and the accidental coupling.

## Verification

- Playwright repro against a local dev build: before the fix, clicking an oracle cell on the Poppie Ondo matrix left the main thread permanently unresponsive (even `page.evaluate(() => 1 + 1)` never ran); after the fix the page stays responsive and the oracle details section renders both custom adapters with live prices.
- New unit tests in `tests/composables/useEulerOracleAdapters.test.ts` cover: no refetch on a miss for an already-loaded chain, reload on chain change, and no subscription leak into a calling effect.
- `eslint` and `nuxt typecheck` pass.

## Notes

Same latent pattern exists elsewhere but is out of scope here: the per-call `toReactive(computed(...))` factories in `useEulerLabels.ts` (`useEulerProductOfVault` etc.) subscribe callers to `labelsData` at construction the same way, but their target rarely changes and no feedback cycle exists today.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary reloads of oracle adapter data when the requested chain has already been loaded.
  * Kept loaded adapter data stable to avoid replacing the current values on repeat requests.
  * Improved shared reactive behavior so adapter updates are reflected consistently across the app without extra recalculations.

* **Tests**
  * Added coverage for chain-based loading, cached reuse, chain switching, and reactive update behavior.

### DIFF
--- a/composables/useEulerOracleAdapters.ts
+++ b/composables/useEulerOracleAdapters.ts
@@ -37,6 +37,13 @@ const normalizeOracleAdapterMap = (
 const loadAllOracleAdapters = async (chainId: number): Promise<void> => {
   if (!Number.isInteger(chainId) || chainId <= 0) return
 
+  // The dataset is fetched whole per chain, so once it's loaded a lookup miss
+  // means the adapter genuinely isn't in it — reloading cannot surface it and
+  // would replace oracleAdaptersRef with a fresh object, re-triggering every
+  // subscriber (see the loadOracleAdapter miss path, which lands here on each
+  // call for an unlisted adapter).
+  if (oracleAdaptersChainId.value === chainId) return
+
   const inflight = pendingOracleAdapterLoads.get(chainId)
   if (inflight) {
     await inflight
@@ -78,8 +85,17 @@ const loadOracleAdapters = async (chainId: number, addresses?: string[]) => {
   await Promise.all(addresses.map(addr => loadOracleAdapter(chainId, addr)))
 }
 
+// Module-level singleton: toReactive() wraps the computed in reactive(), whose
+// isReadonly() probe reads a property through the proxy — a reactive read of
+// the computed at construction time. Built per-call inside useEulerOracleAdapters,
+// that read would subscribe whatever effect is currently running (e.g. a computed
+// that reaches useEulerLabels() mid-evaluation) to oracleAdaptersRef, coupling
+// unrelated reactive state to oracle-adapter loads. Constructing it once at
+// module init (no active effect) keeps the subscription surface to actual readers.
+const oracleAdapters = toReactive(computed(() => oracleAdaptersRef.value))
+
 export const useEulerOracleAdapters = () => ({
-  oracleAdapters: toReactive(computed(() => oracleAdaptersRef.value)),
+  oracleAdapters,
   loadOracleAdapter,
   loadOracleAdapters,
   loadAllOracleAdapters,
--- a/tests/composables/useEulerOracleAdapters.test.ts
+++ b/tests/composables/useEulerOracleAdapters.test.ts
@@ -1,0 +1,81 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { computed } from 'vue'
+
+const { fetchOracleAdapterMap } = vi.hoisted(() => ({
+  fetchOracleAdapterMap: vi.fn(),
+}))
+
+vi.mock('~/composables/useEulerSdk', () => ({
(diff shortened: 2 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `27a51f4df461fd69d92f1dca28c1d082f5b4a89c`.

What changed:
- `useEulerOracleAdapters` now avoids refetching the whole adapter map after the requested chain is already loaded, so known-missing adapters do not repeatedly replace the shared adapter store.
- The reactive `oracleAdapters` view is now a module-level singleton, avoiding accidental subscription of unrelated computed effects when `useEulerOracleAdapters()` is called inside other reactive pipelines.
- New focused tests cover per-chain caching, chain-change reloads, and the bystander-computed subscription regression.

Validation:
- GitHub checks visible on this head: `lint`, `typecheck`, and `test` are passing.
- `npm run test:run -- tests/composables/useEulerOracleAdapters.test.ts` — passed, 3 tests.
- `npm run build` — passed.
- Browser visual smoke, local production build: `/explore/sentora-ondo`, desktop `1440x1000` and mobile `390x844`; selected Matrix → Oracles and confirmed the Sentora Ondo matrix rendered without app error/freeze symptoms. Only observed console noise was Reown remote-config fetch fallback, not specific to this PR.

Screenshots:
- Desktop Oracles matrix: https://gist.githubusercontent.com/LeonardEulerXYZ/f47fda40c1f137969b8d7d9ee1b85704/raw/593677471f3560064591eb6576139463854ef609/euler-xyz-euler-lite-pr698-pr698-desktop-sentora-ondo-oracles.png
- Mobile Oracles matrix: https://gist.githubusercontent.com/LeonardEulerXYZ/f47fda40c1f137969b8d7d9ee1b85704/raw/71c94402ced721034f600335eb003a7bdd6ebce1/euler-xyz-euler-lite-pr698-pr698-mobile-sentora-ondo-oracles.png

- I checked the sibling consumers of this adapter state: discovery matrix, vault overview oracle-adapter block, `useEulerLabels`, and the shared adapter-view utility. The fix is in the shared composable rather than an inline matrix-only guard, so both matrix and overview consumers inherit the corrected behavior.
- The new tests exercise the reusable composable boundary rather than only the visible Explore route. That is the right abstraction point for this regression.
- No missed sibling surface or duplicated one-off helper stood out in the current diff.

Bot feedback:
- CodeRabbit reported no actionable comments for this head; I did not find a contradictory material issue.

Smoke coverage: browser visual smoke + mobile smoke for the changed Explore matrix/Oracles surface; no wallet/signing coverage, not applicable to this read-only change.

kanvgupta (approved):

PR #705, opened 2026-07-08
add HelpScout Beacon integration

girts512:
Embed the official Beacon loader snippet as a nonce-carrying inline script (bodyClose) in nuxt.config.ts and init it with our Beacon ID. Allow the required HelpScout origins in the CSP (connect/style/font/ frame/media-src) per their documented requirements; object-src stays 'none'. Add a regression test locking the connect-src entries.

## Summary by CodeRabbit

- **New Features**
  - Added an in-app Support panel with searchable HelpScout docs and ticket sending with attachments.
  - Enabled HelpScout Beacon with deferred loading, CSP nonce support, and an option to hide the Beacon.
  - Added support diagnostics collection that captures recent console output and errors.

- **Security**
  - Updated CSP to allow required HelpScout/Beacon resources and connections.
  - Redacted sensitive query-string values in captured diagnostics.

- **Documentation**
  - Added documentation for enabling and operating the Support panel.

- **Tests**
  - Extended security and HelpScout support coverage, plus console-capture behavior tests.


utils/console-capture.ts
```diff
@@ -0,0 +1,86 @@
...
+  catch {
+    return String(arg)
+  }
+}
+
+const push = (level: string, args: unknown[]) => {
+  const text = args
+    .map(stringifyArg)
+    .join(' ')
+    .replace(QUERY_VALUE_RE, '$1[redacted]')
```
LeonardEulerXYZ, line 41: 🔴 **Blocking — global console capture can forward credentials and PII.** This regex only redacts query-string values, but the plugin serializes every argument from every console level and `app.vue` later attaches the buffer to HelpScout. A local probe retained a path-based RPC key (`.../v2/<key>`), a bearer token, and structured `{ token, email }` values unchanged. Existing app/dependency errors can contain request URLs, bodies, signatures, or user data, so truncation does not make this boundary safe. Please replace arbitrary console collection with an allowlisted diagnostic payload, or centralize a structured recursive sanitizer that covers URLs, auth headers/tokens, secret-like keys, calldata/signatures, and nested error shapes; add realistic regression cases. `utils/client-observability.ts` already contains a safer allowlisted/redaction pattern that should not be duplicated more weakly here.

LeonardEulerXYZ (changes requested):

Reviewed head `a75a06f71980e72af43bacbe271346a03b922b04`.

### Walkthrough and verdict

This adds the HelpScout loader/CSP surface, hides the launcher on onboarding, enriches support conversations with wallet/app state, and installs an app-wide console ring buffer. The Beacon itself renders and opens correctly on the current Railway preview, but the diagnostic capture introduces a blocking disclosure path: arbitrary console values can be forwarded to HelpScout with only query-string redaction.

### Blocking finding

- `utils/console-capture.ts:23-42` — the global capture serializes every `log/info/warn/error/debug` argument, while `QUERY_VALUE_RE` masks only URL query values. Common path-based RPC credentials, bearer tokens, and structured fields such as `{ token, email }` survive unchanged and are attached on Beacon open from `app.vue:159-171`. A local regression probe against this exact implementation retained all four examples. See the inline comment for the required boundary.

The support integration is app-global, so its privacy boundary must cover every route and dependency logger rather than one known URL shape. The patch also creates a second, substantially weaker redaction implementation beside `utils/client-observability.ts:65-74`. Prefer one shared, tested sanitizer—or, better, an allowlisted diagnostic payload—rather than growing parallel regex sets. The hard-coded Beacon ID noted by CodeRabbit is a valid deployment-hygiene suggestion, but non-blocking here; CodeRabbit's review was attached to an earlier commit and did not cover the later console-capture surface.

### Validation

- `git diff --check` — passed
- Focused Vitest: `tests/server/security.test.ts`, `tests/utils/console-capture.test.ts`, plus a removed local-only credential-retention probe — 28/28 passed
- `npm run build` — passed (existing bundle/sourcemap warnings only)
- Headed Chromium/Xvfb against the Railway preview:
  - desktop 1440×1000: real Explore page, Beacon opens, diagnostics notice visible
  - mobile 390×844: Beacon opens and form fits; it intentionally overlays the app/menu while open
  - no page errors; one non-blocking third-party 403 console resource error

Smoke coverage: browser visual smoke + mobile smoke; disconnected only. Wallet connection/signing was not exercised.

### Screenshot evidence

Desktop — HelpScout form open on Explore (1440×1000):

Desktop HelpScout Beacon open

Mobile — HelpScout form open on Explore (390×844):

Mobile HelpScout Beacon open

PR #708, opened 2026-07-08
fix: restore migration batch prewarm fast path + clarify Aave E-Mode revert

dglowinski:
## Summary

Two migration fixes.

### 1. Regression: adding a migration to the batch re-simulates instead of using the prewarmed plan

The background prewarm still builds the full cross-protocol migration simulation as soon as targets settle, but since `298fe8a5` ("fix: preserve reviewed migration batches") the batch entry's `buildPlan` re-ran that entire simulation again at click time (position resolve → plan build), discarding the warmed plan. That restored the latency the prewarm (`abe4402a`) was meant to remove.

`298fe8a5` did this to simulate against the batch's *planning account* rather than the owner. But for the **first** batch entry the planning account is the fresh owner account the preview was already warmed against (`usePlanAccount` returns the fresh account when no batch layer is active; the batch's empty-case planning account is the same fresh owner snapshot). So we can reuse the prewarmed plan there and only re-simulate when the batch already has entries — whose planning account is a later layer the prewarm didn't cover.

Applied to both:
- Inbound (external → Euler): `pages/position/[number]/borrow/swap.vue`
- Outbound (Euler → Aave): `pages/position/[number]/migrate.vue`

### 2. Clearer Aave E-Mode revert message

Migrating a position into an Aave account that already sits in an incompatible **E-Mode** reverts with Aave V3's `NotBorrowableInEMode()` (`0x57db5bba`), which surfaced as the raw error name in the batch simulation. Mapped it through the existing error-copy pattern (`ERROR_SIGNATURE_MAP` + `ERROR_MESSAGE_MAP`, mirroring `HealthFactorLowerThanLiquidationThreshold`) to explain the cause.

Repro that motivated this: position 3 on Base, owner in Aave E-Mode 4 ("LBTC_cbBTC") — supplying cbBTC / borrowing USDC is not permitted in that E-Mode, so the borrow reverts.

## Testing
- `npm run typecheck` — passes
- `npm run lint` (changed files) — passes
- Selector `0x57db5bba` confirmed on Base as `NotBorrowableInEMode()` by reproducing the borrow revert on-chain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for a new borrow rejection reason so users now see a clearer message when an asset can’t be borrowed in E-Mode.
  * Improved batch migration and borrow swap handling to reuse existing preview results when possible, reducing unnecessary reprocessing and making the flow smoother.

### DIFF
--- a/entities/constants.ts
+++ b/entities/constants.ts
@@ -18,6 +18,7 @@ export const ERROR_MESSAGE_MAP: Record<string, string> = {
   ERC4626ExceededMaxRedeem: 'Redeem exceeds vault limits.',
   SlippageExceeded: 'Slippage exceeded your tolerance. Increase slippage tolerance or refresh the quote.',
   HealthFactorLowerThanLiquidationThreshold: 'Aave account health factor would be below the liquidation threshold.',
+  NotBorrowableInEMode: 'This Aave account is in an E-Mode that is incompatible with this position\'s assets. Exit the E-Mode in the Aave app before migrating.',
   Swapper_SwapError: 'Swap failed. Try increasing slippage tolerance, refreshing the quote, or selecting a different swap provider.',
   Swapper_UnknownHandler: 'Swap provider is not registered. Try selecting a different swap provider.',
   SwapVerifier_skimMin: 'Swap received less than the minimum amount. Increase slippage tolerance or refresh the quote.',
@@ -36,6 +37,7 @@ export const ERROR_MESSAGE_MAP: Record<string, string> = {
 export const ERROR_SIGNATURE_MAP: Record<string, string> = {
   ...EVC_ERROR_SIGNATURES,
   '0x6679996d': 'HealthFactorLowerThanLiquidationThreshold',
+  '0x57db5bba': 'NotBorrowableInEMode', // Aave V3 borrow rejected: asset not borrowable in the account's active E-Mode
 }
 
 export const TTL_INFINITY = BigInt(
--- a/pages/position/[number]/borrow/swap.vue
+++ b/pages/position/[number]/borrow/swap.vue
@@ -2874,6 +2874,17 @@ const addInboundExternalMigrationToBatch = async () => {
     await addBatchEntry({
       ...batchEntry,
       buildPlan: async (account: Account<IHasVaultAddress>) => {
+        // The preview was warmed against the fresh owner account. For the first
+        // batch entry the planning account is that same account, so reuse the
+        // prewarmed plan instead of re-running the full cross-protocol migration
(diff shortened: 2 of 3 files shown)


LeonardEulerXYZ (commented):
Reviewed current PR head `b2b85b34a59d14311f835c27183d9f85e8b4a224`.

Scope reviewed:
- `pages/position/[number]/migrate.vue`: outgoing migration preview now resolves the external migration position once, shares plugin prefetch data, and prepares the simulation/preview plans from the same batch output.
- `pages/position/[number]/borrow/swap.vue`: inbound external migration preview now shares the prefetch payload between Tenderly plan preparation and calldata preview.
- `entities/constants.ts` / `utils/tx-errors.ts`: Aave V3 selectors are mapped to user-facing messages, with decoded-chain friendly messages still preferred before raw fallback output.

Validation performed:
- `npm run test:run -- tests/utils/tx-errors.test.ts tests/composables/useTxBatch.test.ts` — passed, 47 tests.
- `npx eslint entities/constants.ts 'pages/position/[number]/borrow/swap.vue' 'pages/position/[number]/migrate.vue'` — passed; only a Node experimental warning from ESLint module loading.
- `npm run build` — passed after materializing a fresh local dependency tree; initial hardlinked dependency cache was stale and not treated as PR signal.
- Browser smoke: local production build, headed Chromium under Xvfb, desktop and mobile `/position/0/migrate?network=1` route load. The route rendered the expected disconnected shell without blank/loading state. No screenshots posted because the PR has no stable visual delta to evidence without a connected migration fixture; this was route-load/mobile sanity, not wallet/signing coverage.

- The migration fast path is kept within the existing inbound/outgoing migration preview functions rather than introducing a new one-off abstraction. I checked the sibling inbound and outgoing migration flows for the same prefetch/prepare pattern; both now reuse the resolved migration/batch data for the preview path being changed.
- The Aave error copy uses the existing shared `ERROR_SIGNATURE_MAP` / `ERROR_MESSAGE_MAP` path and the existing decoded-chain formatter, rather than adding route-local formatting. That is the right reuse point for sibling simulation/error surfaces.
- Focused tests cover the Aave health-factor selector path. I did not find a blocking consistency issue, though a future small test for `NotBorrowableInEMode` would be a tidy companion if this selector is expected to regress often.

Bot/reviewer comments:
- I found no prior Leonard comments/reviews to clean up, and no active bot finding that needed promotion into a Leonard finding.

Smoke coverage: focused tests + typecheck/lint + production build + narrow desktop/mobile route-load smoke. No wallet/signing smoke.

Seranged (approved):

PR #712, opened 2026-07-09
fix: denominate swap-supply estimates in quoted collateral output

Seranged:
## Problem

On `/position/<n>/supply`, when the user pays with a token that differs from the collateral vault's asset, the amount input is denominated in the pay-with token. The position summary estimates, however, parsed that same string with the collateral vault's asset decimals:

- `amountFixed` fed the sync estimates (LTV, health score, liquidation price/buffer) treating the raw input as collateral units.
- `updateAsyncEstimates` did the same for the projected Net APY and collateral USD value.

Entering 100 USDC to swap into a WETH collateral vault made the summary project a deposit of **100 WETH** instead of the quoted ~0.03 WETH, so all after-values (Net APY, LTV, liquidation price/buffer, health score) were wildly wrong.

## Fix

- Added `swapCollateralDeltaNano`: in swap-supply mode the collateral delta is the effective quote's `amountOut` (already denominated in the collateral asset), and 0 while quotes are loading. `null` in every other flow (direct supply, native wrap, all withdraw modes), which keeps the existing collateral-decimals parsing there.
- `amountFixed` and the async estimate deltas now use it.
- Estimates recompute when the effective quote changes (quotes arriving after the debounced fetch, provider selection, refresh), since the delta now derives from the quote rather than only the typed amount.
- The amount watcher resets quote state before computing estimates so a stale quote's output can't leak into the new amount's estimates.
- The swap-quote zero-check now parses the amount with the pay-with token's decimals instead of the collateral's.

## Testing

- New regression tests in `tests/composables/useCollateralForm.test.ts` covering: swap-supply delta from quoted output, delta reset when quotes clear, direct supply unchanged, withdraw-with-swap unchanged. The two swap-supply cases fail on the previous implementation.
- Full vitest suite passes (1196 tests), `nuxt typecheck` and eslint clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved collateral amount handling for swap-based supply flows so displayed and calculated values now follow the quoted output amount instead of the raw input.
  - Quote updates now refresh estimates more reliably when pricing or provider selection changes, and reset cleanly when quotes are cleared.
  - Direct supply and withdraw flows continue to use the expected collateral-based amount formatting.

- **Tests**
  - Added coverage for swap-supply, direct supply, quote reset, and withdraw-with-swap behavior.

### DIFF
--- a/composables/position/useCollateralForm.ts
+++ b/composables/position/useCollateralForm.ts
@@ -262,8 +262,20 @@ export const useCollateralForm = (options: UseCollateralFormOptions) => {
   })
 
   // --- FixedPoint computeds ---
+  // In swap-supply mode `amount` is denominated in the user-selected "pay
+  // with" token, so parsing it with collateral decimals would treat e.g.
+  // "100" (USDC) as 100 WETH. The collateral delta is the quoted swap output
+  // instead — 0 while no quote is available, so estimates show no change
+  // until quotes arrive. `null` means "amount is collateral-denominated"
+  // (direct supply, native wrap, all withdraw flows).
+  const swapCollateralDeltaNano = computed<bigint | null>(() => {
+    if (options.mode !== 'supply' || !options.needsSwap.value) return null
+    if (!swapEffectiveQuote.value) return 0n
+    const amountOut = BigInt(swapEffectiveQuote.value.amountOut || 0)
+    return amountOut > 0n ? amountOut : 0n
+  })
   const amountFixed = computed(() => FixedPoint.fromValue(
-    valueToNano(amount.value || '0', collateralVault.value?.asset.decimals),
+    swapCollateralDeltaNano.value ?? valueToNano(amount.value || '0', collateralVault.value?.asset.decimals),
     Number(collateralVault.value?.asset.decimals),
   ))
   const borrowedFixed = computed(() => FixedPoint.fromValue(position.value?.borrowed || 0n, borrowVault.value?.shares.decimals || 18))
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
Reviewed current head `720e66aa3df65743eeb77f8e9cdc27b971d1b063`.

Verdict: no blocking findings from code review or local validation. I’m leaving this as a comment rather than approval, per the review policy for this repo.

What changed:
- `useCollateralForm` now treats swap-supply estimate deltas as collateral-denominated quote output (`amountOut`) instead of parsing the user’s pay-with amount as collateral.
- Quote input parsing now uses the effective pay-with asset decimals.
- The amount watcher now clears stale quotes before recomputing estimates, and a quote-delta watcher recomputes estimates when a quote arrives or provider changes.
- Focused tests cover swap-supply quote output denomination, quote clearing, direct supply collateral decimals, and withdraw-with-swap remaining collateral-denominated.

Validation performed:
- `npm run test:run -- tests/composables/useCollateralForm.test.ts` — passed, 4/4 tests
- `npm run build` — passed

- I checked the sibling lend supply implementation, which already uses quote `amountOut` for swap-supply APY projections; this PR brings position supply onto the same invariant rather than adding an inconsistent UI-only workaround.
- Withdraw flows remain intentionally collateral-denominated and are covered by the new regression test.
- No missed sibling surface or reusable-helper extraction issue found for this narrow change.

Bot feedback:
- CodeRabbit reported no actionable comments; I did not find a material bot finding to amplify or refute.

Browser smoke:
- Coverage: incomplete browser visual smoke.
- The PR preview route for position supply loaded with HTTP 200 on desktop and mobile, and the onboarding/migration modal could be dismissed with no material page errors captured.
- After dismissal, the route remained on the supply-collateral shell/spinner and did not expose the actual form fields (`Pay with`, `Review Supply`, summary metrics), so I did not treat screenshots as useful review evidence or post them here. This looks like preview/runtime route readiness noise rather than a diff-specific failure, but it means I did not get full user-flow/form interaction coverage.

kanvgupta (approved):

PR #590, opened 2026-06-19
fix: share market count grammar

Seranged:
Summary
- Fix the remaining singular/plural market count copy in the Securitize vault overview.
- Move market availability copy into a shared formatter used by both vault overview components.
- Add a focused unit test for unavailable, singular, and plural market counts.

Tests
- npx eslint components/entities/vault/overview/VaultOverviewBlockGeneral.vue components/entities/vault/overview/SecuritizeVaultOverview.vue utils/vault-display.ts tests/utils/vault-display.test.ts
- npx vitest run tests/utils/vault-display.test.ts

Follow-up to #587 / #188.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated market availability formatting logic into a shared utility function across vault components.

* **Tests**
  * Added test coverage for market availability formatting functionality.

PR #713, opened 2026-07-09
fix: prevent infinite balance refetch loop for non-checksummed wallet addresses

Seranged:
## Problem

Connecting a wallet that reports its address in a non-EIP-55-checksummed form (WalletConnect wallets commonly report lowercase; injected wallets checksum) hangs the browser tab shortly after the session is established, ending in Chrome's "Aw, Snap! RESULT_CODE_HUNG".

Root cause is in `composables/useWallets.ts`: after every balance fetch, the `finally` block schedules a follow-up run if `needsFetch()` is still true. `needsFetch()` compared the stored last-fetched address (always checksummed via `getAddress()`) against the raw connector address with a cased string compare. For a lowercase connector address the two strings never match, so the follow-up refired unconditionally — an unbounded refetch loop (~25,000 iterations observed before the tab died), each iteration paying a full SDK instance lookup and vault-registry iteration on the main thread.

The loop is driven from app-root watchers, so it triggers on any page as soon as the wallet connects and vaults load. Reproduced end-to-end with a scripted WalletConnect wallet peer: the same address connects cleanly when checksummed and hangs the tab when lowercase.

## Fix

- Compare address identity case-insensitively in `needsFetch()` (`isSameAddress` helper).
- Cap consecutive follow-up refetches (`MAX_CONSECUTIVE_AUTO_REFETCHES = 3`) as a safety valve: any future never-settling condition logs a warning instead of hanging the tab. The counter resets when the state settles and in `resetBalances()`.

## Verification

- New regression tests in `tests/composables/useWallets.test.ts`; mutation-checked (all fail against the old comparison). Full unit suite passes.
- Live check against a production build with a scripted WalletConnect session using a lowercase address: connects and stays responsive for the full observation window (previously hung within ~3 seconds).
- Docs updated (`docs/portfolio-logic.md`) with the refetch-settling safeguards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed wallet balance auto-refresh getting stuck when connector-reported address casing differs from stored values.
  * Added a cap to prevent endless follow-up balance refresh attempts, improving stability if repeated retries occur.

* **Documentation**
  * Documented the new “balance refetch settling” behavior and the retry limit to clarify when additional refreshes may happen.

* **Tests**
  * Added regression tests covering address casing changes and capped auto-refetch behavior (including warning logging).

### DIFF
--- a/composables/useWallets.ts
+++ b/composables/useWallets.ts
@@ -29,6 +29,21 @@ const lastFetchChainId = ref<number | null>(null)
 const lastFetchAddress = ref<string | null>(null)
 let fetchPromise: Promise<void> | null = null
 
+// Connector-reported addresses are not guaranteed to be EIP-55 checksummed
+// (WalletConnect wallets commonly report lowercase), while lastFetchAddress
+// stores the checksummed form. Address identity must therefore be compared
+// case-insensitively — a cased string compare here kept needsFetch() true
+// forever and refetched balances in an unbounded loop, hanging the tab.
+const isSameAddress = (a?: string | null, b?: string | null) =>
+  (a ?? '').toLowerCase() === (b ?? '').toLowerCase()
+
+// Safety valve for the follow-up refetch in updateBalances' finally block:
+// legitimate follow-ups (chain/address changed mid-fetch) settle within a
+// couple of rounds; anything past the cap indicates a needsFetch() invariant
+// that can never be satisfied.
+const MAX_CONSECUTIVE_AUTO_REFETCHES = 3
+let consecutiveAutoRefetches = 0
+
 // Reference-counted flag: when >0, updateBalances includes the full token list
 // (all Uniswap/DefiLlama entries — thousands of tokens on mainnet) so pages
 // with the "pay with" token selector can show non-zero balances first.
@@ -188,9 +203,22 @@ export const useWallets = () => {
     finally {
       isFetching.value = false
       fetchPromise = null
-      // If dependencies changed while we were fetching, schedule a follow-up run
+      // If dependencies changed while we were fetching, schedule a follow-up run.
+      // Capped: if needsFetch() can never settle (an invariant bug), an
+      // uncapped follow-up chain refetches forever and hangs the tab.
       if (needsFetch()) {
-        fetchPromise = updateBalances()
+        if (consecutiveAutoRefetches < MAX_CONSECUTIVE_AUTO_REFETCHES) {
+          consecutiveAutoRefetches++
+          fetchPromise = updateBalances()
+        }
+        else {
+          logWarn('wallets/fetchBalances', 'auto-refetch cap reached — needsFetch() never settled', {
+            data: { chainId: currentChainId, address: balanceAddress.value },
+          })
+        }
+      }
+      else {
+        consecutiveAutoRefetches = 0
       }
     }
   }
@@ -200,7 +228,7 @@ export const useWallets = () => {
     return (isConnected.value || isSpyMode.value)
       && loadedChainId.value === chainId.value
       && !!balanceAddress.value
-      && (lastFetchChainId.value !== chainId.value || !isLoaded.value || lastFetchAddress.value !== balanceAddress.value)
+      && (lastFetchChainId.value !== chainId.value || !isLoaded.value || !isSameAddress(lastFetchAddress.value, balanceAddress.value))
       && !isFetching.value
   }
 
@@ -227,6 +255,7 @@ export const useWallets = () => {
     lastFetchChainId.value = null
     lastFetchAddress.value = null
     fetchPromise = null
+    consecutiveAutoRefetches = 0
   }
 
   const getBalance = (tokenAddress: Address): bigint => {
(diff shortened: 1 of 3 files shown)


LeonardEulerXYZ (commented):
## Leonard review — PR #713

Verdict: COMMENT — no blocking findings found for the current head `3d49a95ae300626fca03bbba869356644986c499`.

### Walkthrough

This PR fixes the wallet-balance refetch loop in `useWallets()` by:

- comparing the last fetched wallet address and the current connector/effective address case-insensitively;
- adding a capped follow-up-refetch safety valve so a future non-settling `needsFetch()` invariant logs and stops instead of spinning indefinitely;
- adding focused regression coverage for lowercase/checksummed connector addresses; and
- documenting the balance-refetch settling invariant.

### Risk assessment

The changed runtime surface is intentionally narrow: app-root wallet balance fetching after a connected wallet/effective address is available. The main production risk was stale or repeated wallet-balance refreshes; the new comparison uses address identity rather than display casing, and the cap is a reasonable defensive backstop for a UI-freezing loop.

I did not find evidence of new signing, transaction-construction, external-network, dependency, workflow, or secret-handling behavior in the diff.

### Validation performed

- `npm run test:run -- tests/composables/useWallets.test.ts` — passed, 3 tests.
- `npm run build` — passed; only existing-style Vite/Rollup chunk/module warnings observed.
- Light browser smoke from the production build:
  - desktop `/portfolio?network=1` loaded real app content with portfolio/connect-wallet text;
  - mobile `/portfolio?network=1` loaded real app content with portfolio/connect-wallet text.

Smoke coverage: route/browser smoke only. I did not exercise a real WalletConnect session or wallet signing flow in this review; the changed loop is covered by the focused composable tests and the PR author's described WalletConnect repro/verification.

Screenshots: not posted — this is not a visual UI change, and the browser smoke was a route-health check rather than visual evidence.

### Scalability / maintainability hygiene pass

I searched sibling address/effective-address flows and checked the relevant balance-fetching surfaces. The bug is specific to the singleton `useWallets()` settled-fetch state: `lastFetchAddress` is stored checksummed while `effectiveAddress` can retain connector casing. The fix is placed at that comparison boundary and backed by a focused composable test. Similar sibling flows generally normalize with `getAddress()` or lower-case map keys before identity comparisons, so I do not see a missed parallel surface requiring the same fix.

The local `isSameAddress()` helper is acceptable here because the scope is a private invariant inside `useWallets()`, not a repeated UI formatter/business rule. If more address-identity comparisons appear across composables later, extracting a shared `sameAddress()` utility plus unit tests would be the cleaner PR #590-style direction.

### Bot/reviewer feedback

CodeRabbit had only an in-progress autogenerated status comment when checked; there were no material bot findings to verify or refute.

### Notes

No Leonard-owned prior comments/reviews/inline comments existed for this PR, so there was no cleanup to perform.

LeonardEulerXYZ (commented):
## Leonard review — PR #713

Verdict: COMMENT — no blocking findings found for current head `26aea285677765a4a2b443348c9053fe497c9bf4`.

### Walkthrough

This re-review covers the new head after the original Leonard review. The PR now:

- compares `lastFetchAddress` and the active wallet/effective address by address identity rather than connector casing;
- caps consecutive automatic follow-up balance refetches so a future never-settling `needsFetch()` condition warns and stops instead of hanging the tab;
- adds direct regression coverage for lowercase/checksummed WalletConnect-style addresses and for the auto-refetch cap; and
- documents the balance-refetch settling invariant.

### Risk assessment

The changed runtime surface remains narrow: app-root wallet balance fetching for connected/spy effective addresses after vault registry readiness. I did not find new signing, transaction-building, dependency, workflow, external proxy, or credential-handling behavior in the diff.

The important failure mode was a main-thread refetch loop. The case-insensitive comparison addresses the known WalletConnect casing trigger, and the cap is a sensible safety valve for future invariant drift.

### Validation performed

- `npm run test:run -- tests/composables/useWallets.test.ts` — passed, 4 tests.
- `npm run build` — passed; only existing-style Vite/Rollup chunk/module warnings observed.
- Light production-build browser smoke with headed Chromium:
  - desktop `/portfolio?network=1` loaded real portfolio/connect-wallet content;
  - mobile `/portfolio?network=1` loaded real portfolio/connect-wallet content.

Smoke coverage: route/browser smoke plus mobile route smoke. I did not exercise a real WalletConnect session or wallet signing flow in this re-review; the loop invariant is covered by focused composable tests, including the new cap test.

Screenshots: not posted — this PR is not a visual UI change, and the browser pass was route-health coverage rather than visual evidence.

### Scalability / maintainability hygiene pass

I searched sibling effective-address/address-comparison surfaces and balance consumers. The bug is specific to `useWallets()`' singleton settled-fetch state: a checksummed `lastFetchAddress` was compared against connector-provided casing. The fix sits at that state boundary, and the tests now cover both the original casing loop and CodeRabbit's requested cap behavior.

No missed sibling route/flow stood out: other address comparisons I checked either normalize both sides with `getAddress()` or compare lower-cased values. The local `isSameAddress()` helper is acceptable for this private `useWallets()` invariant. If this pattern starts repeating across composables, a shared address-identity helper with unit tests would be the cleaner PR #590-style extraction.

### Bot/reviewer feedback

CodeRabbit's prior material note asked for a direct test of the auto-refetch cap. The current head adds `caps follow-up refetches when needsFetch() can never settle`, and the focused test suite passed, so I consider that suggestion resolved.

### Comment lifecycle

No Leonard-owned top-level issue comment or inline review comments were present to edit/delete. The previous Leonard formal review is on the older head `3d49a95ae300626fca03bbba869356644986c499`; this review supersedes it for the current head.

kasperpawlowski (approved):

PR #717, opened 2026-07-09
fix: expose the connector address in EIP-55 checksummed form

Seranged:
Follow-up to #713. That PR fixed the concrete tab-hang by making one comparison casing-proof; this one removes the bug class at the source so raw-casing comparisons against the connector address become structurally impossible.

## Change

- `useWagmi().address` now returns the EIP-55 checksummed form via `normalizeAddress` (single choke point). Connectors are not guaranteed to report checksummed casing — AppKit's WalletConnect connector returns the wallet's session account string verbatim, and wallets may emit lowercase (spec-legal per CAIP-10 / the eip155 namespace profile). `checksummedAddress` simplifies to an alias. `useEffectiveAddress().effectiveAddress` becomes uniformly checksummed since the spy branch already was.
- Normalization runs once per address change in a computed — a single keccak hash at the boundary, mirroring what upstream wagmi's own WalletConnect connector does (`provider.accounts.map(x => getAddress(x))`); AppKit's replacement connector dropped it.

## Consumer sweep

Audited every consumer of `address` / `checksummedAddress` / `friendlyAddress` / `effectiveAddress` across composables, components, pages, utils, and services for case-sensitive comparisons, address-derived cache/storage keys, and strict compares against wagmi-internal values. All but one already normalize (`getAddress`/`toLowerCase`/`isSameAddress`) or only pass the value to SDK/API calls; no localStorage/useState keys are built from the wallet address.

The one behavioral dependency found: the Keyring credential check compared `cred.trader` (echoed by the Keyring browser extension in its own casing) strictly against the exposed address — now compared case-insensitively so a casing difference can't force a spurious re-auth.

## Verification

- New `tests/composables/useWagmi.test.ts` feeds a lowercase connector address through mocked wagmi and asserts the exposed values are checksummed; mutation-checked (reverting the normalization fails it).
- The #713 regression tests (refetch loop, auto-refetch cap) still pass — the case-insensitive compare and cap remain as defense-in-depth.
- Full unit suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet address handling so addresses are compared and displayed consistently, even when casing differs.
  * Reduced the chance of repeated balance refresh loops caused by address formatting mismatches.

* **Tests**
  * Added coverage for address normalization, preserving existing checksummed addresses, and handling missing addresses.

* **Documentation**
  * Clarified address casing behavior in balance refresh guidance.

### DIFF
--- a/composables/useKeyring/index.ts
+++ b/composables/useKeyring/index.ts
@@ -160,9 +160,11 @@ export const useKeyring = (vaultAddress: string | Ref<string>) => {
 
       const state = await KeyringConnect.getExtensionState()
       const cred = state?.credentialData
+      // The extension echoes the trader address in its own casing — compare
+      // case-insensitively so a casing difference doesn't force a re-auth.
       if (
         cred
-        && cred.trader === userAddress.value
+        && cred.trader?.toLowerCase() === userAddress.value?.toLowerCase()
         && cred.policyId === policyId.value
         && cred.chainId === chainId.value
       ) {
--- a/composables/useWagmi.ts
+++ b/composables/useWagmi.ts
@@ -1,7 +1,8 @@
 import { useAccount, useDisconnect, useSwitchChain, useEnsName, useConfig } from '@wagmi/vue'
 import { connect as connectWallet, getConnectors } from '@wagmi/vue/actions'
-import { getAddress, isAddress, type Address } from 'viem'
+import type { Address } from 'viem'
 import { logWarn } from '~/utils/errorHandling'
+import { normalizeAddress } from '~/utils/normalizeAddress'
 import { truncate } from '~/utils/string-utils'
 import { useAddressScreen } from '~/composables/useAddressScreen'
 import { parseChainId } from '~/entities/chainRegistry'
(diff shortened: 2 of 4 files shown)


LeonardEulerXYZ (commented):
## Leonard review — PR #717

Verdict: Reviewed — no blockers found.

Reviewed head: `edaa88cf35759ce154b216c8cd19528b31bf1d0f`

### What changed
- `useWagmi().address` now normalizes the screened connector address to EIP-55 form at the shared wallet boundary.
- `checksummedAddress` becomes an alias of the normalized address output.
- Keyring’s cached credential check compares `cred.trader` and the wallet address case-insensitively.
- Docs and focused `useWagmi` regression coverage were added for lowercased connector addresses.

### Findings
No blocking findings.

### Validation
- Focused tests passed: `npm run test:run -- tests/composables/useWagmi.test.ts tests/composables/useWallets.test.ts tests/composables/useEffectiveAddress.test.ts`
- Typecheck passed: `npm run typecheck`
- Production build passed: `npm run build`
- Diff hygiene/security sweep: no new storage keys, dependency/script changes, external calls, secrets, or suspicious executable surfaces in the patch.
- Existing CI at this head shows lint/typecheck/test passing; CodeRabbit was still processing and had no substantive current-head findings to verify at review time.

### Scalability / maintainability hygiene pass
This is the right abstraction level for the bug class: the casing fix is placed at the shared `useWagmi().address` choke point rather than duplicated in individual portfolio/action components. I checked the sibling `useWagmi`/`useEffectiveAddress` consumers and the address-sensitive wallet-balance path; the remaining sibling flows either already normalize with `getAddress`, compare lowercase keys intentionally, or only pass the normalized value onward to SDK/API calls. No missed sibling surface or one-off helper smell found.

### Smoke coverage
Browser/mobile smoke not run: this PR changes wallet connector address normalization and Keyring comparison logic, not a visual layout/interaction surface. Focused composable regression tests, typecheck, and build exercise the changed invariant more directly than route-load screenshots would. Wallet/signing coverage was not run.

Screenshots: not applicable — no visual UI change.

kasperpawlowski (approved):

PR #718, opened 2026-07-09
fix(portfolio): show an error state instead of $0.00 when the portfolio fetch fails

Seranged:
## Problem

When `fetchAndUpdatePortfolio` in `useEulerAccount` throws, the catch block records a diagnostic but still calls `markLoaded()`, leaving `visiblePortfolio`/`allPortfolio` `undefined`. The portfolio page only substitutes "—" when positions exist with missing prices, so an outright load failure falls through to formatted zeros:

> $0.00 Supplied · $0.00 Borrowed · $0.00 Net Worth

For an account with real deposits and borrows this is indistinguishable from an empty account, and the recorded `portfolioDiagnostics` are consumed nowhere in the UI — the failure is fully swallowed. In a lending context, showing a confident $0.00 where a user has funds can drive wrong decisions.

## Fix

- Add an explicit `hasPositionsFetchError` flag in `useEulerAccount`, set when a fetch throws and cleared at the start of every fresh attempt.
- Expose `hasPortfolioLoadError`, computed as "a fetch failed **and** there is no portfolio to fall back on". Gating on the absence of data means a transient background-refresh failure keeps the last-good figures on screen, while a genuine cold-load failure lets the UI show an error.
- On the portfolio page, render an error banner with a **Retry** action and show `—`/`-` for the value and performance figures instead of zeros. The `data-value` capture attributes reflect the error too, so parity/e2e captures don't record misleading zeros.

This mirrors the existing `hasQueryFailure` → "Position data is incomplete" handling already used on the per-position page.

## Testing

- Added a unit test in `tests/composables/useEulerAccount.test.ts` covering the failed-fetch path (asserts `hasPortfolioLoadError` is set, totals stay zero, and an error-severity diagnostic is present).
- `npx vitest run tests/composables/useEulerAccount.test.ts` — passes.
- ESLint clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clearer handling for portfolio load failures, so the app can distinguish between first-time load errors and temporary refresh issues.
  * Portfolio stats now show placeholders when data can’t be loaded, instead of displaying misleading values.
  * An error banner with a retry option now appears when the portfolio fails to load, making recovery easier.

### DIFF
--- a/composables/useEulerAccount.ts
+++ b/composables/useEulerAccount.ts
@@ -28,6 +28,10 @@ const isPositionsLoaded = ref(false)
 const isDepositsLoading = ref(true)
 const isDepositsLoaded = ref(false)
 const isShowAllPositions = ref(false)
+// True when the most recent portfolio fetch threw, as opposed to succeeding
+// with partial per-asset pricing issues. Surfaced to the UI so a failed load
+// renders an error state instead of formatted zeros — see `hasPortfolioLoadError`.
+const hasPositionsFetchError = ref(false)
 
 // Transparent layer overlay: when a non-zero batch layer is active, the
 // simulated portfolio is served for both the visible and all-positions views,
@@ -101,6 +105,7 @@ export const useEulerAccount = () => {
     visiblePortfolio.value = undefined
     allPortfolio.value = undefined
     portfolioDiagnostics.value = []
+    hasPositionsFetchError.value = false
     isPositionsLoaded.value = false
     isPositionsLoading.value = true
     isDepositsLoaded.value = false
@@ -121,6 +126,7 @@ export const useEulerAccount = () => {
     const gen = positionGuard.current()
 
     try {
+      hasPositionsFetchError.value = false
       if (!walletAddress) {
         visiblePortfolio.value = undefined
         allPortfolio.value = undefined
@@ -162,6 +168,7 @@ export const useEulerAccount = () => {
     catch (error) {
       if (positionGuard.isStale(gen)) return
       logWarn('useEulerAccount/fetchAndUpdatePortfolio', error)
+      hasPositionsFetchError.value = true
       portfolioDiagnostics.value = [{
         code: 'SOURCE_UNAVAILABLE',
         severity: 'error',
@@ -227,6 +234,12 @@ export const useEulerAccount = () => {
   startWatchers()
   onScopeDispose(releaseWatchers)
 
+  // A portfolio fetch failed AND we have no (stale) portfolio to fall back on.
+  // Gating on the absence of data means a transient background-refresh failure
+  // keeps the last-good figures on screen, while a genuine cold-load failure
+  // lets the UI show an error instead of misleading $0.00 totals.
+  const hasPortfolioLoadError = computed(() => hasPositionsFetchError.value && !portfolio.value)
+
   const portfolioRoe = computed(() => portfolio.value?.roe ?? 0)
   const portfolioNetApy = computed(() => portfolio.value?.netApy ?? 0)
   const totalSuppliedValue = computed(() => usdWadToNumber(portfolio.value?.totalSuppliedValueUsd))
@@ -276,6 +289,7 @@ export const useEulerAccount = () => {
   return {
     portfolio,
     portfolioDiagnostics,
+    hasPortfolioLoadError,
     borrowPositions,
     depositPositions,
     removedBorrowPositions,
--- a/composables/useEulerAccount.ts
+++ b/composables/useEulerAccount.ts
@@ -126,11 +126,11 @@ export const useEulerAccount = () => {
     const gen = positionGuard.current()
 
     try {
-      hasPositionsFetchError.value = false
       if (!walletAddress) {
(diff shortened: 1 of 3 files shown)


LeonardEulerXYZ (changes requested):

Reviewed head `6a970c665e54f247a0cd1d0ac96dc019ebbe00b5`.

This is the right direction: a cold portfolio fetch failure should not be rendered as a confident empty account. The new composable flag and portfolio-page placeholders cover the initial thrown-fetch path, and the focused test exercises that path.

One blocking edge remains, matching CodeRabbit's major comment after verification:

- `composables/useEulerAccount.ts`: `hasPositionsFetchError` is cleared synchronously at the start of each new fetch attempt. After a cold-load failure, clicking Retry or waiting for a refresh immediately drops `hasPortfolioLoadError` while `portfolio` is still `undefined`, so the portfolio cards can show `0.00%` / `$0` again for the duration of the retry. That reintroduces the exact misleading-zero state this PR is trying to remove, just during the retry window. Keep the error/placeholder state until a new portfolio result is actually confirmed, then clear the error on the success path.

Validation performed:
- `npx vitest run tests/composables/useEulerAccount.test.ts` — passed.
- `npm run build` — passed, with existing Vite/Rollup chunk/dynamic-import warnings only.
- Local throwaway regression test for “failed cold load → pending retry” — failed as expected on this head: `hasPortfolioLoadError` became `false` while `portfolio` was still `undefined`.

- Scope is appropriately centered in `useEulerAccount`, rather than a page-only formatting patch. That is the right reusable layer for portfolio consumers.
- I checked sibling portfolio/error surfaces: position pages already have separate `hasQueryFailure` handling; this PR’s shared portfolio-load state is specifically for account portfolio data and does not need to be copied into position-card components.
- The remaining issue is also best fixed in the shared composable, not with more page-level guards.

Bot feedback:
- Verified CodeRabbit’s major retry-window finding as valid and blocking.
- The placeholder-character nit is minor and not merge-blocking.

Browser/mobile smoke:
- Incomplete for the isolated portfolio-fetch failure. A local headed/Xvfb attempt with forced RPC failures exercised a broader prerequisite failure mode rather than just `fetchPortfolio`, so I did not use or post those screenshots as evidence. The blocking retry invariant is covered more directly by the composable-level regression test above.

LeonardEulerXYZ (commented):
## Leonard review — current head `95a23fc2f70ab476c08c776d44a0f371284b9c35`

Verdict: no blocking findings. Submitting as a comment review, not approval.

The previous blocker on head `6a970c66` is resolved: `hasPositionsFetchError` is no longer cleared at fetch start, so a retry/background refresh keeps the cold-load error state visible until a successful portfolio response installs real data. The new regression test covers that exact pending-retry window.

What changed:
- `useEulerAccount` now tracks thrown portfolio fetches separately from partial per-position pricing diagnostics.
- `hasPortfolioLoadError` is only true when the latest fetch threw and no stale portfolio exists to fall back to.
- `pages/portfolio.vue` shows an explicit portfolio-load error banner plus placeholders instead of confident `$0` / `0.00%` values.
- Retry uses a fresh preemptive portfolio refresh.

Validation performed:
- Diff and surrounding-code review for `composables/useEulerAccount.ts`, `pages/portfolio.vue`, and `tests/composables/useEulerAccount.test.ts`.
- Focused regression: `npx vitest run tests/composables/useEulerAccount.test.ts` — passed, including cold-load failure and pending-retry persistence.
- `npm run lint -- composables/useEulerAccount.ts pages/portfolio.vue tests/composables/useEulerAccount.test.ts` — exited 0; only pre-existing repo-wide `any` warnings surfaced outside this PR’s changed files.
- `npm run build` — passed; build emitted existing chunk/sourcemap-style warnings only.
- Headed Chromium/Xvfb smoke against a local production build: `/portfolio?network=1&spy=0x68e7...8264` on desktop and mobile rendered the portfolio shell/tabs without material page errors. This was route/mobile coverage, not wallet/signing coverage.

- Sibling portfolio value/performance surfaces on the top-level portfolio page are covered by the shared `hasPortfolioLoadError` gate, not separate one-off guards per metric.
- Position-detail `hasQueryFailure` handling and portfolio-item query-failure handling are related but not the same data boundary: they cover incomplete position/vault data after a portfolio exists, while this PR covers the cold portfolio-fetch failure before there is any portfolio to render. I do not see a missed sibling surface that would still show the same cold-load `$0` bug.
- The retry-persistence invariant is now in a focused composable test, which is the right level for this state machine.

Bot-feedback pass:
- CodeRabbit’s earlier major concern about clearing the error at fetch start was valid on the old head and is fixed here.
- Its remaining placeholder-character nit is low-value/non-blocking; I don’t see it changing correctness or user safety for this fix.

Coverage caveats:
- I did not run wallet/signing smoke; this PR changes read-only portfolio load/error presentation only.
- I did not post screenshots: the meaningful error state was validated through a controlled composable failure/retry test rather than a synthetic browser network-failure screenshot, and the browser pass was only route/mobile sanity coverage.

PR #722, opened 2026-07-10
fix: bind prepared plans to wallet context

Seranged:
## Summary
- Prepared-plan execution proceeds only while the connected wallet account and network match the reviewed envelope.
- Signing and transaction callbacks remain bound to the prepared account and chain throughout execution.

## Changes
- Read the current wallet snapshot at execution entry and immediately before every prepared signature or transaction send.
- Reject missing or mismatched wallet account and chain data without substituting the browsed network.
- Keep raw-plan execution unchanged and cover account drift, chain drift, unavailable wallet-chain state, and matching-context execution.

## Test plan
- [x] npm run test:run -- tests/composables/useEulerTx.test.ts
- [x] npm run test:run
- [x] npm run typecheck
- [x] npm run lint
- [x] git diff --check origin/development...HEAD

### DIFF
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -1268,12 +1268,11 @@ export const useEulerTx = () => {
       ? prepared.account
       : prepared.account.owner
     const assertPreparedWalletContext = () => {
-      const currentOwner = requireOwner()
-      const currentChainId = requireChainId()
-      if (getAddress(currentOwner) !== getAddress(preparedOwner)) {
+      const currentWallet = getAccount(config)
+      if (!currentWallet.address || getAddress(currentWallet.address) !== getAddress(preparedOwner)) {
         throw new Error('Wallet account changed since this transaction was prepared. Review the transaction again.')
       }
-      if (currentChainId !== prepared.chainId) {
+      if (currentWallet.chainId === undefined || currentWallet.chainId !== prepared.chainId) {
         throw new Error('Wallet network changed since this transaction was prepared. Review the transaction again.')
       }
     }
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -58,6 +58,11 @@ const SUB_ACCOUNT_SNAPSHOT_FETCH_OPTIONS = {
   populateUserRewards: false,
 } as const
 type PrefetchPluginAccount = Account<IHasVaultAddress> | Address
+type PreparedWalletContext = {
+  account: Address
+  chainId: number
+  assertCurrent: () => void
+}
 
 const isOkxWallet = async (connector?: { id?: string, name?: string, getProvider?: () => Promise<unknown> }) => {
   if (!connector) return false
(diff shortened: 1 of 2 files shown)


composables/useEulerTx.ts
```diff
@@ -1250,15 +1264,38 @@ export const useEulerTx = () => {
     if (isSpyMode.value) {
       throw new Error('Transactions are disabled in spy mode')
     }
+    const preparedOwner = typeof prepared.account === 'string'
+      ? prepared.account
+      : prepared.account.owner
+    const assertPreparedWalletContext = () => {
+      const currentOwner = requireOwner()
+      const currentChainId = requireChainId()
```
LeonardEulerXYZ: **Blocking — require the connected wallet chain, not the browsing fallback.** `requireChainId()` reads `wagmiChainId ?? addressesChainId`. During a wallet transition `wagmiChainId` can be unavailable while the route chain still equals `prepared.chainId`, so this assertion passes and the subsequent typed-data callback can prompt a signature without confirming the wallet network (the signing API is bound only to `account`). A local regression setting the wallet chain to `undefined` resolved successfully and called signing instead of rejecting. Please validate against a wallet-only source here (for example the current wagmi account snapshot) and fail closed when that chain is absent; add the undefined-chain case to the focused tests.

LeonardEulerXYZ (changes requested):
Reviewed head `2ce6f3ee82de95500748190397ff763032ea9e50`.

The prepared-account binding and per-callback rechecks are the right shape, and all located `executePreparedPlan` consumers inherit them centrally. One fail-closed gap remains: the network check can substitute the browsed route chain when the connected wallet chain is unavailable, so a prepared typed-data signature may still be requested without confirming the wallet network. See the inline finding.

Validation:
- `git diff --check origin/development...HEAD`
- `npm run test:run -- tests/composables/useEulerTx.test.ts` — 3/3 passed
- `npm run build` — passed
- local throwaway regression with wallet chain `undefined` — failed as expected: execution resolved and called signing instead of rejecting; test was removed afterward

Scalability / maintainability hygiene: prepared-plan callers are covered through the shared composable. Raw-plan and CoW executors retain separate callback paths and are outside this PR's stated prepared-plan scope; this should not be read as application-wide wallet-context binding.

Bot feedback: CodeRabbit reported no actionable comments, but its pass did not cover the unavailable-wallet-chain fallback above.

Smoke coverage: focused code/test trust-boundary validation only. Browser/mobile and wallet signing smoke were not run because the material branch requires mutating wallet connection state between preparation and the signing callback; the focused regression is the discriminating check. Screenshots are not applicable.

composables/useEulerTx.ts
Seranged: Fixed in 3383a17f. Prepared execution now validates the imperative Wagmi account snapshot before entry, every signature, and every send; a missing wallet chain fails closed. Added a regression covering chainId becoming undefined before signing. Focused tests (4/4), typecheck, focused ESLint, and all PR checks pass.

LeonardEulerXYZ (commented):
Reviewed current head `3383a17f1c07d65d738a0355d383fcd0f797ac86`.

Verdict: the prior blocker is resolved, and I found no remaining blocker in the scoped prepared-plan path.

The follow-up now reads the imperative Wagmi account snapshot, fails closed when account or chain is absent/mismatched, repeats that assertion before every prepared signature/send, and binds both Wagmi mutations to the prepared account (plus chain for sends). The new missing-chain regression covers the failure mode from my earlier review.

Validation:
- `npm run test:run -- tests/composables/useEulerTx.test.ts` — 4/4 passed
- `npx eslint composables/useEulerTx.ts tests/composables/useEulerTx.test.ts` — passed
- CI lint/typecheck/test and Railway dev build — green for this head

Scalability / maintainability hygiene: all located `executePreparedPlan` consumers inherit the guard centrally, including batch/action/migration flows. The active raw `executePlan` fallbacks and standalone migration-authorization signing remain separate, pre-existing paths and are not hardened by this PR; that matches the stated “raw-plan execution unchanged” scope, so this verdict should not be read as application-wide wallet-context binding. A useful non-blocking follow-up would be to bind those fallback callbacks similarly. Connector identity is also outside the asserted account+network context; with the same account/chain, switching connectors can change which connector handles the request and can make the once-sampled OKX delay classification stale.

Bot feedback: CodeRabbit had no actionable inline findings; its walkthrough is consistent with the current prepared-path implementation.

Smoke coverage: focused wallet-boundary code/test validation only. Browser/mobile and live wallet-signing smoke were not run because a static route load cannot exercise wallet account/network mutation between preparation and callback execution. Screenshots are not applicable.

This COMMENT review supersedes my old blocker; no approval was submitted under the automatic-review policy.

LeonardEulerXYZ (approved):
Approving the previously reviewed current head `3383a17f1c07d65d738a0355d383fcd0f797ac86`. The completed review found no blocking correctness, security, or maintainability issues, and its recorded validation passed.

LeonardEulerXYZ:
**Leonard review correction:** I retract my approval in review #4673194847. Leonard is not authorized to approve Euler Lite PRs; the intended verdict remains **COMMENT / no blocking findings**. A repository maintainer must dismiss the submitted approval because GitHub does not permit this account to dismiss or delete it.

PR #725, opened 2026-07-10
fix: isolate label loading by chain

Seranged:
## Summary
- Scope in-flight label fetches by target chain so rapid network switches cannot share or publish wrong-chain results.
- Keep label readiness, loading, snapshot, and wrap-pair state aligned with the latest active chain load.

## Changes
- Reuse only the underlying fetch for the same target chain while per-call generations control publication.
- Guard state writes and asynchronous wrap-pair probes with load generation, probe generation, loaded-label chain, and current-chain checks.
- Cover A to B, A to B to A, force-refresh replacement, and stale-probe races with deterministic composable tests.

## Test plan
- [x] npm run test:run -- tests/composables/useEulerLabels.test.ts tests/composables/useEulerAccount.test.ts
- [x] npm run typecheck
- [x] npx eslint composables/useEulerLabels.ts tests/composables/useEulerLabels.test.ts
- [x] git diff --check

## Residual
- PR #619 also changes useEulerLabels and currently has a conflicting merge state; whichever change lands second must preserve these request-scoping guards during conflict resolution.

### DIFF
--- a/composables/useEulerLabels.ts
+++ b/composables/useEulerLabels.ts
@@ -52,7 +52,8 @@ const labelsChainId = ref<number | null>(null)
 const labelsVersion = ref(0)
 const isLoading = ref(false)
 const isReady = ref(false)
-let pendingLabelsLoad: Promise<void> | undefined
+const pendingLabelsFetches = new Map<number, Promise<EulerLabelsData>>()
+let labelsLoadGeneration = 0
 let wrapPairProbeGeneration = 0
 const wrapPairs = shallowReactive<Record<string, string>>({})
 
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (commented):
No blocking findings at `4b4712052630f9762b9c2a5b817938909ee86abb`.

The per-chain in-flight map and latest-load generation guard keep products/entities/vault-label publication coherent across A→B and A→B→A switches. The identity-checked cleanup preserves a force-refresh replacement, and the additional chain/load/probe checks prevent stale wrap-pair writeback. The map remains bounded by the configured chain set and settled entries are removed.

Validation performed:
- `npm run test:run -- tests/composables/useEulerLabels.test.ts tests/composables/useEulerAccount.test.ts tests/composables/useVaults.test.ts tests/composables/useGeoBlock-wrap-pair.test.ts tests/utils/euler-labels-utils.test.ts tests/composables/useEulerOracleAdapters.test.ts` (44 tests passed)
- `npm run typecheck`
- `npx eslint composables/useEulerLabels.ts tests/composables/useEulerLabels.test.ts`
- `git diff --check origin/development...HEAD`

I did not run browser/API smoke: the changed behavior is deterministic in-flight ordering and stale-publication suppression, which the deferred unit tests exercise directly; the live labels proxy cannot reliably reproduce those races.

dglowinski (approved):

PR #726, opened 2026-07-10
fix: sanitize vault validation errors

Seranged:
## Summary
- Route vault-validation failures through the structured server logger so provider errors are reduced to safe, useful metadata.
- Keep vault routes fail-open when upstream validation cannot complete.

## Changes
- Log the middleware context, chain ID, and viem failure through the top-level `err` serializer.
- Cover a credential-bearing `HttpRequestError` with real Pino serialization, including host/status retention and RPC credential/request-payload redaction.

## Test plan
- [x] `npx vitest run tests/server/ensure-vault.test.ts tests/utils/logger.test.ts tests/utils/viem-errors.test.ts`
- [x] `npm run typecheck`
- [x] `npx eslint server/middleware/ensure-vault.ts tests/server/ensure-vault.test.ts`
- [x] `git diff --check`

### DIFF
--- a/server/middleware/ensure-vault.ts
+++ b/server/middleware/ensure-vault.ts
@@ -2,6 +2,7 @@ import { getRequestURL, sendRedirect } from 'h3'
 import { VaultType } from '@eulerxyz/euler-v2-sdk'
 import { getAddress, isAddress, type Address } from 'viem'
 import { withWallClock } from '../utils/fetchWithTimeout'
+import { logger } from '../utils/logger'
 import { getServerSdk } from '../utils/sdk-server'
 
 /**
@@ -26,7 +27,7 @@ const resolveVaultTypes = async (
     )
   }
   catch (err) {
-    console.warn('[ensure-vault] failed to validate vault route', err)
+    logger.warn({ ctx: 'ensure-vault', chainId, err }, 'failed to validate vault route')
     return undefined
   }
 }
--- a/tests/server/ensure-vault.test.ts
+++ b/tests/server/ensure-vault.test.ts
@@ -1,0 +1,119 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { HttpRequestError } from 'viem'
+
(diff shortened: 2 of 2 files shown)


server/middleware/ensure-vault.ts
```diff
@@ -26,7 +27,7 @@ const resolveVaultTypes = async (
     )
   }
   catch (err) {
-    console.warn('[ensure-vault] failed to validate vault route', err)
+    logger.warn({ ctx: 'ensure-vault', chainId, err }, 'failed to validate vault route')
```
LeonardEulerXYZ, line 30: **Medium — generic errors can still leak sensitive request details.** This catch accepts any throwable, while `server/utils/logger.ts` retains `summary.shortMessage`; for a non-Viem `Error`, `classifyViemError` copies `error.message` verbatim. An SDK/network error whose message contains a credential-bearing RPC URL or calldata will therefore be emitted in `err.shortMessage`. The new regression test only covers `HttpRequestError`, whose Viem short message is the fixed safe string `HTTP request failed.` Please make the server serializer's short message safe by construction (fixed classification text, removal, or robust redaction) and add a generic-`Error` regression case containing URL credentials and calldata.

LeonardEulerXYZ (changes requested):
Reviewed at `863b48af9ef5432675a42088587da41cfed0b342`.

**Verdict: changes requested** — the canonical Viem `HttpRequestError` path is safely reduced, but the catch accepts arbitrary errors and the shared serializer preserves generic `Error.message` verbatim as `shortMessage`. That leaves the stated log-sanitization boundary incomplete; see the inline finding.

Validation:
- focused ensure-vault/logger/Viem error tests: 42 passed
- Nuxt typecheck: passed
- focused ESLint: passed
- `git diff --check`: passed
- repository CI and preview deployment: green

The structured `err` field correctly activates Pino's serializer, and fail-open route behavior is preserved. Browser smoke was not run because this is server-log serialization rather than UI behavior.

PR #728, opened 2026-07-10
fix: charge RPC batches by operation cost

Seranged:
## Summary
- Make RPC proxy rate-limit consumption reflect every batch operation and the upstream work each request can trigger.
- Reject unbounded log queries and oversized simulation inputs before budget consumption or an upstream fetch.

## Changes
- Assign method-specific base costs and add bounded modifiers for calldata, explicit gas, state overrides, log ranges, and fee history.
- Cap RPC params, calldata, gas, override accounts and storage, fee-history depth, and log-filter breadth.
- Preserve the Viem and Wagmi request shapes used by Lite, including state overrides, block-hash logs, and latest-block polling.
- Add deterministic route coverage for exact costs, 100-operation batches, validation boundaries, and limiter rejection.

## Test plan
- [x] npm run test:run -- tests/server/rpc-proxy-route.test.ts
- [x] npm run test:run
- [x] npm run typecheck
- [x] npm run lint
- [x] git diff --check

### DIFF
--- a/server/api/internal/rpc/[chainId].ts
+++ b/server/api/internal/rpc/[chainId].ts
@@ -27,6 +27,25 @@ const ALLOWED_METHODS = new Set([
 ])
 
 const MAX_BATCH_SIZE = 100
+const MAX_PARAMS_BYTES = 256 * 1024
+const MAX_CALLDATA_BYTES = 256 * 1024
+const MAX_EXPLICIT_GAS = 50_000_000n
+const MAX_STATE_OVERRIDE_ACCOUNTS = 128
+const MAX_STATE_OVERRIDE_SLOTS = 2_048
+const MAX_STATE_OVERRIDE_CODE_BYTES = 64 * 1024
+const MAX_FEE_HISTORY_BLOCKS = 128n
+const MAX_REWARD_PERCENTILES = 20
+const MAX_LOG_BLOCK_RANGE = 10_000n
+const MAX_LOG_ADDRESSES = 20
+const MAX_LOG_TOPIC_POSITIONS = 4
+const MAX_LOG_TOPIC_OR_VALUES = 20
+
+const CALLDATA_COST_CHUNK_BYTES = 16 * 1024
+const GAS_COST_CHUNK = 5_000_000n
+const STATE_OVERRIDE_SLOT_COST_CHUNK = 16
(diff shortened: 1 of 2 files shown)


server/api/internal/rpc/[chainId].ts
```diff
@@ -42,6 +61,242 @@ interface JsonRpcRequest {
...
+  for (const key of ['data', 'input'] as const) {
+    if (transaction[key] === undefined) continue
+    calldataBytes += hexDataByteLength(transaction[key], 'RPC calldata')
+  }
+  if (calldataBytes > MAX_CALLDATA_BYTES) {
+    return invalidParams(`RPC calldata exceeds ${MAX_CALLDATA_BYTES} bytes`, 413)
+  }
+
+  let cost = steppedModifier(calldataBytes, CALLDATA_COST_CHUNK_BYTES)
+  if (transaction.gas !== undefined) {
```
Comment (LeonardEulerXYZ) [inline server/api/internal/rpc/[chainId].ts:159]: **Medium — omitting `gas` bypasses the execution-cost charge.** `gas` is optional for `eth_call`; when absent, the RPC node still executes with its configured/default call gas cap, which can be as expensive as the maximum allowed explicit gas. Here the modifier only runs when the client supplies `transaction.gas`, so an adversary can remove that field and reduce a potentially max-cost call to the base cost of 2. Please conservatively charge omitted gas against an appropriate default/cap (or otherwise bound the upstream execution), and add a regression test proving that removing `gas` cannot make the same expensive call materially cheaper.

LeonardEulerXYZ (changes requested):
Reviewed at `8d707f6025ed73eb9bb9895477a1cc0e02097f4e`.

**Verdict: changes requested** — batch aggregation, parameter bounds, state-override accounting, bounded log filters, and fee-history limits are well covered, but `eth_call` can bypass the gas-cost modifier by omitting the optional `gas` field. That preserves the expensive upstream execution while receiving only the cheapest call charge; see the inline finding.

Validation:
- focused RPC route tests: 37 passed
- Nuxt typecheck: passed
- focused ESLint: passed
- `git diff --check`: passed
- repository CI and preview deployment: green

The 100-operation batch is now charged operation-by-operation, malformed/oversized inputs are rejected before limiter consumption/upstream fetch, and the abstraction remains centralized. Browser screenshots are not applicable to this server/API cost-accounting change.

PR #730, opened 2026-07-10
fix: classify SDK 1.1.4 migration and rewards queries in the cache policy

Seranged:
Supersedes #729: bumps `@eulerxyz/euler-v2-sdk` 1.1.2 → 1.1.4 directly and classifies the queries the bump introduces, addressing the blocking review on #729 (the position-migration queries had no `SDK_QUERY_POLICY` classification).

## Problem

SDK 1.1.3 routes six new `PositionMigrationService` queries (`queryListPositions`, `queryListTargets`, `queryGetPosition`, `queryGetAuthorization`, `queryEulerTargetVaultData`, `queryEulerSourceVaultAssets`) through Lite's `buildQuery`. With no policy row, both `getEulerSdk()` and `getEulerSdkFresh()` fell back to `DEFAULT_STALE_TIME_MS` (5 min) on the shared query client, and no post-tx eviction applied. Reproduced locally: two fresh-wrapper calls with identical inputs ran the underlying read once, and a browsing-SDK read seeded the cache the plan-time SDK then reused. Migration planning could therefore reuse stale external position balances (debt accrues per block) and authorization state, and a completed migration kept serving the pre-migration position for up to 5 minutes.

A completeness sweep also found four rewards queries silently defaulting since SDK 1.1.2: `queryFuulClaimChecks`, `queryFuulTotals`, `queryTurtleMerkleProofs`, `queryTurtleStreams`.

## Fix

- **Policy rows for all ten queries.** Position/authorization reads take balance-class windows (1 min browse / 15 s plan-time) with `invalidateAfterTx`; discovery lists take 5 min / 1 min with `invalidateAfterTx`; Euler-side target vault data takes the governance-config class (5 min / 1 min); source vault assets (immutable) 5 min flat. Fuul/Turtle rewards take the existing rewards default, and `queryFuulClaimChecks` joins the post-claim invalidation list in `useSdkRewards`.
- **Completeness test.** `tests/utils/sdk-query-policy.test.ts` now builds the real SDK offline with a recording `buildQuery` and fails when any wrapped query name lacks a policy row, or when a row matches no wrapped name. A future SDK bump adding queries fails CI until they are classified.
- **SDK 1.1.4** adds `PositionMigrationService` to the `EulerSDKQueryName` union (euler-xyz/euler-sdks#75), so the new rows typecheck without a local type widening. The obsolete `as EulerSDKQueryName` casts in `useSdkRewards` are dropped.

[...]

PR #727, opened 2026-07-10
fix: pin Doppler CLI artifact

Seranged:
## Summary
- Pins Doppler CLI 3.76.0 to fixed release artifacts and checksums for amd64 and arm64 images.
- Restricts the production image to a verified static Doppler binary for its target architecture.

## Changes
- Uses a dedicated build-platform stage with automatic TARGETARCH selection.
- Downloads through HTTPS-only redirects with fail-on-error behavior.
- Verifies literal architecture-specific SHA-256 values, extracts only doppler, applies mode 0755, and rejects unsupported architectures.
- Copies the verified binary into the distroless production image.

## Test plan
- [x] Verify the downloaded amd64 and arm64 archives against the pinned SHA-256 values.
- [x] Inspect the extracted binaries as statically linked Linux x86-64 and aarch64 executables with mode 0755.
- [x] Execute the Dockerfile architecture guard with TARGETARCH=386 and confirm exit status 1 before download.
- [x] Confirm no remote Doppler installer URL remains and run git diff --check.
- [ ] Run Docker Buildx targets and a full-image version smoke test; Docker and Buildx are unavailable in this environment.

PR #719, opened 2026-07-09
test(golden): reduce the tx-plan golden suite to a version canary; drop legacy worktree

Seranged:
## Background

The golden suite (`tests/golden/tx-plan-builders.test.ts`) began as legacy→SDK *migration parity* scaffolding: it diffed the current SDK planners against the deleted legacy `useEulerOperations` builders, imported from a pinned pre-migration git worktree via a separate vitest config, excluded from the normal test run.

Two facts make a comprehensive version of this the wrong tool now:

1. **The migration is done** — the legacy builders are gone, so parity-against-legacy has no remaining purpose.
2. **The SDK already tests its own builders comprehensively** — `euler-xyz/euler-sdks` ships `test/executionService.test.ts` (~85KB) covering every `plan*` builder, including CoW and cross-protocol paths. Re-asserting that calldata here just duplicates the SDK's suite.

## What this does

Reduces the suite to a **version canary** and removes the legacy apparatus.

- **Canary, not correctness.** Keeps one representative fixture per encoder family — deposit, borrow+collateral, repay, same-asset migration, collateral swap, multiply-with-swap (6 tests). Its stated job: catch an SDK bump (dependabot / preview-SDK) that silently changes the byte-for-byte calldata a Lite user signs — the one thing the SDK's own tests can't do in this repo's CI.
- **Legacy removed.** The `useEulerOperations` imports, the legacy `OperationsContext` harness, `normalizeLegacyPlan`, the legacy-only mocks, `vitest.golden.config.ts`, the legacy-worktree setup, and the golden-only vitest setup are all gone.
- **Fixture-based.** Each test runs the SDK planner, normalizes the resolved plan, and asserts it against a committed `plans/*.json`. Regenerate deliberate changes with `npm run test:golden:update`; a missing fixture is a hard failure, never a silent self-approval.
- **Folded into the main config** (dropped the `tests/golden` exclude), so it runs in `npm run test:run` and the existing CI `test` job — no separate config, worktree, or job.
- Trimmed the unused plan + swap-quote fixtures and `fetch-fixtures.ts` scenarios to match, and dropped the stale migration-parity `COVERAGE.md`.

## Why not comprehensive

Builder *correctness* is the SDK's responsibility and is already tested upstream. What this repo uniquely needs to guard is the SDK/app boundary across upgrades — which a small canary does at a fraction of the maintenance cost. (Testing euler-lite's own `useEulerTx` wrapper/branch-selection layer is worthwhile too, but that's a separate composable-level suite, not this.)

[...]

PR #714, opened 2026-07-09
fix: log v3 proxy failure context

Seranged:
## Summary
- Add structured V3 proxy failure context so timeout/non-ok alerts identify the public vault endpoint that failed.
- Keep logs bounded to allowlisted fields instead of raw request paths or account addresses.

## Changes
- Include `v3ChainId`, `v3VaultAddress`, `v3VaultKind`, and safe range/pagination fields on V3 upstream failure logs.
- Cover vault totals graphs, open-interest, and bad-debt request shapes while omitting dynamic account addresses.
- Add V3 proxy utility coverage for the new log-field extraction.

## Test plan
- [x] `./node_modules/.bin/vitest run tests/server/v3-proxy.test.ts tests/server/v3-proxy-route.test.ts`
- [x] `npx eslint server/utils/v3-proxy.ts 'server/api/internal/v3/[...path].ts' tests/server/v3-proxy.test.ts`
- [x] `npm run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of V3 proxy requests so failures are logged with more consistent request details.
  * Enhanced safe request metadata capture for proxy-related errors, helping support troubleshooting without exposing sensitive values.

PR #731, opened 2026-07-10
chore: promote development to master

Seranged:
## Fixes
- Bump `@eulerxyz/euler-v2-sdk` 1.1.2 → 1.1.4 and classify its position-migration and rewards queries in the cache policy so migration planning never reuses stale position balances or authorization state (#730) @Seranged
  - SDK 1.1.3 routes position-migration reads through `buildQuery` for caching and deduplication, and fixes max same-asset collateral migrations to skim the full redeemed balance; 1.1.4 adds the migration queries to `EulerSDKQueryName`
- Isolate label loading by chain so rapid network switches cannot share or publish wrong-chain label results (#725) @Seranged
- Expose the connector address in EIP-55 checksummed form at the wagmi boundary, removing the raw-casing comparison bug class at the source (#717) @Seranged
- Add structured V3 proxy failure context (chain, vault, safe request-shape fields) so timeout/non-ok alerts identify the failing endpoint (#714) @Seranged
- Pin the Doppler CLI to fixed 3.76.0 release artifacts with per-architecture SHA-256 verification in the production image (#727) @Seranged

## Refactors
- Default stable-to-stable swap contexts to 0.05% slippage, with legacy override storage migrated into the explicit override shape (#446) @Seranged

## Tests
- Reduce the golden tx-plan suite to an SDK version canary now that the SDK owns builder correctness (#719) @Seranged

## Validation
- `npm run typecheck`
- `npm run test:run` - 131 files passed; 1247 tests passed
- `npm run build`
- Validated at `107ef2cb`, the `development` head of this PR


kasperpawlowski (approved):

Seranged:
Post-merge review of this release diff (12 candidate findings adversarially verified; none release-blocking). Two minor follow-ups survived:

1. **`composables/useSlippage.ts:178` — legacy override migration edge (plausible, low severity).** A pre-existing custom slippage in the (0.05, 0.3] band (e.g. 0.2%) is migrated with `defaultSlippageAtSet = DEFAULT_SLIPPAGE (0.3)`. On a stable-to-stable pair, `isSlippageOverrideActive` returns false (value 0.2 > stable default 0.05, and 0.3 !== 0.05), so the sync watchEffect nulls the override and clears storage — the user's 0.2% is silently reset to the 0.05% stable default. The old code kept any value ≤ 0.3 active on every pair. Failure direction is safe (tighter slippage), and the migration tests only cover 0.01 and 3, sidestepping this band — worth either asserting as intended or preserving mid-band overrides.

2. **`server/utils/v3-proxy.ts:36` — duplicated regexes (confirmed cleanup).** `ADDRESS_RE = /^0x[a-fA-F0-9]{40}$/` and `INTEGER_RE = /^[0-9]+$/` are byte-identical to the private copies in `server/utils/observability.ts` (and `rewards-proxy-allowlist.ts` carries a third address copy). The same v3 handler invokes both classifiers, so the copies can drift independently — export them from `observability.ts` and reuse.

PR #743, opened 2026-07-21
fix: keep Keyring onboarding state coherent

Seranged:
## Summary
- Keep Keyring-gated operations blocked while hook metadata and credential status are loading.
- Route extension installation through the Keyring SDK so install and return navigation use the supported flow.
- Reject extension credentials that do not match the active wallet, policy, and network.

## Changes
- Reconcile install, progress, incomplete, error, and ready extension states through one scoped handler.
- Surface status-check loading and incomplete verification feedback in the form.
- Add focused composable coverage for loading gates, SDK launch, extension loss, incomplete checks, and credential context validation.

## Test plan
- [x] `npm run test:run` — 1,377 passed, 1 skipped
- [x] `npm run typecheck`
- [x] ESLint on the changed Keyring files and tests
- [ ] Complete a real Keyring KYC and credential-injection transaction with an eligible CEX account

## Summary by CodeRabbit

- **New Features**
  - Added clearer status messages and loading indicators during keyring verification.
  - Verification errors now display specific messages when available.
  - The keyring extension can be launched directly from the verification flow.
  - Verification now better handles wallet, policy, network, and extension status changes.

- **Bug Fixes**
  - Improved transitions and recovery when verification is incomplete or the extension becomes unavailable.

- **Tests**
  - Added coverage for keyring verification states and credential validation.

### DIFF
--- a/composables/useKeyring/index.ts
+++ b/composables/useKeyring/index.ts
@@ -5,6 +5,7 @@ import {
   KeyringConnect,
   type CredentialData,
   type ExtensionSDKConfig,
+  type ExtensionState,
 } from '@keyringnetwork/keyring-connect-sdk'
 import { keyringHookTargetAbi } from '~/abis/keyring'
 import { isVaultKeyring } from '~/utils/eulerLabelsUtils'
@@ -23,6 +24,19 @@ export enum KeyringFlowState {
   Ready = 'ready',
   Error = 'error',
 }
+
+export const isCredentialForContext = (
+  credential: CredentialData | null | undefined,
+  trader: string | undefined,
+  policyId: number | undefined,
+  chainId: number | undefined,
+): credential is CredentialData => Boolean(
+  credential
+  && trader
+  && credential.trader?.toLowerCase() === trader.toLowerCase()
+  && credential.policyId === policyId
+  && credential.chainId === chainId,
+)
 
 const readHookTargetField = async <T>(
   rpcUrl: string,
@@ -60,8 +74,9 @@ export const useKeyring = (vaultAddress: string | Ref<string>) => {
   const credentialData = ref<CredentialData | null>(null)
   const flowState = ref<KeyringFlowState>(KeyringFlowState.Idle)
   const error = ref<string>()
+  const isCheckingStatus = ref(false)
+  const statusMessage = ref<string>()
 
-  let statusCheckInterval: ReturnType<typeof setInterval> | null = null
   let unsubscribeExtension: (() => void) | null = null
 
   const isKeyringVault = computed(() => isVaultKeyring(addressRef.value))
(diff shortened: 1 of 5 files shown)


composables/useKeyring/index.ts
```diff
@@ -289,12 +344,15 @@ export const useKeyring = (vaultAddress: string | Ref<string>) => {
...
       credentialData.value = null
       hasValidCredential.value = false
-      if (isKeyringVault.value && hookTarget.value && userAddress.value) {
-        checkCredential()
-      }
-      else if (!isKeyringVault.value) {
+      if (!isKeyringVault.value || !userAddress.value) {
         flowState.value = KeyringFlowState.Idle
       }
+      else if (hookTarget.value) {
```
LeonardEulerXYZ, line 350: **Blocking — obsolete checks can clear the gate for a new context.** This watcher starts `checkCredential()` for every wallet/chain/vault change, but earlier calls remain able to write `policyId`, `hasValidCredential`, `credentialData`, and `flowState`. I reproduced: wallet A's check was held pending, wallet B switched in and reached `Install`, then A resolved `true`; shared state became `hasValidCredential = true` / `Idle`, so B was no longer blocked. Please invalidate prior async work (or compare a captured wallet + chain + hook target/policy generation before every state commit) and add an out-of-order context-switch test. A simple global in-flight-promise reuse is insufficient because the new context must receive its own check.

LeonardEulerXYZ (changes requested):
## Review summary

**Verdict: changes requested** — the new loading gate is directionally correct, but an older credential check can still overwrite a newer wallet/network/vault context and clear the gate.

### Blocking finding
- `useKeyring` starts a new asynchronous onchain check for each context change without invalidating earlier work or verifying the captured context before committing results. A slow check for wallet A can resolve after wallet B's check and set `hasValidCredential = true` / `flowState = Idle` for B. This is a fail-open UI state for Keyring-gated operations; the eventual transaction may still revert onchain, but the client has incorrectly admitted an unchecked context. See the inline comment.

### Validation
- Focused `useKeyring` suite: 9/9 passed.
- Local throwaway wallet-switch race regression: reproduced the blocker (`hasValidCredential` became `true` after the obsolete wallet-A request resolved). The throwaway test was removed afterward.
- Typecheck: passed.
- `git diff --check`: passed.
- Hosted lint, typecheck, test, and dev-build checks: green for the reviewed head.

The credential-context matcher is sensibly centralized, and direct Keyring UI consumption remains behind `useOperationGuard`; I found no new one-off display helper. The missing reusable invariant is generation-safe async state: every credential check and extension callback should be tied to the wallet, chain, vault/hook target, and policy that launched it, with a focused context-switch regression test. CodeRabbit also flagged stale/overlapping work; that direction is valid, but merely reusing one in-flight promise would be unsafe across context changes because it can preserve the old context rather than invalidate it.

### Smoke coverage
Browser/mobile/wallet smoke not run: reproducing the changed flow requires an eligible Keyring vault/account and extension credential, while the blocking race is deterministic at the composable boundary. No screenshots were posted.

PR #746, opened 2026-07-21
feat: require scrolling terms to bottom before accepting

kasperpawlowski:
## Summary

Disables the **Accept** button in the *Acknowledge terms* modal until the user has scrolled the terms text to the bottom, so the full disclosure is surfaced before it can be accepted.

## Behaviour

- Accept starts disabled and unlocks once the terms container is scrolled to its end (with an 8px tolerance for sub-pixel rounding / momentum scrolling).
- Once the end has been reached, Accept stays enabled even if the user scrolls back up.
- If the terms content is short enough that it does not overflow (nothing to scroll), Accept is enabled immediately — handled on mount and via a resize observer, so viewport changes are covered too.
- Reject is unaffected.

## Implementation

- `AcknowledgeTermsModal.vue`: track scroll position on the terms container via a template ref + `@scroll` handler, gate the Accept button with `:disabled`, and re-check on mount and container resize (`useResizeObserver`).
- Extracted the scroll-position check into a pure, framework-agnostic helper (`termsScrollGate.ts`), following the existing sibling-module pattern (`cowSwapReviewState.ts`).

## Tests

- Added `tests/components/termsScrollGate.test.ts` covering top / partway / exact-bottom / within-threshold / beyond-threshold / non-overflowing / custom-threshold cases.
- `vitest` (new test), `eslint`, and `nuxt typecheck` all pass.

## Summary by CodeRabbit

* **New Features**
  * Terms acceptance is now scroll-gated in the acknowledgement modal: **Accept** remains disabled until the user reaches the end of the terms.
  * The button can unlock automatically when the content doesn’t overflow, and stays enabled once unlocked.
* **Bug Fixes**
  * Improved “scrolled to bottom” detection using a small threshold to handle rounding/sub-pixel scrolling differences.
  * Re-evaluates unlock state on initial render and when the terms area is resized.
* **Tests**
  * Added unit tests covering top/middle/bottom, non-scrollable content, default threshold, and custom threshold behavior.

### DIFF
--- a/components/entities/operation/AcknowledgeTermsModal.vue
+++ b/components/entities/operation/AcknowledgeTermsModal.vue
@@ -1,4 +1,7 @@
 <script setup lang="ts">
+import { useResizeObserver } from '@vueuse/core'
+import { isScrolledToEnd } from './termsScrollGate'
+
 const props = defineProps<{
   onAccept?: () => void
   onReject?: () => void
@@ -33,6 +36,27 @@ const terms = [
     text: `I acknowledge that Safe Harbor programs have separate rules, and by using the Protocol, I consent to the Whitehat Agreement terms, including eligible fund rescues, bounty deductions, associated risks (including loss and taxes), and holding Euler Community Members harmless.`,
   },
 ]
+
+// Gate the Accept button until the user has scrolled the terms to the bottom,
+// so it can only be clicked after the full text has been surfaced. Once the end
+// has been reached it stays unlocked, even if the user scrolls back up.
+const termsContainer = ref<HTMLElement | null>(null)
+const hasScrolledToEnd = ref(false)
+
+const checkScrolledToEnd = () => {
+  const el = termsContainer.value
+  if (!el || hasScrolledToEnd.value) {
+    return
+  }
+  if (isScrolledToEnd(el)) {
+    hasScrolledToEnd.value = true
+  }
+}
+
+// Re-evaluate on mount and whenever the container is resized (e.g. viewport
+// changes) so a non-overflowing list unlocks the button immediately.
+useResizeObserver(termsContainer, checkScrolledToEnd)
+onMounted(() => nextTick(checkScrolledToEnd))
 
 const onReject = () => {
   if (props.onReject) {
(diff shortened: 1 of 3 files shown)


components/entities/operation/AcknowledgeTermsModal.vue
```diff
@@ -78,7 +102,11 @@ const onAccept = () => {
         I further represent and warrant:
       </div>
 
-      <div class="bg-surface-secondary rounded-12 overflow-y-auto flex-1 styled-scrollbar">
+      <div
+        ref="termsContainer"
+        class="bg-surface-secondary rounded-12 overflow-y-auto flex-1 styled-scrollbar"
```
LeonardEulerXYZ, line 110: **Blocking — preserve keyboard access across browsers.** This required-scroll region has no explicit focus target. The smoke passes in current Chromium because Chrome 132+ makes generic scrollers keyboard-focusable by default, but the element still reports `tabIndex === -1`; relying on that engine behavior is not portable to browsers/configurations that skip overflow `<div>` elements during Tab navigation. There, focus moves from the Risk Disclosures link to Reject/disabled Accept, so a keyboard-only user cannot change this region's `scrollTop` and cannot accept the terms. Please make the scroller explicitly focusable (for example `tabindex="0"`), give it an accessible label/region semantics and visible focus treatment, and cover Tab + End/PageDown unlocking in a component/browser regression.

LeonardEulerXYZ (dismissed):

Reviewed head `23dd9b5f418da707d0e9f55379c6f022415dcc69`.

The scroll-end calculation and sticky unlock behavior are sound, but the new required interaction needs an explicit keyboard focus target rather than relying on Chromium's automatic focusability for generic scrollers. See the inline finding.

### Validation
- `npm run test:run -- tests/components/termsScrollGate.test.ts` — 7/7 passed
- GitHub CI `lint`, `test`, and `typecheck` — passed on the reviewed head
- Headed Chromium interaction smoke at desktop 1280×900 and mobile 390×844:
  - Accept starts disabled while terms overflow
  - keyboard Tab + End reaches the bottom and enables Accept
  - unlock remains sticky after scrolling back up
  - closing/reopening resets Accept to disabled
  - no page errors

Smoke coverage: desktop browser visual smoke + mobile browser/keyboard smoke. Wallet/signing smoke was not applicable.

### Scalability / maintainability
The behavior is centralized in the one shared `AcknowledgeTermsModal` used by operation flows; the pure scroll predicate is extracted and has focused boundary tests. I found no sibling terms-acceptance implementation carrying the old behavior. CodeRabbit's suggestion for a committed modal integration test is reasonable and non-blocking; the browser smoke above verifies the wiring today, while a regression test should also cover keyboard focus, sticky unlock, and reopen reset.

### Visual evidence
Desktop — initial/top state (Accept disabled):

Desktop initial terms state

Desktop — keyboard-scrolled end state (Accept enabled):

Desktop end terms state

Mobile — initial/top state (Accept disabled):

Mobile initial terms state

Mobile — keyboard-scrolled end state (Accept enabled):

Mobile end terms state

Seranged (approved):

PR #766, opened 2026-07-29
Speed up first batch additions

dglowinski:
## Summary
- Reuse the account snapshots the forms already loaded when adding the first batch entry, instead of refetching them at add-time.
- Remove the redundant `fetchAccount({ populateAll: true })` from the user-visible Add to batch path while preserving safe fallbacks.

## Changes
- Add a neutral prefetch registry (`batchPrefetchState`) for the planning account (`useFreshAccount`) and the enriched base account (`useEulerAccount`). No composable dependencies, so the prefetch hand-off does not add another import cycle.
- Read those accounts **only** from the registry, never per entry. `usePlanAccount().account` and `useEulerAccount().portfolio` are layer-aware — they resolve to the active *simulated* layer once the batch is non-empty — so they must never become the batch's own layer 0. The registry holds pre-overlay loader output by construction.
- Validate chain + owner before reuse (`resolvePrefetchedAccounts`), shared by the `addEntry` preflight, `getEntryPlanningAccount`, and `resimulate`. A wallet switch replaces the loaders' accounts asynchronously, so the batch can observe the previous wallet's snapshot with the new owner already active.
- Keep the enriched base account fixed for the cart lifetime and simulate directly against it.
- Prime ERC20 slot hints from the Earn and wallet-repay forms as the lend page already did, with `background: true` so speculative page-load probes no longer register as pending work.

### Behaviour changes worth flagging
- **The first entry's plan is now built against a different account shape.** It was `fetchAccount({ populateAll: true })`; it is now the planning account's `PLAN_ACCOUNT_FETCH_OPTIONS` (`populateVaults: true`, no market prices or user rewards). This is the same account the non-batch submit path already builds against via `usePlanAccount`, so plan construction with this shape is already exercised in production — but it is a behaviour change, not purely a caching win.
- **Layer 0 is shape-equivalent.** `portfolioService.fetchPortfolio` internally calls `fetchAccount(chainId, address, { populateAll: true })`, identical to `fetchBaseAccountSnapshot`, so reusing the portfolio account as layer 0 does not change what the simulated layers stitch onto. It also comes from the same `getEulerSdkFresh` instance (`source: 'fast'` has no production caller).
- **Speculative priming no longer gates submit.** `isResolvingStateOverrideHints` is a global counter that hard-disables submit and add-to-batch on every mounted form and replaces the tooltip with the generic "Complete the form fields above to continue." Page-load priming now opts out via `background: true`; on a cold cache the simulator falls back to access-list discovery, as it did before priming existed. This also fixes the pre-existing flash on the lend page.

### Incidental fix
- `primeSlotHintsFor` snapshotted `slotHints.value` *before* awaiting its probes, so two concurrent primes clobbered each other's results. It now merges after the await. Covered directly.

### Deliberately not included
- A batch-side mirror of slot hints. The SDK's `fetchErc20SlotHints` already memoises module-scope by `chainId:token`, so once any form primes a token the batch's own probe short-circuits to a Map lookup with no RPC. Mirroring it would buy the removal of an `await`, not a round-trip, at the cost of a chain-keyed Map, a three-way merge with ambiguous precedence, and its own teardown. The batch still probes only tokens missing from `batchSlotHints`, which is what the simulator reads.

## Test plan
- [x] Full suite: 1489 passed, 1 skipped.
- [x] `npm run typecheck`.
- [x] ESLint against every changed file.
- [x] Add a deposit to a batch in Chrome and verify the portfolio redirect and simulated-state overlay.
- Coverage added for: prefetched reuse with no add-time `fetchAccount`, layer 0 seeding for account-free entries, rejection of wrong-chain and wrong-owner prefetches, chain-switch hint isolation, the concurrent-prime merge, and background vs foreground pending-work accounting.

### DIFF
--- a/composables/batchPrefetchState.ts
+++ b/composables/batchPrefetchState.ts
@@ -1,16 +1,28 @@
-import type { Account, IHasVaultAddress, SlotHints } from '@eulerxyz/euler-v2-sdk'
+import type { Account, IHasVaultAddress } from '@eulerxyz/euler-v2-sdk'
 
 /**
- * Form-load data that the batch builder can safely reuse on the first add.
+ * Form-load accounts the batch builder can safely reuse on the first add.
  *
  * This module deliberately has no composable dependencies. Account, wallet, and
  * batch composables import each other for their simulated-state overlays, so a
  * small neutral registry keeps the prefetch hand-off free of another import
  * cycle.
+ *
+ * Both accounts are stored *pre-overlay* — they are written by the loaders
+ * (`useFreshAccount`, `useEulerAccount`), never read back off the layer-aware
+ * `usePlanAccount` / `useEulerAccount().portfolio` computeds. That matters:
+ * those computeds return the active batch layer's *simulated* account when a
+ * layer is active, which must never become the batch's own layer 0. Consumers
+ * still have to validate chain + owner before reuse (see `isAccountForContext`
+ * in `useTxBatch`) because a wallet or chain switch can land before the
+ * matching loader run replaces these.
+ *
+ * Slot hints are deliberately *not* mirrored here: the SDK's `fetchErc20SlotHints`
+ * already memoises module-scope by `chainId:token`, so once any form has primed a
+ * token the batch's own probe short-circuits to a Map lookup with no RPC.
  */
 let planningAccount: Account<IHasVaultAddress> | undefined
 let baseAccount: Account<IHasVaultAddress> | undefined
-const slotHintsByChain = new Map<number, SlotHints>()
 
 export const setBatchPrefetchedPlanningAccount = (
   account: Account<IHasVaultAddress> | undefined,
@@ -28,22 +40,12 @@ export const setBatchPrefetchedBaseAccount = (
 
 export const getBatchPrefetchedBaseAccount = () => baseAccount
 
-export const mergeBatchPrefetchedSlotHints = (
-  chainId: number,
-  slotHints: SlotHints,
-) => {
-  slotHintsByChain.set(chainId, {
-    ...slotHintsByChain.get(chainId),
-    ...slotHints,
-  })
-}
-
-export const getBatchPrefetchedSlotHints = (chainId: number): SlotHints => ({
-  ...slotHintsByChain.get(chainId),
-})
-
+/**
+ * Drops both prefetched accounts. Production resets flow through the owning
+ * loaders (`useFreshAccount.reset`, `useEulerAccount.resetLoadingState`); this
+ * exists so tests can isolate module state between cases.
+ */
 export const resetBatchPrefetchState = () => {
   planningAccount = undefined
   baseAccount = undefined
-  slotHintsByChain.clear()
 }
--- a/composables/batchPrefetchState.ts
+++ b/composables/batchPrefetchState.ts
@@ -1,0 +1,49 @@
+import type { Account, IHasVaultAddress, SlotHints } from '@eulerxyz/euler-v2-sdk'
+
+/**
+ * Form-load data that the batch builder can safely reuse on the first add.
+ *
+ * This module deliberately has no composable dependencies. Account, wallet, and
+ * batch composables import each other for their simulated-state overlays, so a
+ * small neutral registry keeps the prefetch hand-off free of another import
+ * cycle.
+ */
+let planningAccount: Account<IHasVaultAddress> | undefined
+let baseAccount: Account<IHasVaultAddress> | undefined
+const slotHintsByChain = new Map<number, SlotHints>()
+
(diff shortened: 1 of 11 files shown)


LeonardEulerXYZ (changes requested):
## Review: changes requested

The optimization is sound in direction, and the owner/chain checks plus fallback paths are well structured. Two correctness boundaries still need fixing before merge:

1. **Batch resimulation now gives SDK plugins a cart-pinned `Account`.**  
   At [`useTxBatch.ts:1675`](https://github.com/euler-xyz/euler-lite/pull/766#discussion_r3672890446), replacing `ownerAddr` with `baseAccount` changes Pyth plugin resolution: SDK `1.2.1` trusts an already-populated `Account`, but re-fetches current controllers/collaterals when given the owner address. Consequently, resimulation can select feeds from the stale layer-0 snapshot while execute-time preparation still resolves from the address. Keep `baseAccount` for layer-0 stitching, but pass `ownerAddr` to `simulateTransactionPlan` unless preview/execute plugin divergence is explicitly intended.

2. **Local slot hints are not isolated by chain.**  
   At [`useStateOverrideOptions.ts:77`](https://github.com/euler-xyz/euler-lite/pull/766#discussion_r3672890459), `next` contains the composable's entire uncleared hint bag and is merged under the current `cid`. I reproduced: prime token A on chain 1, switch the same composable to chain 8453, prime token B, and chain 8453's registry now contains token A's chain-1 storage indices. A later batch can therefore skip probing and forge the wrong ERC20 balance/allowance slots. Clear/localize hints on chain change, or merge only the hints returned by this call into the chain-scoped registry.

### Independent verification at `284dd1e1`

- GitHub CI: lint, typecheck, tests all green; Railway preview deployed successfully.
- Local focused tests: **44/44 passed** (`useTxBatch`, `useEulerAccount`, `useWalletRepay`).
- Local `npm run typecheck`: passed.
- Local production build: passed.
- SDK freshness probe: confirmed address input invokes `fetchAccount`, while a populated `Account` bypasses it.
- Chain-isolation regression probe: failed as described above (chain-1 hint present in chain-8453 bucket).
- Preview smoke: app and lend form loaded; no changed-flow transaction was sent.

I completed a second whole-diff pass beyond these two threads and found no additional merge-blocking issue. Once both boundaries have regression tests and are fixed, this should be straightforward to re-review.

LeonardEulerXYZ (approved):
## Re-review: approved

Reviewed current head `e41de6ba67c188cc31cf158c53c637e25e4c4048`. Both prior blockers are resolved:

- Batch resimulation again passes `ownerAddr` into `simulateTransactionPlan`, so SDK plugins resolve current account context, while the prefetched `baseAccount` remains pinned for layer-0 stitching.
- Slot probing now merges only call-local results into the captured chain bucket, clears the local mirror synchronously on chain changes, and prevents late old-chain probes from repopulating current-chain local state.

The new regressions cover both an ordinary chain switch and an in-flight probe completing after the switch.

### Verification

- Focused tests: **46/46 passed** (`useStateOverrideOptions`, `useTxBatch`, `useEulerAccount`, `useWalletRepay`)
- `npm run typecheck`: passed
- Targeted ESLint: passed
- `npm run build`: passed
- Full diff check / security-pattern sweep: clean
- GitHub CI lint, typecheck, and tests: green
- Railway PR preview deployment: green

No remaining merge-blocking findings from this review.

LeonardEulerXYZ (approved):
## Re-review: approved

Reviewed current head `9c1354883705194a074b24ae3ba48e47eb5d8a53` and the full PR against `development`.

The follow-up hardens the first-entry fast path cleanly:

- Entry callers can no longer supply layer-aware planning/base accounts. `useTxBatch` resolves only pre-overlay loader snapshots from the neutral registry and validates both owner and chain before reuse.
- Account-free first entries seed layer 0 through the same resolver, preserving the no-refetch optimization.
- The redundant cross-composable slot-hint registry is gone. The batch resolves only approval tokens missing from its own simulator input; already-primed tokens hit the SDK's chain/token cache.
- Speculative lend, earn, and wallet-repay priming no longer globally disables submit/add-to-batch while RPC discovery is in flight.
- Earn now skips native-currency probing, and local concurrent primes merge after `await` instead of dropping one another.

### Verification

- Focused tests: **49/49 passed** (`useStateOverrideOptions`, `useTxBatch`, `useEulerAccount`, `useWalletRepay`)
- `npm run typecheck`: passed
- ESLint over every changed file: passed
- `npm run build`: passed
- Full diff check / sensitive-pattern sweep: clean
- GitHub lint, typecheck, tests, CodeRabbit, and Railway preview: green

No remaining merge-blocking findings from this review.

Seranged (approved):

PR #775, opened 2026-07-31
Security fixes - RPC/Reward provider links

kanvgupta:
RPC secret-safe logging
reward URL scheme validation
script-safe chain configuration
 encoded V3 path rejection

## Summary by CodeRabbit

- **Security**
  - Unsafe or malformed external campaign URLs are now blocked.
  - Inline configuration is safely escaped to prevent script injection.
  - Proxy requests with encoded path characters are rejected earlier.

- **Bug Fixes**
  - Vault validation warnings now use structured application logging.

- **Tests**
  - Added coverage for URL validation, configuration security, logging practices, and proxy request handling.

### DIFF
--- a/entities/reward-campaign.ts
+++ b/entities/reward-campaign.ts
@@ -6,6 +6,7 @@ import type {
   RewardSource as SdkRewardSource,
   UserReward as SdkUserReward,
 } from '@eulerxyz/euler-v2-sdk'
+import { safeExternalHttpUrl } from '~/utils/external-url'
 
 export type { RewardAction }
 
@@ -104,7 +105,8 @@ export const rewardCampaignToken = (campaign: RewardCampaign): RewardCampaignDis
 })
 
 export const rewardCampaignSourceUrl = (campaign: RewardCampaign): string | undefined => {
-  if (campaign.sourceUrl) return campaign.sourceUrl
+  // V3/provider data reaches multiple `href` bindings through this chokepoint.
+  if (campaign.sourceUrl) return safeExternalHttpUrl(campaign.sourceUrl)
   if (campaign.source === 'merkl') return undefined
   return campaign.source === 'turtle'
     ? `https://dashboard.turtle.xyz/organizations/${TURTLE_DASHBOARD_ORGANIZATION_ID}/incentives/streams/${campaign.campaignId}`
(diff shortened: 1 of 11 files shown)


LeonardEulerXYZ (commented):
## Leonard review — draft-safe verdict

**No blocking findings on `ba635a99cbad63a350c8d3edf1ca3c76c5358d56`.**

I reviewed all 11 changed files and traced the four affected boundaries through their consumers:

- reward campaign `sourceUrl` → displayed provider links
- app/chain runtime config → inline script payloads
- inbound V3 proxy path → validation, target construction, and API-key attachment
- vault-route failures → structured error serialization

### Validation performed

- `git diff --check origin/development...HEAD` — pass
- Node `v24.18.1` / npm 11:
  - `npm run test:run` — **79 files passed, 1 skipped; 1,599 tests passed, 4 skipped**
  - `npm run typecheck` — pass
  - `npm run build` — pass
- Targeted security suites — **65/65 passed**
- Railway preview smoke:
  - valid `/api/internal/v3/tokens?chainId=1&limit=1` — 200
  - encoded slash, backslash, and NUL path variants — 400 before proxying
- Runtime injection probe with `APP_TITLE=</script><script>…</script>` — emitted JSON contained `\u003c`, no raw script delimiter, and round-tripped to the original value
- Live reward fixture check — sampled campaign source URLs were absolute HTTPS URLs; unsafe-scheme count was zero
- Current GitHub lint, typecheck, test, CodeRabbit, Cursor, and Railway checks are green

### Residual notes

The two CodeRabbit comments are reasonable test-strengthening suggestions—directly exercising the encoded-path branch and rendering plugin output in the regression test—but I do not consider either a merge blocker. I independently exercised both behaviors at runtime.

Review state is **COMMENT** rather than approval because the PR is still draft.

Seranged (approved):
All four hardening changes verified correct: traced every `sourceUrl` href sink through `rewardCampaignDisplays()`, confirmed the v3-proxy encoded-byte rejection breaks no legitimate client paths (only hex/decimal segments are ever encoded), confirmed `escapeScriptJson` composes correctly with the CSP nonce injection, and confirmed the logger switch actually redacts RPC API keys. Full suite passes.

Non-blocking follow-ups noted for a later PR:
- `ENCODED_PATH_BYTE_RE` check is duplicated in `v3-proxy.ts` (the copy in `validateV3ProxyUrl` path via `isV3ProxyPathAllowed` at line 75 is unreachable in production).
- `escapeScriptJson` re-export in `server/plugins/app-config.ts` exists only for a test import; the test should import the util directly.
- Two more remote-controlled href sinks could reuse the new sanitizer: `intrinsicApyInfo.source` in `VaultApyModal.vue` and `entity.url` in the vault overview components.
- `docs/architecture.md` Security Architecture table should mention the v3 proxy allowlist + encoded-byte rejection and the inline-config escaping.

PR #778, opened 2026-08-04
feat: show uncovered losses on earn vault statistics

kasperpawlowski:
## Summary

Adds an **Uncovered losses** row to the EulerEarn vault Statistics block. Until now `lostAssets` was not surfaced anywhere in Lite — the only reference in the repo was a test asserting that `update_lost_assets` stays out of the earn activity feed.

The value is shown in USD, using the same `formatAssetValue` → `formatCompactUsdValue` path as the Total supply and Available liquidity rows directly above it, with the same asset-denominated fallback when no price is available.

## Why it nets off address(1)

`EulerEarn` never writes `lostAssets` back down once a shortfall is covered — `_accruedFeeAndAssets` only ever increases it. Coverage instead shows up as shares parked at `address(1)`, which is the canonical method documented on `IEulerEarn.lostAssets`:

> In order to cover those lost assets, it is advised to supply on behalf of `address(1)` on the vault (canonical method).

Displaying the raw value would therefore permanently overstate the shortfall on any vault whose loss has been made good. The row instead shows `lostAssets − convertToAssets(balanceOf(address(1)))`, clamped at zero (coverage can overshoot, since donated shares keep earning after the fact).

For reference, euler-creator shows a "Lost Assets" row from the raw lens value with no netting — this is a deliberate improvement on that, which is also why the label is "Uncovered losses" rather than "Lost assets". A tooltip explains the mechanism.

## Cost

The `balanceOf` read is skipped entirely when `lostAssets` is zero, so the overwhelmingly common case — every healthy vault — costs no extra RPC. When it does fire, the shared public client already batches it (`batch: { batchSize: 100, wait: 100 }`). A failed read resolves to `undefined`, which falls back to treating the shortfall as *fully* uncovered rather than understating it.

## Changes

- `utils/vault/earn-losses.ts` (new) — `EARN_LOSS_COVERAGE_ADDRESS` plus the pure `computeUncoveredLosses` helper
- `composables/useEarnLossCoverage.ts` (new) — reads the coverage balance, race-guarded, `logWarn` on failure
- `components/entities/vault/overview/earn/VaultOverviewEarnBlockStats.vue` — the new row
- `tests/utils/vault/earn-losses.test.ts` (new) — 7 unit tests

## Follow-up

This belongs in the SDK long term, but it cannot fully land there yet. Lite runs `eulerEarnServiceAdapter` as either `v3` or `onchain` per chain. The on-chain adapter could add the `balanceOf` to its existing batched reads for free, but the V3 adapter maps `lostAssets` straight from the API response, so V3-backed chains would silently regress to the un-netted value until euler-data-v3 exposes the field. Reading on-chain here is correct under both backends today; once `uncoveredLostAssets` exists on `IEulerEarn` plus the matching API field, this composable deletes cleanly.

## Test plan

- [x] `npx vitest run tests/utils/vault/earn-losses.test.ts` — 7 pass, covering zero shortfall, no coverage, partial coverage, over-coverage clamp, non-1:1 share price conversion, and an unresolved read
- [x] `npm run typecheck` — 44 errors, identical to the untouched `development` baseline (pre-existing SDK drift in `utils/activity-display.ts` and `utils/sdk-query-policy.ts`); none in the files touched here
- [x] `npx eslint` clean on all four files
- [x] `npm run test:run` — remaining failures are pre-existing flaky timeouts on `development` (clean baseline on the same worktree failed 10; with these changes, 2)
- [ ] Manual: confirm the row renders `$0` on a healthy earn vault with no extra RPC call, and shows the netted figure on a vault with a recorded shortfall

## Summary by CodeRabbit

* **New Features**
  * Added an “Uncovered losses” statistic to vault overviews.
  * Displays formatted loss amounts, USD values, asset symbols, and explanatory tooltip text.
  * Preserves accurate values for supported off-chain assets.

* **Bug Fixes**
  * Prevented outdated asynchronous formatting results from overwriting newer values.
  * Improved handling when loss-value formatting fails.


LeonardEulerXYZ (changes requested):
## Review summary

**Changes requested** on `e00aa79c1a5a7118e051224e04f06127881a87d4`.

The loss-netting arithmetic and EulerEarn `address(1)` mechanism are sound. One reactive-state bug remains: while a new loss-bearing vault's `balanceOf(address(1))` call is pending, the composable continues exposing the previous vault's coverage shares. The overview then combines the new vault with the old vault's shares and can temporarily understate the current vault's uncovered loss.

The race guard prevents an old RPC response from being committed later, but it does not invalidate the already-committed value when a new valid read starts. Clear the result before the await and render a placeholder while `isCoverageLoading` is true; reserve the conservative `undefined -> full lostAssets` fallback for a settled failed read. Please add a composable regression test that switches from a covered vault to an uncovered vault while the second read is pending.

### Independent reproduction

A throwaway composable-level Vitest reproduced the bug on this exact head:

- vault A resolves with `coverageShares = 90n`
- switch to vault B and leave B's read pending
- `isCoverageLoading === true`, but `coverageShares === 90n` instead of `undefined`

The test failed with `expected 90n to be undefined`. The probe was removed afterward; the review worktree is clean.

### Live protocol fixture

Direct latest-state calls for mainnet TelosC Surge USDC (`0x49C5733d71511A78a3E12925ea832f49031c97e9`) returned:

- `lostAssets = 238869503`
- address(1) shares = `238870000`
- `convertToAssets(shares) = 244068111`
- expected uncovered assets = `0`

This confirms the intended netting path against a real fully covered loss.

### Validation

- `git diff --check origin/development...HEAD` - pass
- `npx vitest run tests/utils/vault/earn-losses.test.ts` - 7/7 pass
- `npm run typecheck` - pass
- `npm run test:run` - 169 files, 1,591 tests pass
- `npm run build` - pass
- targeted ESLint on all four changed files - pass
- Browser smoke: incomplete; the review environment blocked preview HTTP access, so I did not retry through another route.

I did not elevate the async formatter comment to an additional blocker: the current `off-chain` formatter path has no delayed network I/O, the same pattern already exists for adjacent rows, and I did not reproduce a production-path stale overwrite. A generation guard would still be reasonable hardening.

LeonardEulerXYZ (changes requested):
## Rereview summary

**Changes requested** on `8bea11aeba23662353a406ea11d3fc7bcb8a23b5`.

The original data-layer blocker is fixed: `useEarnLossCoverage` now clears the previous vault's shares before awaiting, distinguishes loading from settled failure, and has a discriminating vault-switch regression test. The live TelosC Surge USDC preview also settles to `Uncovered losses: $0` as intended.

One new blocker is introduced by the formatter/loading composition in this head. The loading branch sets the display to `-` but does not advance `uncoveredLossesGuard`. If a formatter started before loading is still pending, it remains the guard's current generation and can resolve afterward, replacing the loading placeholder with the previous vault's formatted value. This reopens a narrower stale cross-vault display window.

A deterministic reviewer-only test using the exact `watchEffect` + `runGuarded` control flow failed with:

```
expected 'stale-vault-value' to be '-'
```

Invalidate the guard when entering the loading branch (or at the beginning of every effect run) and add a regression test that leaves the old formatter pending, flips loading to true, then resolves the old formatter. The reviewer probe was removed afterward; the worktree is clean.

### Validation

- Incremental fix delta and complete base-to-head diff reviewed: 7 changed files
- `git diff --check` - pass
- committed focused suites - 20/20 pass
- `npm run typecheck` - pass
- `npm run test:run` - 1,596 passed, 1 skipped
- `npm run build` - pass
- targeted ESLint across all changed files - pass
- Railway preview - desktop route loaded; TelosC Surge USDC renders `Uncovered losses $0` with no visible layout/clipping defect
- Mobile browser smoke - not completed: the local Playwright browser revision and system Chromium are unavailable

No additional blockers found in the fresh whole-head pass.

composables/useEarnLossCoverage.ts
```diff
@@ -0,0 +1,66 @@
...
+    const rpcClient = client.value
+    const gen = guard.next()
+
+    if (!target || target.lostAssets <= 0n || !isAddress(target.address) || !rpcClient) {
+      coverageShares.value = undefined
+      isLoading.value = false
+      return
+    }
+
+    isLoading.value = true
```
Seranged: Resolved in 791cbf89 by removing this coverage-read path entirely. The EulerEarn lens subtracts convertToAssets(balanceOf(address(1))) before exposing lostAssets, and both the SDK on-chain adapter and V3 preserve that normalized value. Reading address(1) again in Lite double-counted coverage. Live checks matched V3 to the on-chain normalized result for uncovered and fully-covered vaults; the component now formats vault.lostAssets directly.

components/entities/vault/overview/earn/VaultOverviewEarnBlockStats.vue
```diff
@@ -31,6 +33,33 @@ watchEffect(async () => {
...
+const uncoveredLosses = computed(() => computeUncoveredLosses(vault, coverageShares.value))
+
+const uncoveredLossesDisplay = ref('-')
+const uncoveredLossesGuard = createRaceGuard()
+
+watchEffect(async () => {
+  // Withhold the figure until the coverage read settles. Mid-flight the shares
+  // read as `undefined`, which would render the gross shortfall and overstate
+  // what is unbacked on a vault whose loss has already been covered.
+  if (isCoverageLoading.value) {
```
Seranged: Resolved in 791cbf89. Removing the duplicate coverage RPC also removes the loading early-return that skipped guard advancement. Every uncovered-loss effect run now enters runGuarded, so a superseded formatter cannot commit. The out-of-order regression remains in tests/utils/race-guard.test.ts; focused 15/15, full 1,588/1,588, typecheck, lint, and production build pass. A local live-vault smoke rendered raw 6361518648400 as $6.36M.

LeonardEulerXYZ (approved):
## Rereview summary

**Approved** on `791cbf89ac14d34adfd6f5f4806177c903841666`.

Both previous blockers are resolved by removing the duplicate Lite-side coverage read. The current component formats `vault.lostAssets` directly and protects asynchronous formatting with the shared race guard. This is the correct boundary: EulerEarn’s lens already subtracts `convertToAssets(balanceOf(address(1)))`, and SDK v1.2.3 preserves that normalized value through both its onchain and V3 adapters.

### Direct parity proof

At mainnet block `25687571`, TelosC Surge USDC (`0x49C5733d71511A78a3E12925ea832f49031c97e9`) returned:

- raw vault `lostAssets`: `238869503`
- coverage-address shares: `238870000`
- `convertToAssets(coverageShares)`: `244068657`
- expected normalized loss: `0`
- deployed EulerEarn lens `lostAssets`: `0`
- hosted V3 Earn detail `lostAssets`: `0`

I also traced the exact installed SDK (`@eulerxyz/euler-v2-sdk` 1.2.3, tag commit `1430b56d`) through `EulerEarnOnchainAdapter`, `convertEulerEarnVaultInfoFullToIEulerEarn`, and `EulerEarnV3Adapter`; both paths populate the entity’s `lostAssets` from the already-normalized source field.

### Validation

- complete base-to-head diff and incremental `8bea11ae..791cbf89` fix reviewed
- `git diff --check origin/development...HEAD` — pass
- focused race/pricing suites — 15/15 pass
- `npm run typecheck` — pass
- `npm run lint` — pass (six pre-existing warnings, zero errors)
- `npm run test:run` — 1,584 passed, 1 skipped
- `npm run build` — pass
- GitHub lint/typecheck/test checks — green

### Rendered smoke

Headed Chromium under Xvfb against the local production build, route:

`/earn/0x49C5733d71511A78a3E12925ea832f49031c97e9?network=1`

- real V3 response: `data-value="0"`, rendered `Uncovered losses $0`
- browser-boundary positive fixture (`lostAssets = 123456789`): rendered `Uncovered losses $123.43` on desktop
- same positive fixture: rendered `$123.43` inside the mobile **Vault information** drawer
- visual captures show the row and tooltip icon without clipping or overlap
- no page errors or Euler API failures; only expected AppKit 403s from the intentionally blank local project ID

No remaining findings.

Seranged (approved):

kanvgupta (approved):

dglowinski (approved):

PR #780, opened 2026-08-06
Keep rEUL claims and unlocks outside transaction batches

dglowinski:
## Summary

- Run rEUL reward claims and rEUL unlocks as standalone transactions.
- Require any pending transaction batch to be cleared before either rEUL action.
- Keep other reward claims batchable as usual.
- Require fresh, non-worsening lock data for every rEUL unlock review and confirmation.
- Render the SDK rEUL unlock operation as `Unlock` with the reviewed EUL amount.

## Changes

- Hide `Add to batch` for rEUL claim cards and rEUL unlock cards.
- Disable standalone rEUL `Claim` or `Unlock` while the transaction cart has entries and show an inline `Clear batch` action.
- Re-check the cart before opening or confirming either standalone rEUL transaction.
- After an rEUL claim, clear stale lock rows and await their replacement before the claim flow finishes.
- Re-fetch the selected lock before opening and confirming an unlock; abort when the lock is missing, unavailable, has different principal, or has a worse burn or unlockable quote.
- Accept natural monotonic vesting improvement between review and confirmation.
- Keep the shared lock poller attached to a live consumer until the final subscriber unmounts.
- Use the configured chain-scoped rEUL address as the authority once deployment config is available.
- Pass `allowRemainderLoss: true` when building the reviewed rEUL unlock plan.
- Decode `withdrawToByLockTimestamp` batch items as `Unlock` and attach the reviewed EUL amount to the transaction-review row.

## Test plan

- [x] Focused unlock-review and transaction-step decoding tests: 25/25
- [x] Full test suite: 169 files, 1,590 tests
- [x] Typecheck
- [x] Changed-file ESLint
- [x] Production build

## Summary by CodeRabbit

* **New Features**
  * Added refreshed lock validation before rEUL unlocks and claims.
  * Added clear-batch actions and safeguards when transactions are already batched.
  * Added clearer handling for changed, missing, or unavailable locks.
  * Added rEUL unlock transaction recognition in activity details.
* **Bug Fixes**
  * Prevented unlocking or claiming with outdated lock information.
  * Ensured lock data refreshes correctly after claims and during active loading.
* **Tests**
  * Expanded coverage for lock refreshes, batch behavior, validation, and transaction decoding.


LeonardEulerXYZ (changes requested):
## Review summary

**Verdict: changes requested — one transaction-liveness blocker.**

The address/provider matching and explicit `allowRemainderLoss: true` wiring are directionally sound. The serialized in-builder check also correctly notices a reward claim that finishes queueing after the user clicks Unlock. However, its rejection currently corrupts cart-level execution state.

### Blocking finding

`components/entities/reward/RewardUnlockItem.vue:127-129`

When an rEUL claim is already queued, the unlock builder throws the new separate-transaction guidance. `useTxBatch.addEntry()` catches every builder rejection and writes it to cart-level `simError` (`composables/useTxBatch.ts:2131-2141`). Because the unlock entry was never appended, `entries` remains the same claim-only cart; no entries watcher re-simulates it. `canExecuteBatch` and `executeBatch` both then refuse execution while `simError` is set (`:2332-2337`, `:2390-2396`).

Observable consequence: the UI tells the user to claim rEUL separately, but the still-valid claim already in the cart cannot be executed until the cart is cleared/rebuilt.

Please preserve both safety properties:

1. stable claim already present: reject before calling `addBatchEntry` so only the local toast changes;
2. claim addition still in flight: keep a serialized queue-time check, but make this policy refusal a non-cart-mutating add failure so it does not overwrite the healthy cart's simulation state.

Add a regression at the `useTxBatch`/component boundary that starts with an executable claim entry, rejects an unlock add, and asserts the claim remains executable with no new cart-level `simError`.

This independently confirms the same current-head issue already anchored at https://github.com/euler-xyz/euler-lite/pull/780#discussion_r3727061594; I am not adding a duplicate inline thread.

### Validation boundary

Reviewed exact head `4b7b2e7550c295d23bb7011a1b0f2350dc903e86` against `development`; inspected all five changed files plus `useTxBatch`, reward-provider mapping, and SDK v1.2.3 consumer compatibility. The incremental commit from `e42a4a2b` only adds Turtle to the helper test and does not affect this finding. Hosted lint, typecheck, test, and Railway preview checks are green. A local Lite dependency install/test rerun was not completed because the isolated `npm ci` confirmation timed out; no browser smoke was run. The blocker follows directly from the exact-head state transitions above and is not contradicted by the current focused tests, which do not exercise rejected-add liveness.

components/entities/reward/RewardUnlockItem.vue
```diff
@@ -119,12 +120,22 @@ const onAddToBatchClick = async () => {
     await addBatchEntry({
       label: 'Unlock rEUL',
       requiresPlanningAccount: false,
-      buildPlan: async () => buildUnlockREULPlan([item.timestamp]),
+      // This builder runs inside useTxBatch's serialized add queue. Checking
+      // here also covers a claim whose own add was still in flight when the
+      // user clicked Unlock.
+      buildPlan: async () => {
+        if (hasEarlierREULClaim(batchEntries.value, reulTokenContractAddress.value)) {
+          throw new Error('Claim rEUL in a separate transaction, then refresh the lock before unlocking it.')
```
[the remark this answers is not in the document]
dglowinski: Implemented on 14081ceb87d4c84e2d8807db2ad8f2c9049fdee2 as a standalone-only rEUL policy. rEUL claim and unlock cards have no Add to batch action; an active cart disables their direct action and offers Clear batch. Non-rEUL reward claims retain normal batching. rEUL unlock review and confirmation preserve lock principal, accept natural monotonic vesting improvement, and reject missing, unavailable, or worse fresh quotes. The shared lock poller remains attached to a live consumer until the final subscriber unmounts. The transaction-review decoder recognizes the SDK withdrawToByLockTimestamp batch item as Unlock and shows the reviewed EUL amount. Validation: focused 25/25, full 1,590/1,590, typecheck, changed-file ESLint, and production build.

LeonardEulerXYZ (changes requested):
## Re-review summary

**Verdict: changes requested — original cart-liveness blocker is resolved, but one remainder-loss review race remains.**

The new standalone-only policy removes rEUL claims and unlocks from transaction batches while retaining normal batching for non-rEUL rewards. Both click and confirmation paths re-check the cart, so the previously reported `simError` poisoning path is gone.

### Blocking finding: a post-claim refresh leaves stale burn data actionable

A standalone claim can still be followed by an unlock review built from the pre-claim lock row:

1. Reward claim execution calls `executePlan()` and then refreshes rewards (`PortfolioSdkRewardItem.vue:108-110`).
2. `executePlan()` synchronously triggers the portfolio refresh counter after the receipt (`useEulerTx.ts:1369-1380`).
3. The portfolio watcher starts `void refreshLocks(false)` (`pages/portfolio.vue:217-220`).
4. Passing `false` deliberately does not set `isLocksLoading` (`useREULLocks.ts:49-50,73-80`), so `pages/portfolio/rewards.vue:177-199` keeps rendering the old lock rows while that RPC call is in flight.
5. `RewardUnlockItem` displays `item.amountToBeBurned` from that row in the review (`:100-113,144-153`), but confirmation rebuilds the operation using only the lock timestamp (`:77-89`).
6. The resulting SDK call contains `lockTimestamp` plus `allowRemainderLoss: true`, with no maximum accepted burn (`useREULLocks.ts:145-154`; SDK `reulLockService.ts:120-139`).

Therefore the transaction can authorize the contract's current remainder loss even when it differs from the amount the user reviewed. Making claim and unlock separate transactions removes same-batch state drift, but does not itself guarantee a fresh quote between them.

Please make the freshness boundary explicit. For example:

- after an rEUL claim, await a lock refresh that disables/removes stale unlock rows; and
- before opening or confirming the unlock review, fetch/revalidate the current lock and abort for a new review if `amountToBeBurned` changed.

Add a discriminating regression with a deferred post-claim lock refresh: the stale row must not remain actionable, and confirmation must not execute when the current burn differs from the reviewed burn.

### Non-blocking note

`isREULReward` treats any token whose symbol case-folds to `reul` as rEUL even when its address differs from the configured token (`PortfolioSdkRewardItem.vue:61-65`). Prefer the chain-scoped configured address as the authority once available; symbol fallback can otherwise disable batching for an unrelated same-symbol token.

### Validation

Reviewed exact head `ab9835ef48315866fa4b916cdd7a50d4a0d3b518` against current `development`, including all three incremental commits and all four final changed files. Hosted lint, typecheck, test, CodeRabbit, and Railway preview checks are green. The final diff contains no component/integration regression for the new standalone and refresh behavior; the sole committed test change verifies `allowRemainderLoss: true` wiring in `useREULLocks`.

tests/components/reulUnlockReview.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      reviewedLock,
+      () => pendingRefresh,
+      execute,
+    )
+
+    expect(execute).not.toHaveBeenCalled()
+    resolveRefresh([{
+      ...reviewedLock,
+      unlockableAmount: 70n,
+      amountToBeBurned: 30n,
```
LeonardEulerXYZ, line 60: **Blocking test adequacy:** This deferred fixture changes both `unlockableAmount` and `amountToBeBurned`, so it does not prove the burn comparison. In a disposable exact-head mutant, I removed only `currentLock.amountToBeBurned !== reviewedLock.amountToBeBurned`; this test file still passed 2/2 because line 29 changed the unlockable amount. Please keep `unlockableAmount` equal to the reviewed value and change only `amountToBeBurned`, then assert `{ status: 'changed' }` and no execution.

LeonardEulerXYZ (changes requested):
## Re-review summary

**Verdict: changes requested — the production fix resolves the prior safety blockers, but the required burn-drift regression is not yet discriminating.**

Reviewed exact head `84f6202fc28b703efe404d2e80a444ba722ed266` against current `development`.

### Prior blockers

The implementation now preserves the intended safety invariants:

- rEUL claims and unlocks are standalone-only; ordinary non-rEUL reward batching remains available;
- both click and confirmation paths fail closed when a batch becomes active;
- a required post-claim refresh clears stale lock rows immediately and leaves them cleared on failure;
- unlock review and confirmation each fetch the current lock and refuse a changed/missing/unavailable quote;
- the configured chain-scoped rEUL address is authoritative once available.

### Remaining blocker: burn-only drift is not covered

The new deferred regression changes both `unlockableAmount` and `amountToBeBurned`. I removed only the production `amountToBeBurned` comparison in a disposable exact-head mutant and reran `tests/components/reulUnlockReview.test.ts`; both tests still passed. Thus the suite admits the precise unsafe regression this fix is meant to prevent: confirmation may proceed when the reviewed burn changes, provided the unlockable amount is unchanged.

Please add a deferred case that keeps `unlockableAmount` constant, changes only `amountToBeBurned`, and requires no execution plus `{ status: 'changed' }`. This is a test-only blocker; the current production comparison itself is correct.

### Validation

- focused changed tests: 5/5 passed;
- full suite: 1,587/1,587 passed;
- `npm run typecheck`: passed;
- ESLint on all six changed files: passed;
- `git diff --check`: passed;
- synthetic merge with current `origin/development`: clean;
- hosted lint, typecheck, test, CodeRabbit, and Railway build checks: green.

Browser smoke: not run — this re-review is a pure transaction-state/test-discrimination pass with no additional rendered-flow uncertainty. The reviewed worktree remained clean; the disposable mutant was removed.

tests/components/reulUnlockReview.test.ts
dglowinski: Fixed in 77d17108ff6f31f88d6ef927c571612bbcbf347e. The deferred fixture now keeps unlockableAmount at the reviewed 80 and changes only amountToBeBurned from 20 to 30. It still requires { status: changed } and verifies the execution callback is never called. Focused 5/5, full 1,587/1,587, ESLint, and diff checks pass. Production code is unchanged.

components/entities/reward/reulUnlockReview.ts
```diff
@@ -0,0 +1,40 @@
...
+  refreshLocks: RefreshREULLocks,
+): Promise<REULLockReviewValidation> => {
+  const refreshedLocks = await refreshLocks()
+  if (!refreshedLocks) return { status: 'unavailable' }
+
+  const currentLock = refreshedLocks.find(lock => lock.timestamp === reviewedLock.timestamp)
+  if (!currentLock) return { status: 'missing' }
+
+  if (
+    currentLock.unlockableAmount !== reviewedLock.unlockableAmount
```
LeonardEulerXYZ: [P1] Accept monotonic vesting instead of requiring exact quote equality

These fields are not stable while a lock is vesting. `REULLockService.fetchLocks()` reads `getWithdrawAmountsByLockTimestamp`, and the deployed `RewardToken` schedule derives the unlock share linearly from `block.timestamp` between day 1 and day 180. The review fetch and confirmation fetch therefore normally land on different values even when no claim or other state change occurred.

For example, a 1 rEUL lock at day 30 gains `51,727,705,359` wei of unlockable amount in one second, with burn decreasing by the same amount. This exact comparison consequently rejects an ordinary early unlock after the user has had time to read the modal.

Please preserve lock identity/principal and reject only a worse quote—for example, require `current.amount === reviewed.amount`, `current.unlockableAmount >= reviewed.unlockableAmount`, and `current.amountToBeBurned <= reviewed.amountToBeBurned`—then add a regression where the confirmation quote has naturally improved with time.

components/entities/portfolio/PortfolioSdkRewardItem.vue
```diff
@@ -38,11 +38,13 @@ const rewardClaimKey = computed(() => [
 ].join(':'))
 
 const { buildClaimRewardPlan, refreshRewards } = useSdkRewards()
-const { addEntry: addBatchEntry, entries: batchEntries } = useTxBatch()
+const { refreshLocks } = useREULLocks()
```
LeonardEulerXYZ, line 41: [P2] Do not let each reward row own teardown of the singleton lock poller

`useREULLocks()` stores `interval` at module scope, but every invocation registers an `onUnmounted` callback that clears that shared interval. This new call from every `PortfolioSdkRewardItem` means a normal claim/refresh can remove any reward row and silently stop rEUL polling for the still-mounted rewards page and remaining lock rows. Their active/chain watchers do not transition, so none restarts the interval.

Please keep this row on a refresh-only API passed from the page-owned subscriber, or make the shared polling lifecycle reference-counted/single-owner. A sibling-unmount regression test would make the ownership invariant explicit.

LeonardEulerXYZ (changes requested):
Reviewed exact head `77d17108ff6f31f88d6ef927c571612bbcbf347e` against `development`.

The earlier stale-lock/cart-liveness blocker is resolved: required refreshes now clear rendered stale rows, post-claim lock refresh is awaited, and missing/unavailable lock state fails closed.

Two blockers remain:

1. Confirmation requires exact equality for values that change every second under the rEUL vesting formula, so ordinary early unlocks are rejected after the modal is read.
2. Adding the module-global `useREULLocks()` poller to every reward row lets any claimed/unmounted row clear polling for still-mounted consumers.

Validation completed on the exact head:
- focused rEUL tests: 5/5 passed
- full suite: 1,583/1,583 passed across 168 files
- `npm run typecheck`: passed
- changed-file ESLint: passed
- `npm run build`: passed
- GitHub lint/typecheck/test and Railway preview checks: green
- SDK 1.2.3 and official `RewardToken` source at `evk-periphery@eb88a689` inspected for lock/burn semantics

The green tests do not exercise natural per-second quote improvement or shared-poller sibling unmount, which is why both regressions survive.

components/entities/reward/reulUnlockReview.ts
dglowinski: Fixed in b02521cb41f6921a25731a51661b7111672609bd. Revalidation now requires unchanged principal, current unlockableAmount greater than or equal to the reviewed amount, and current amountToBeBurned less than or equal to the reviewed burn. The new regression advances both values in the natural favorable direction and requires execution. Focused 7/7, full 1,589/1,589, typecheck, ESLint, and build pass.

components/entities/portfolio/PortfolioSdkRewardItem.vue
dglowinski: Fixed in b02521cb41f6921a25731a51661b7111672609bd. Poller callbacks are registered per mounted consumer; each interval tick selects a still-live callback, and teardown stops the shared interval only when the final consumer unregisters. The sibling-unmount regression proves polling continues after the first unmount and stops after the last. Focused 7/7, full 1,589/1,589, typecheck, ESLint, and build pass.

LeonardEulerXYZ (approved):
Current-head re-review of `b02521cb` against `development` (`883ac674`).

The two prior blockers are resolved:
- confirmation now accepts monotonic vesting progress only (same principal, no lower unlockable amount, no higher burn) while still failing closed on worsening, missing, or unavailable quotes;
- the module-global rEUL poller is now shared across consumers and remains alive until the final consumer unmounts.

Validation:
- focused rEUL tests: 7/7 passed;
- regression discrimination: the same tests fail on prior head `77d17108` exactly at the two repaired behaviors;
- Nuxt typecheck passed;
- ESLint passed on all six changed files;
- production build passed;
- synthetic merge with current `development` is conflict-free;
- hosted test, lint, typecheck, CodeRabbit, and Railway preview checks are green.

No blocking findings on this head.

LeonardEulerXYZ (commented):
Non-blocking follow-up from the independent current-head pass; this does **not** change the approval:

1. Each mounted reward row now owns a `useREULLocks()` watcher. An account or chain change can therefore start one lock fetch per consumer. `lockGuard` invalidates older responses but does not deduplicate requests, so a rate-limited newest request can leave locks empty even if an earlier duplicate succeeded. Consider centralizing the reactive watcher or deduplicating in-flight loads; a two-consumer account/chain-change test should assert exactly one additional fetch.

2. Releasing the final consumer stops polling but retains module-global `isLoaded` and `locks`. An A-unmount → B-remount flow can briefly show A's rows until the next 60-second poll. Confirmation still performs a required fresh fetch and fails closed, so this is stale cross-account UI rather than a transaction-safety blocker. Reset the cache on final release or key it by account+chain, with an A→B remount regression.

LeonardEulerXYZ (approved):
Current-head re-review of `14081ceb` against `development` (`883a730b`).

The incremental commit correctly recognizes `withdrawToByLockTimestamp(address,uint256,bool)` as an **Unlock** action in the operation review and displays the reviewed EUL asset/amount. The SDK's nested operation is flattened to the underlying call before selector decoding.

Validation:
- focused decoder + rEUL safety tests: 29/29 passed;
- regression discrimination: the new decoder assertion fails on prior head `b02521cb` at the intended `Unknown operation` behavior;
- Nuxt typecheck passed;
- changed-file ESLint passed;
- production build passed;
- synthetic merge with current `development` is conflict-free;
- hosted test, lint, typecheck, and Railway preview checks are green.

The previously noted cache/poller lifecycle items remain non-blocking follow-ups; this commit neither addresses nor worsens them. No blocking findings on this head.

LeonardEulerXYZ (approved):
Current-head re-review of `3150d5ba` against `development` (`883a730b`).

This commit cleanly completes the SDK/Lite integration by pinning `@eulerxyz/euler-v2-sdk` `1.2.4`.

Provenance and compatibility checks:
- npm `1.2.4` reports git head `aab1ddda`, the merged SDK #88 commit;
- downloaded tarball SHA-512 exactly matches `package-lock.json`;
- the published declaration requires `allowRemainderLoss: boolean`, and shipped JavaScript rejects non-booleans before encoding;
- Lite has one installed SDK copy at `1.2.4` and one production unlock caller, which passes `allowRemainderLoss: true` explicitly after displaying the burn in review.

Validation from a clean `npm ci --ignore-scripts` install:
- full Lite test suite passed;
- Nuxt typecheck passed;
- full ESLint passed;
- production build passed;
- synthetic merge with current `development` is conflict-free;
- hosted test, typecheck, CodeRabbit, and Railway preview are green (hosted lint was still running at review time; equivalent full local lint passed).

No blocking findings on this head.

composables/useREULLocks.ts
```diff
@@ -126,7 +142,10 @@ export const useREULLocks = () => {
   })
 
   onUnmounted(() => {
-    if (interval) {
+    if (released) return
+    released = true
+    pollConsumers.delete(consumerId)
+    if (pollConsumers.size === 0 && interval) {
```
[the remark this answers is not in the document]
dglowinski: Fixed in 03b7f403. Final-consumer teardown now advances lockGuard, resets isLoaded, locks, and loading state regardless of interval state, and clears the interval when present. The regression starts an in-flight refresh, tears down the final consumer, proves the stale completion returns null without repopulating state, and proves the next mount fetches fresh locks. Validation: focused 5/5, full 1,591/1,591, typecheck, and ESLint with 0 errors.

tests/composables/useREULLocks.test.ts
```diff
@@ -118,7 +124,67 @@ describe('useREULLocks', () => {
...
+      useREULLocks()
+      useREULLocks()
+    })
+
+    expect(unmountCallbacks).toHaveLength(2)
+    unmountCallbacks[0]?.()
+    expect(clearIntervalSpy).not.toHaveBeenCalled()
+    const callsBeforePoll = fetchLocks.mock.calls.length
+    await vi.advanceTimersByTimeAsync(60_000)
+    expect(fetchLocks.mock.calls.length).toBeGreaterThan(callsBeforePoll)
```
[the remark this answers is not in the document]
dglowinski: Fixed in 03b7f403. The poller regression now waits for the initial fetch to settle exactly once before unmounting the first sibling and recording the baseline, then requires exactly one additional fetch after the 60-second tick. Initial async completion can no longer satisfy the polling assertion. Validation: focused 5/5, full 1,591/1,591, typecheck, and ESLint with 0 errors.

LeonardEulerXYZ (approved):
Current-head re-review of `03b7f403` against `development` (`883a730b`).

The incremental fix correctly resets the module-global rEUL lock cache on final consumer release:
- advances the race generation so in-flight responses cannot repopulate stale rows;
- clears locks and loading state;
- marks the cache unloaded so the next consumer fetches immediately;
- preserves the shared poller until the final sibling unmounts.

Validation:
- focused rEUL + decoder tests: 30/30 passed;
- regression discrimination: the new final-unmount assertion fails on prior head `3150d5ba` because stale rows remain;
- Nuxt typecheck passed;
- changed-file ESLint passed;
- production build passed;
- synthetic merge with current `development` is conflict-free;
- hosted test, lint, typecheck, CodeRabbit, and Railway preview are green.

The separate duplicate-RPC-on-account/chain-change item remains a non-blocking follow-up and is unchanged by this commit. No blocking findings on this head.

LeonardEulerXYZ (approved):
Current-head re-review of `00533d1d` against `development` (`883a730b`).

This closes a transaction-review failure path: a build or simulation failure can no longer fall through to a review modal without a valid, simulated plan. Build failures surface a decoded preparation error; simulation failures surface their simulation message; only `{ status: 'ready', plan }` opens review.

Validation:
- focused rEUL + decoder tests: 32/32 passed;
- mutation discrimination: permissive build/simulation failure handling causes exactly the two new preparation tests to fail;
- Nuxt typecheck passed;
- changed-file ESLint passed;
- production build passed;
- synthetic merge with current `development` is conflict-free;
- CodeRabbit is green; hosted CI/Railway state was read immediately before submission and is recorded in the command output.

No blocking findings on this head.

Seranged (approved):

PR #782, opened 2026-08-06
Bundle review and execution safeguards

dglowinski:
## Summary

- make transaction review describe the exact approvals, payable calls, reward unlocks, and execution envelope sent to the wallet
- execute the prepared plan the user reviewed instead of rebuilding transaction-critical fields from reactive state at confirmation time
- bind every approval, signature, Safe submission, and raw send to the reviewed account and chain

## Issues and fixes

- **LITE-284 / EULE2-8 — Label approvals with the actual token.** Approval rows could display the operation's primary asset even when the wallet was authorizing a different collateral or Permit2 token. Each row resolves and displays its decoded token, spender, amount, symbol, decimals, and icon independently.
- **LITE-294 / EULE2-12 — Show Fuul claim fees in reward review.** Fuul claims could send a native fee that was absent from review. Payable display steps carry the exact native value and destination through direct-call and EVC-batch decoding.
- **LITE-278 / EULE2-27 — Prevent claim-before-unlock rEUL burns.** A claim followed by an unlock could change the lock balance after review and allow newly claimed rEUL to be burned or redirected under the unlock remainder policy. rEUL unlocks use freshly resolved lock state, remain isolated from unsafe reward batching, and preserve an executable batch when an unrelated unlock is rejected.
- **LITE-285 / EULE2-37 — Execute the exact reviewed plan.** Rebuilding from live quote, amount, position, or account refs at confirmation could produce calldata different from the reviewed simulation. Review stores a canonical prepared envelope and execution accepts only that envelope; narrowly permitted external inputs such as authenticated Pyth updates are refreshed without changing the reviewed operation identity.
- **LITE-287 / EULE2-38 — Pin raw execution to the reviewed wallet.** A wallet or network switch while review was open could route old calldata through the active wallet context. The reviewed owner and chain are captured and asserted before approvals, signatures, Safe submission, and broadcast.

## Changes

- decode rEUL unlocks and Fuul fees into actionable review rows instead of generic or incomplete operations
- fingerprint reviewed plans across account, chain, recipients, assets, amounts, quote calldata, and cleanup options
- reject context or plan drift and require a fresh review instead of silently rebuilding
- reserve Safe submissions before the wallet call and retain unresolved submission state for reconciliation

## Test plan

- [x] focused review and execution suite: 78 tests
- [x] `npm run typecheck`
- [x] `npm run lint` (existing unrelated warnings only)
- [ ] switch account and network with review open and confirm execution is rejected before any signature or send
- [ ] review an rEUL unlock and Fuul claim and confirm amount, destination, approval token, and native fee match the wallet request


LeonardEulerXYZ (changes requested):
Blocking: batch execution is not bound to the prepared envelope shown in review.

`BatchReviewModal.vue` prepares and renders one envelope, but `useTxBatch.ts:2347-2351` sends prerequisites, rebuilds the merged execution plan, prepares it again, and executes the second envelope without semantic comparison. A live allowance change can therefore insert an approval absent from review; migration builders can also produce a different signed plan after prerequisite grants.

Please execute the exact reviewed prepared envelope, or enforce a canonical equality policy that permits only narrowly defined volatile deltas. Any prerequisite/signature-driven plan delta needs an explicit reviewed representation and regression coverage.

Reviewed exact head `d017e80c2cde6b6b5538b14178e77d298bc0f585`; 37/37 changed files covered. Focused tests: 90 passed. Typecheck, lint, build, and diff-check passed.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `1482e93d12364f4d92f82a7fe30a9d065caa88b5`.

The reviewed-envelope plumbing is materially improved, but the canonical equality exception remains too broad. `utils/reviewed-execution.ts` classifies a call as Pyth solely by the first four calldata bytes (`updatePriceFeeds(bytes[])`). It then canonicalizes the call's entire `data` and `value`, without authenticating the target as the reviewed/configured Pyth contract. Therefore, any reviewed call to any target with that selector may change arbitrary calldata and native value between review and execution and still pass `requireReviewedBatchPreparedExecution()`.

I reproduced this at the current head with an arbitrary target: reviewed `0xef9e5e2801`, value `1`; candidate `0xef9e5e28deadbeef`, value `999`; the equality check accepted the candidate by identity.

Please bind the volatile exception to the exact reviewed Pyth call identity (at minimum authenticated target + selector + unchanged non-volatile fields), and decode/validate that only the intended update payload and its corresponding bounded fee changed. Add a negative regression for a non-Pyth target sharing the selector and for malformed/extra calldata.

Validation:

- focused reviewed-envelope, batch, migration, reward, and decoding suites: 103 passed
- typecheck, lint, production build, diff check, and clean-worktree check: passed
- fresh security/supply-chain delta pass: no unrelated finding

The branch is also draft and conflicts with current `development`; reconcile and retest after fixing this blocker.

LeonardEulerXYZ (changes requested):
Supplementary blockers from an independent full-diff challenge pass on this unchanged head:

1. Safe/WalletConnect indeterminate submission is retryable. `waitForSafeTransactionExecution()` throws `SafeTransactionStatusUnknownError` when terminal status cannot be established, but `executeBatch()` treats the generic catch as “the batch never landed,” attempts prerequisite revocation, and leaves the cart retryable. The first Safe transaction may still execute. Persist the submitted Safe hash and an account/chain-scoped pending lock before polling; do not permit retry or assume abort until reconciliation proves a terminal result.

2. `executePreparedPlan()` binds normal sends to `preparedOwner`/chain, but its typed-data callback calls `signTypedDataAsync(typedData)` without reasserting account/chain/connector or pinning `account: preparedOwner`. Identity drift during async execution can therefore route an authorization prompt through a different active connection. Assert the reviewed execution context at every signature/send boundary and explicitly pin the signer where supported.

3. The placeholder-signature exception is also broader than the intended authorization slot. `getReviewedSignaturePlaceholderCalls()` marks every flattened call in an entry flagged `calldataUsesPlaceholderSignatures`; canonicalization then masks generic zero-filled dynamic 65-byte regions rather than a decoded selector/argument path. An unrelated allowlisted call with that shape can change without failing equality. Carry exact authorization-slot metadata or decode and restrict by contract, selector, and argument index; add a negative regression for an unrelated 65-byte dynamic field.

These are additional to the selector-only Pyth exception in my existing changes-requested review. The branch is draft and conflicting with current `development`, whose later security fixes must be preserved during reconciliation.

LeonardEulerXYZ (changes requested):
Further execution-binding findings from a second independent challenge pass on this unchanged head:

1. Migration prerequisite grants are outside the reviewed canonical envelope. The modal reviews `TransactionPlanPrepared`, but `sendExecutionPrerequisites()` later rebuilds and broadcasts standalone grant transactions before the main-plan equality check. Their exact targets, calldata, values, and ordering are not bound to the reviewed artifact. A changed grant can therefore be broadcast before comparison of the main plan fails. Capture and canonically compare the finalized prerequisite envelope immediately before each broadcast, or surface those exact transactions for review.

2. Direct migration confirmation is also unbound. The migration modal receives placeholder `calldataPrepared` with `allowConfirmWithoutPlan`, then `sendMigration()` refetches authorization state, signs or broadcasts grants, rebuilds, prepares, simulates, and executes a fresh plan without comparing it to the displayed placeholder envelope. Apply the same narrowly scoped reviewed-envelope comparison to direct migration.

3. The placeholder matcher does not model the supported migration ABIs. It recognizes a dynamic `bytes` field of length 65, while Aave/MetaMorpho encode static `v,r,s` and Morpho uses a static signature tuple. A real `delegationWithSig(...,uint8,bytes32,bytes32)` probe contains no dynamic-65 marker, so placeholder-to-real replacement fails closed with “The batch changed after review.” This is not a bypass, but it makes affected signed migrations non-executable. Replace shape scanning with connector/function-specific decoding and real-ABI tests for every supported authorization type.

These supplement the existing changes-requested findings; the branch remains draft and conflicting.

LeonardEulerXYZ (approved):
Re-reviewed exact head `964014ac71c410a63ad410d1d0f7ce6838dc6d75`.

The prior execution-binding blockers are resolved:

- Pyth refresh tolerance now requires the official chain-specific target, canonical `updatePriceFeeds(bytes[])` calldata, non-empty updates, and a bounded native fee.
- Migration signature substitutions are decoded and reconstructed through the supported connector ABIs (`delegationWithSig`, ERC-2612 `permit`, and Morpho `setAuthorizationWithSig`) instead of masking arbitrary 65-byte regions.
- Standalone migration prerequisites are bound before broadcast to the reviewed account, chain, ordered targets, calldata, values, and restoration envelope; duplicate grants may only be omitted as an order-preserving subsequence.
- Direct and batched migration execution compare the rebuilt prepared plan with the reviewed envelope before execution, while typed-data and transaction sends remain pinned to the reviewed wallet context.
- Safe unknown-status handling and reconciliation remain fail closed.

Focused reviewed-execution, batch, Euler transaction, Safe, and migration-authorization tests passed; typecheck, lint, production build, diff check, current-base composition, and the complete changed-file/security pass found no blocker.

The branch inherits current `development`'s package/lock mismatch (`npm ci` reports missing `cac@6.7.14` and `crossws@0.4.10`); this PR has zero package-file delta, and the focused gates pass via `npm install --package-lock=false`. This is baseline repository debt, not a branch regression.

Code verdict: approve. The PR is still draft.

composables/useTxBatch.ts
```diff
@@ -2608,15 +2796,55 @@ export const useTxBatch = () => {
...
       logWarn('useTxBatch/executeBatch', error)
+      if (safeSubmissionStarted) {
+        // A Safe hash was returned. Any subsequent provider/polling failure is
+        // indeterminate until reconciliation; do not revoke prerequisites or
+        // leave the cart retryable while that transaction may still execute.
+        execError.value = safeReservation?.errorMessage
+          ?? 'The Safe transaction status is unresolved. Verify it before retrying.'
+        return
+      }
+      if (safeReservation) {
```
LeonardEulerXYZ: **[P1] Preserve the Safe lock when submission is ambiguous before a hash is returned.** `onSafePreflight` durably reserves this batch before the wallet request, but this catch deletes that reservation whenever `onSafeSubmission` did not run. A Safe/provider bridge can accept `wallet_sendCalls` and then lose/reject the response before the hash reaches the app; in that state the proposal may execute even though `safeSubmissionStarted` is still false. Clearing here then runs `revokeAfterAbort` and makes the cart retryable, so the original proposal and a retry can both execute (for example, doubling a borrow). The persisted type and reconciliation message already model a reservation with no hash, but this branch removes it. Keep the no-hash reservation for every non-provably-prebroadcast failure (or clear only a definitive user rejection), block retry/revocation, and add a test where `onSafePreflight` fires and the provider rejects before `onSafeSubmission`; the current code returns `pendingSafeSubmission = null` in that fixture.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `49ecbeef`. The new account/network invalidation for an open operation review is sound, and the repository gates are green. One Safe ambiguity blocker remains in the composed execution path.

Local validation:
- focused security/lifecycle suite: 183 passed
- full suite: 1,790 passed, 1 skipped
- typecheck: passed
- lint: passed (6 existing warnings)
- production build: passed
- adversarial pre-hash Safe fixture: 1 failed as expected (`pendingSafeSubmission` became `null`)

This branch existed in the earlier reviewed head as well; my previous approval missed it.

composables/useWalletReviewContext.ts
```diff
@@ -0,0 +1,43 @@
...
+  onInvalidated: (change: WalletExecutionContextChange) => void
+}) => {
+  const expectedAccount = account.value
+  const expectedChainId = chainId.value
+  const isInvalidated = ref(false)
+
+  watch(
+    [account, chainId, isSubmitting],
+    ([currentAccount, currentChainId, submitting]) => {
+      if (isInvalidated.value || submitting) return
```
LeonardEulerXYZ: **[P1] Latch wallet drift even while submission is active.** Returning here forgets the event entirely. If account A submits, then the wallet moves A → B → A before the promise settles, `useSafeExecutionDetachment` abandons the tracked execution on the first switch and the flow suppresses its post-tx modal teardown. When `isSubmitting` becomes false, the live context again equals A, so this hook leaves the same review enabled and a second confirm can submit it again. This is especially dangerous after a successful borrow/withdraw proposal. Set the invalidated latch immediately on the first drift; defer only `onInvalidated`/close until submission settles. Add account and chain round-trip tests, not only persistent-drift tests.

composables/useTxBatch.ts
```diff
@@ -2608,15 +2811,59 @@ export const useTxBatch = () => {
       await sendExecutionPrerequisites(grantedRevokes)
```
LeonardEulerXYZ: **[P1] Validate batch identity before broadcasting current prerequisites.** `BatchReviewModal` keeps the prepared envelope captured on mount, while later cart edits clear only `latchedBundledExecution` and eventually re-enable confirmation after resimulation. On this sequential path, a grant-bearing entry added after review reaches `sendExecutionPrerequisites` before the rebuilt plan is compared with the reviewed envelope below. The core batch is rejected, but unreviewed approvals/operator grants have already been prompted or broadcast; cleanup is best-effort and can fail. Capture a cart revision/fingerprint with the review and reject it before this line, and close/disable an open review on every entry mutation. A discriminating test should review entry A, commit grant-bearing entry B afterward, then assert no send boundary is touched.

composables/useEulerTx.ts
```diff
@@ -1577,17 +1615,36 @@ export const useEulerTx = () => {
...
 
-    finalizeExecution(result)
+    finalizeExecution(result, cid)
     return result
   }
 
-  const executePreparedPlan = async (prepared: TransactionPlanPrepared) => {
+  const executePreparedPlan = async (
+    prepared: TransactionPlanPrepared,
+    options?: {
```
LeonardEulerXYZ: **[P1] Direct Safe operations need an authoritative unresolved-submission record too.** These lifecycle callbacks are optional, and the direct reviewed-operation callers do not provide them. If Safe returns a hash and polling times out, the page reloads, or the connector is abandoned, only the in-memory tracked-execution slot exists; it is released/abandoned and there is no durable account/chain/hash record to reconcile or block retry. A second borrow, withdrawal, migration, reward claim, etc. can therefore be proposed while the first remains executable. Apply the same durable reservation/reconciliation invariant used by batch execution to every Safe submission, including pre-hash ambiguity.

composables/useEulerTx.ts
```diff
@@ -1641,41 +1698,111 @@ export const useEulerTx = () => {
...
+        onSafePreflight: options?.onSafePreflight,
+        onSafeSubmission: options?.onSafeSubmission,
       })
       if (bundled) {
         finalizeExecution(bundled)
         return bundled
       }
     }
 
+    if (safeWalletProvider) await options?.onSafePreflight?.()
```
LeonardEulerXYZ: **[P1] Do not degrade a reviewed Safe bundle into sequential writes.** The modal labels approvals as bundled when Safe is detected, but if `getSafeWalletProvider` is unavailable at confirmation this branch is skipped and execution falls through to `executePreparedTransactionPlan` with Permit2 disabled. For a known Safe connector that means separate approval and main-transaction proposals rather than the single atomic proposal shown in review; rejecting/failing the main proposal can leave the approval standing. Once review promises bundled Safe execution, provider loss must fail closed and require re-review/reconnect, as the batch path already does, rather than silently changing ceremony and atomicity.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `315b0a3a`. The new commit correctly fixes the previously reported pre-hash batch reservation bug; its ambiguous-response, definitive-rejection, and storage-failure tests are discriminating. Four independent transaction-integrity blockers remain in unchanged paths.

Exact-head validation: affected suites 96/96 passed; typecheck passed; diff check passed; worktree clean. The prior full suite/build/lint run was green at `49ecbeef`, and this push changes only `useTxBatch` plus its tests.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `ce02e438`. The previous four reported paths are materially improved: transient wallet drift is latched; stale prerequisites are revision-bound; direct Safe operations now retain unresolved locks; and Safe provider loss fails closed. Remaining blockers are at the next composition boundaries: cart mutation during final helper awaits, cross-tab reservation ownership, malformed-lock handling, and late Safe classification changing Permit2 review into approval execution.

Local exact-head verification: focused changed-area suites 138/138; full suite 1,811/1,811; typecheck passed; diff check clean. GitHub lint/typecheck/test/security checks are green. I also ran a reviewer-only interleaving probe for the cross-tab finding and reverted it afterward; the committed worktree is clean.

composables/useTxBatch.ts
```diff
@@ -2588,39 +2794,117 @@ export const useTxBatch = () => {
...
+          ? getAddress(prepared.account)
+          : getAddress(prepared.account.owner)
+        if (batchRevision.value !== reviewedRevision) throw new Error(BATCH_REVIEW_INVALIDATED_ERROR)
         const result = await executePreparedPlanWithPlainCalls(prepared, {
           before: collected.grants,
           after: collected.revokes,
-        }, { allowSingleCall: true })
+        }, {
+          allowSingleCall: true,
+          onSafePreflight: () => {
```
LeonardEulerXYZ: **[P1] Recheck the reviewed revision at the terminal wallet boundary.** `batchRevision` is checked at line 2804, but this helper then awaits SDK/provider/approval work before calling this preflight immediately before `sendCalls`. `addEntry`/`removeEntry` remain callable during that window, so the old reviewed proposal can still broadcast after the cart changed; success then `clearBatch()`s the newly mutated cart. The sequential path has the same gap, and EOA execution has no equivalent callback. Thread the revision assertion into every final wallet write, not only into prerequisite sends and the helper entry point.

composables/useTxBatch.ts
```diff
@@ -1745,7 +1813,62 @@ export const useTxBatch = () => {
...
+      return pendingSafeSubmissions.value.find(pending =>
+        pending.account === normalizedOwner && pending.chainId === currentChainId,
+      ) ?? null
+    }
+    catch {
+      return null
+    }
+  })
+  const setPendingSafeSubmission = (pending: PersistedPendingSafeBatchSubmission) => {
+    refreshPendingSafeSubmissions()
```
LeonardEulerXYZ: **[P1] Do not overwrite a reservation acquired by another tab.** This refresh is followed by filtering out every record for the account/chain and writing the new one unconditionally. There are awaits between `reconcilePendingSafeSubmission()` and Safe preflight; if another tab reserves an operation in that interval, this setter deletes its lock and the batch reaches the wallet. A reviewer-only test wrote an operation reservation during `restorePendingBeforeRetry`; `onSafePreflight` replaced it, the wallet request ran, and rejection removed the original lock. Reservation acquisition needs cross-tab serialized compare-and-claim ownership (and the direct-operation check/write path needs the same atomicity), rather than refresh-then-replace.

utils/pending-safe-batch-submission.ts
```diff
@@ -0,0 +1,162 @@
...
+  storage: StorageLike,
+): PersistedPendingSafeBatchSubmission[] => {
+  try {
+    const raw = storage.getItem(PENDING_SAFE_BATCH_STORAGE_KEY)
+    if (!raw) return []
+    const values = parse(raw)
+    if (!Array.isArray(values)) return []
+    return values.map(normalizeRecord).filter((value): value is PersistedPendingSafeBatchSubmission => Boolean(value))
+  }
+  catch {
```
LeonardEulerXYZ: **[P1] Fail closed when the durable Safe lock cannot be read or validated.** Parse errors, `getItem` failures, non-array payloads, and invalid records all collapse to `[]`, which every caller interprets as `no unresolved submission` and may overwrite with a new reservation. A runtime probe confirmed both malformed JSON and throwing storage return an empty list. Because this state is the duplicate-proposal guard after an ambiguous Safe response, unreadable/invalid existing state must block submission for reconciliation instead of authorizing a retry.

composables/useEulerTx.ts
```diff
@@ -1605,12 +1848,13 @@ export const useEulerTx = () => {
       ? getAddress(prepared.account)
       : getAddress(prepared.account.owner)
 
-    // Known Safe even when provider acquisition failed — the degraded
-    // sequential path must still never sign permit2 messages.
+    // Direct Safe connectors remain identifiable when provider acquisition
+    // fails; the reviewed-envelope marker covers WalletConnect Safe sessions.
     const isKnownSafe = Boolean(safeWalletProvider) || isSafeConnectorIdentity(connector)
+    const requiresSafeLifecycle = isKnownSafe || isReviewedSafeExecutionRequired(prepared)
```
LeonardEulerXYZ: **[P1] Do not silently execute a repaired approval envelope that was never reviewed.** If Safe detection resolves after the modal prepared a Permit2 envelope, the modal still displays `reviewedExecution.plan` with `Sign permit2 message`. This branch can replace it with fresh ERC-20 approval calls and then broadcast that different plan. The new test explicitly blesses the repaired object, but does not prove review/execution parity. A late Safe classification that changes resolved authorization items must invalidate/reprepare and render a new review (or prove exact canonical equivalence); it cannot mutate the consent-bearing envelope inside execution.

dglowinski:

PR #783, opened 2026-08-06
Bundle policy freshness guards

dglowinski:
## Summary

- keep TOS, VPN, geographic, and vault-authority decisions bound to the active wallet and chain throughout review and execution
- reject stale asynchronous policy results after navigation, account changes, network changes, or policy refreshes
- make every transaction confirmation react to current blockers and assert them again immediately before execution

## Issues and fixes

- **LITE-295 / EULE2-20 — Bind TOS checks to the current wallet.** An async TOS result could complete after an account or chain switch and clear the active context's blocker or signature state. Checks capture an immutable wallet context and generation, discard stale completions, and keep the blocker registered while the current result is pending or invalid.
- **LITE-298 / EULE2-31 — Preserve positive VPN screening results.** A browser-detected VPN could be lost when the server checked only request headers and a clean address result overwrote the local verdict. Client and server preserve a positive VPN signal and combine it with address screening instead of allowing either layer to clear it.
- **LITE-308 / EULE2-36 — Verify canonical governor before trust.** Discovery-list membership alone could mark a vault verified even when its live governor, admin, or Earn owner had drifted. The guard resolves the chain-specific vault type, calls the canonical authority verifier, recomputes on relevant changes, and fails closed when required authority data is missing.
- **LITE-300 / EULE2-43 — Invalidate review when geo policy changes.** A review opened before country detection completed, or before the country became blocked, could remain confirmable. Pending country state blocks mutations, open reviews reactively disable confirmation, and the active operation blocker is asserted again in the send path.

## Changes

- scope TOS signature state to account and chain and reject stale async writes
- carry positive VPN detection through the internal screening API
- use canonical EVK, Securitize, or Earn authority verification, with escrow kept as an explicit trust anchor
- register operation blockers centrally so direct, CoW, Earn, and batch flows share the same final execution gate

## Test plan

- [x] production build
- [x] full Lite suite: 169 files and 1,591 tests; one build-dependent test skipped before build
- [x] post-build logger bundle suite: 4 tests
- [x] focused policy regressions: 56 tests
- [x] typecheck and lint (six existing unrelated warnings only)
- [ ] keep a review open through account, chain, country, and vault-authority changes and confirm every stale context is blocked

## Summary by CodeRabbit

* **New Features**
  * Added clearer blocking messages and “Action required” prompts during batch and operation reviews.
  * Rechecks Terms of Use, regional, vault verification, and other policy restrictions before transactions are submitted.
  * CoW Swap actions now stop if the connected wallet or network changes after review.
  * Added geographic restriction messaging for pending, unknown, or blocked regions.
* **Bug Fixes**
  * VPN detection now responds to positive client detection and disconnects before remote screening.
  * Improved protection against stale policy and account states during transaction execution.


LeonardEulerXYZ (changes requested):
Two blocking final-write gaps remain.

1. The global batch path does not assert current operation policy. `BatchReviewModal.vue:384-397` and `useTxBatch.ts:2332-2337` gate on simulation/balance/preparation state, but not the current entry-derived policy. Page-owned guards can resolve blocked or unmount after an entry is added, while the global cart can still prepare and execute it.

2. CoW review/signing does not react to policy changes while open. `CowSwapReviewModal.vue:120-126,250-258` gates only spy/submission state; `useCowSwapExecutionCore.ts:104-125` can then request approvals, permits, and the order signature without a fresh policy assertion.

Please bind current policy to every batch entry at final execution and assert it immediately before each CoW approval/signature/write. Add regression tests for policy changes after review opens and after navigation unmounts the originating form.

Reviewed exact head `1340845b755a4e6ae397d159e20b9cb0c32a3331`; 20/20 changed files covered. Full suite: 1,595 passed, 1 skipped. Typecheck, lint, build, and diff-check passed.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `1ee187103d699ef971b2462a69b1ee452077fa9c`.

The new final-write checks close the original same-account timing gaps, including the interval after a CoW signature and before `submitOrder`. One blocking lifecycle gap remains: the callbacks retained after source-page unmount are not fully live across wallet identity changes.

`useOperationGuard()` captures callbacks whose ToS and Keyring state is maintained by component-scoped watchers. On unmount, those watchers stop (`useTosGuard` no longer resets/rechecks `sessionAccepted`/`hasSigned`; `useKeyring` marks itself disposed and no longer resets/rechecks `hasValidCredential`). The callback survives, but its identity-scoped state can remain from the old wallet. A concrete sequence is:

1. wallet A accepts ToS or has a valid Keyring credential;
2. open CoW review and navigate so the source guard unmounts;
3. switch to wallet B or another chain;
4. confirm the still-open review.

The retained callback can report no block from A's stale acceptance/credential state, and all new `assertOperationPolicyChecks()` calls then pass for B. The batch path independently clears its cart on account/chain change, but the CoW review path has no equivalent identity binding.

Please capture the reviewed wallet account + chain with the policy checks and fail closed on drift before every CoW approval/signature/submission, or move the relevant identity reset/recheck lifecycle into a scope that survives the operation. Add a regression covering account/chain change after source unmount and before CoW confirmation.

Hosted lint, typecheck, test, CodeRabbit, and preview checks are green; diff check and current-base synthetic merge are clean. Full 31/31 changed-file coverage and a fresh security/supply-chain pass found no unrelated blocker. The PR remains draft `[DON'T MERGE]`.

LeonardEulerXYZ (changes requested):
Supplementary blocker from an independent execution-boundary challenge pass on this unchanged head:

Batch policy is still checked too late for migration prerequisite writes. `sendExecutionPrerequisites()` awaits `buildExecutionPrerequisites()` and then sends `preTxs`; the fresh retained-policy assertion occurs only later, immediately before the merged batch. If policy becomes blocking while the prerequisite builder is awaiting, an authorization/approval grant is still written before policy is checked again.

A discriminating probe flipped policy to blocked inside the async prerequisite builder and returned a grant; `sendPlainTransactions()` was nevertheless called.

Assert the captured entry policy immediately before each prerequisite wallet write—not merely before the final batch—and preserve the existing revoke-on-abort behavior if policy changes after an earlier grant has landed. Add the above timing regression.

This is additional to the CoW account/chain lifecycle blocker in my existing changes-requested review.

LeonardEulerXYZ (changes requested):
Further stale-policy path from a second independent challenge pass on this unchanged head:

The retained Keyring callback can become stale through time alone after source-page unmount. `useOperationGuard()` reads `keyring.isVerificationRequired`, but `useKeyring()` cancels its credential-expiry timer on unmount. A credential valid when the batch or CoW review is created can therefore remain represented as valid indefinitely after navigation, even after `credentialData.validUntil` passes. The retained final-write assertions invoke the callback, but the state behind it no longer advances.

Evaluate expiry directly from `credentialData.validUntil` and current time inside the retained policy check, or maintain expiry in an operation-independent scope that survives unmount. Add a regression that queues/opens with a valid credential, unmounts the source guard, advances past expiry, and verifies every approval/signature/submission boundary fails closed.

This supplements the existing account/chain lifecycle and prerequisite-write findings.

LeonardEulerXYZ (approved):
Re-reviewed exact head `8ea2f453baee49b88568591e85db327fec5eb319`.

The prior stale-policy and wallet-context blockers are resolved:

- Batch prerequisite policy is reasserted after asynchronous construction and immediately before every individual wallet write, then reasserted again before terminal prepared execution.
- Retained Keyring policy evaluates `validUntil` against wall-clock time, so component unmount cannot freeze an expiring credential as valid.
- CoW execution is bound to the reviewed account and chain, with policy/context assertions before each transaction send, typed-data signature, and SDK phase transition including order submission.
- Existing abort/revoke handling remains in place when policy changes after an earlier prerequisite grant.

Focused operation-guard, batch, ToS race, and unverified-vault tests passed. Deterministic install, typecheck, lint, production build, diff check, current-base composition, hosted checks, and the complete changed-file/security pass all passed.

Code verdict: approve. The PR is still draft.

composables/useTxBatch.ts
```diff
@@ -2609,6 +2623,9 @@ export const useTxBatch = () => {
       const executionPlan = await buildMergedExecutionPlan()
       await estimateGasForPlan(executionPlan)
       const prepared = await prepareTransactionPlan(executionPlan)
+      // The final gate runs after every asynchronous prerequisite and planning
+      // step, immediately before the wallet-facing execution call.
+      assertOperationPolicyChecks(entries.value.flatMap(entry => entry.policyChecks))
```
LeonardEulerXYZ, line 2633: **Blocking — the Safe branch bypasses this final policy gate.** `latchedBundledExecution` enters the branch above, awaits SDK/plan preparation, calls `executePreparedPlanWithPlainCalls()`, and returns before reaching this assertion. A retained policy check that flips during `prepareTransactionPlan()` therefore does not stop the Safe proposal write. I reproduced this with a throwaway test: the policy flips inside the awaited prepare mock, yet `executePreparedPlanWithPlainCalls()` is called once. Apply the same retained-policy assertion after Safe preparation and immediately before its wallet-facing call (ideally with a helper-level pre-send hook for awaits inside Safe provider/connector resolution), and add that regression.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `0d1e83534335202b51707484af6ea41121122439` after resolving the merge with `development` (`b1ec117b`). The declared conflicts preserve both parents' direct modal and CoW test behavior, and the committed focused policy suite, typecheck, production build, hosted CI, security review, and preview checks are green.

One blocking cross-parent execution gap remains: the newly merged Safe bundled-batch path returns before PR #783's terminal retained-policy assertion. A policy change while the Safe proposal is being prepared can therefore still reach the wallet-facing bundled execution call.

I reproduced this with a local throwaway regression: a retained policy callback is clear when the batch starts, flips to blocked inside the awaited `prepareTransactionPlan()`, and the test expects no Safe proposal write. All 58 committed `useTxBatch` tests passed, but this added case failed because `executePreparedPlanWithPlainCalls()` was called once. The scratch test was restored afterward.

Please assert all captured entry policy checks after Safe bundle preparation and immediately before `executePreparedPlanWithPlainCalls()`. Preferably also thread a `beforeSend`/assertion hook into the Safe execution helper so later awaits inside provider/connector/approval resolution cannot reopen the interval. Add the discriminating Safe-bundle regression.

Validation on the clean exact-head worktree:
- focused committed policy/security tests: 115/115 passed
- local throwaway Safe policy-timing regression: failed as described (58 existing tests passed)
- `npm run typecheck`: passed
- `npm run build`: passed
- `git diff --check` and conflict-marker sweep: passed
- 32/32 PR files accounted for; fresh merge-interaction and security/supply-chain passes completed

Attribution: this blocker is newly exposed by composition with the Safe bundled-execution path merged from `development`; it was not present on the previously approved `8ea2f453` head.

composables/useTxBatch.ts
dglowinski: Fixed in 72bcd3a6. The Safe bundled path now captures the retained entry-policy callbacks, asserts them after plan preparation, and passes a second assertion into the Safe helper immediately before sendCalls, after provider/connector/approval preparation awaits. Added both the requested policy-flip-during-prepare regression and a helper-boundary regression. Validation: 85 focused tests, full 1,757-test suite, typecheck, changed-file ESLint, production build, and diff check all pass.

composables/useEulerTx.ts
```diff
@@ -1299,6 +1300,7 @@ export const useEulerTx = () => {
     let lastBroadcastData: Hex | undefined
     try {
       for (const [index, tx] of txs.entries()) {
+        options?.beforeSend?.(index)
```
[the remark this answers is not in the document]
dglowinski: Fixed in cce19357. sendPlainTransactions now threads its indexed policy callback into buildSendTransaction, which invokes it after any OKX post-approve delay and the wallet-context assertion, immediately before sendTransactionAsync. The regression tests prove the second check does not run during the 3-second delay, blocks the wallet write when policy changes, and does not repeat a consumed delay.

composables/useGeoBlock.ts
```diff
@@ -227,6 +227,16 @@ export const isAnyVaultBlockedByCountry = (...addresses: string[]): boolean => {
...
 
+export const GEO_POLICY_PENDING_REASON = 'Checking regional availability'
+export const GEO_POLICY_BLOCKED_REASON = 'This operation is not available in your region'
+
+export const getVaultOperationGeoBlockReason = (addresses: string[]): string | undefined => {
+  if (country.value === undefined) return GEO_POLICY_PENDING_REASON
+  if (country.value === null) return GEO_POLICY_BLOCKED_REASON
+  if (addresses.some(address => isVaultBlockedByCountry(address))) return GEO_POLICY_BLOCKED_REASON
+  return undefined
+}
```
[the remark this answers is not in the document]
dglowinski: Fixed in cce19357. getVaultOperationGeoBlockReason now applies the global sanctioned-country rule before iterating vault addresses, so an empty unresolved address list remains fail-closed. tests/composables/useGeoBlock.test.ts covers the empty-list sanctioned case.

pages/earn/[vault]/index.vue
```diff
@@ -213,6 +213,10 @@ const addToBatch = async () => {
 const send = async (execution: TrackedExecutionScope) => {
   try {
     isSubmitting.value = true
+    if (isOperationBlocked.value || isGeoBlocked.value) {
+      error(operationBlockReason.value ?? 'This operation is not available in your region')
+      return
+    }
```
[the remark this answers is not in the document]
Comment (dglowinski) [reply pages/earn/[vault]/index.vue]: Fixed in cce19357. executePlan now accepts a final beforeSend assertion and runs it immediately before both sendCalls and sendTransactionAsync. The earn supply flow captures its live policy checks before asynchronous plan work and passes the retained assertion through. Regression tests cover both ordinary Safe and direct-wallet paths.

dglowinski:
@LeonardEulerXYZ Addressed the remaining final-boundary policy findings in cce19357.

- indexed prerequisite policy checks now run after the OKX delay, immediately before each wallet write
- sanctioned countries fail closed even while the reviewed vault-address list is empty
- ordinary earn execution reasserts retained live policy immediately before both direct and Safe submissions

Validation: focused policy suites 65/65; full suite 1,762/1,762; post-build bundle check 4/4; typecheck; full lint (0 errors, 6 pre-existing warnings); production build; git diff --check.

composables/guards/useTosGuard.ts
```diff
@@ -60,47 +62,63 @@ export const useTosGuard = () => {
...
+    const isCurrentCheck = () =>
+      generation === checkGeneration.value
+      && address.value === checkedAddress
+      && chainId.value === checkedChainId
+
     if (!enableTosSignature) {
       hasSigned.value = true
       return
     }
     if (hasSigned.value === true) return
```
LeonardEulerXYZ: [P1] Key this early return to the checked account and chain. `hasSigned` is global, while the account/chain reset watchers stop when the last guard unmounts. If A is signed, the user leaves operation pages, switches to B, and then mounts a new guard, this returns on A's stale `true` without querying B. I reproduced that remount sequence: the regression expected two contract reads but observed only A's. Store the identity alongside this result, or invalidate/recheck unconditionally on mount; cover account and chain changes while no guard is mounted.

composables/cowswap/useCowSwapExecutionCore.ts
```diff
@@ -120,13 +142,25 @@ export const useCowSwapExecutionCore = () => {
...
+          return sendTransaction(transaction)
+        },
+        signTypedData: (typedData) => {
+          assertCurrentExecutionPolicy()
+          return signTypedData(typedData)
+        },
+        onProgress: (progress) => {
+          // The SDK emits before each CoW phase, including submitOrder. This
+          // closes the interval after signing but before the order is posted.
+          assertCurrentExecutionPolicy()
```
LeonardEulerXYZ: [P1] Do not let this assertion turn an accepted order into an apparent failure. The installed SDK awaits `submitCowSwapOrder()`, receives the UID, then emits `completed`. If policy changes after acceptance, this throws before `onProgress()` records that UID; the catch path resets to `idle`, so a retry may place a duplicate order. I reproduced the sequence with a terminal progress event carrying an accepted UID. Assert through the pre-submit `submitOrder` phase, but treat `completed` as observation and always retain the accepted UID/status.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `865e56573b8101fbb06b08eee6c5e1dd027e6cf9`. The prior direct-review lifetime, ordinary-batch final-send, and typed-data boundary gaps are resolved. Two lifecycle blockers remain.

1. **High — ToS acceptance can cross an unobserved wallet/network change.** `hasSigned` and `sessionAccepted` are global, but their identity-reset watchers exist only while `useTosGuard()` is mounted. If wallet A is accepted, all operation pages unmount, the user switches to wallet B or another chain, and a new guard mounts, `checkHasSigned()` returns early on A's stale `hasSigned === true`. The new flow is therefore treated as accepted without checking B. A throwaway remount regression expected a second contract read for B and failed: only A was queried.

2. **Medium — CoW can report failure after an order was already accepted.** The installed SDK posts the order, receives its UID, and only then emits `completed`. Lite asserts policy before handling that terminal event. A policy/context change in this interval throws after the irreversible POST, leaves `orderUid` unset, and resets status to `idle`; retrying can place a duplicate live order. A throwaway regression simulating an accepted UID followed by `completed` failed with `CoW operation policy changed` instead of retaining the UID.

Please key ToS acceptance/check state to account + chain (or invalidate and recheck unconditionally on each mount), and restrict CoW policy assertions to pre-side-effect phases while always recording terminal accepted-order progress. Add both discriminating regressions.

Validation on the clean exact-head checkout: focused 115/115 passed; full 185 files / 1,775 tests passed; typecheck, lint, production build, diff check, and current-`development` merge composition passed. The built desktop shell rendered normally; wallet/data execution remained environment-limited.

composables/guards/useTosGuard.ts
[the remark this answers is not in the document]
dglowinski: Fixed in ee270be5. composables/guards/useTosGuard.ts now stores the accepted chain/account context globally and synchronously invalidates hasSigned, sessionAccepted, load failure, and stale check generations whenever a guard mounts or the wallet context changes. This covers switches that happen while no guard watcher exists. tests/composables/useTosGuardRace.test.ts reproduces the unobserved A-to-B switch and verifies the new mount blocks immediately, reads B on-chain, and requires fresh acceptance. Focused policy tests pass (9/9); full suite, typecheck, lint, and production build also pass.

composables/cowswap/useCowSwapExecutionCore.ts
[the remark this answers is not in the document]
dglowinski: Fixed in ee270be5. The exact installed SDK emits submitOrder before posting the signed order, but emits completed only after executeCowSwapPlanItem returns the accepted order UID. composables/cowswap/useCowSwapExecutionCore.ts therefore keeps the policy assertion at every pre-acceptance progress boundary, including submitOrder, and treats completed as post-irreversible so it always records the UID. tests/composables/useCowSwapExecutionCore.test.ts flips policy after submitOrder and verifies completed still resolves to submitted with the accepted UID and no error, preventing a duplicate retry. Focused policy tests pass (9/9); full suite, typecheck, lint, and production build also pass.

dglowinski:
@LeonardEulerXYZ Two new Cursor Security findings on the previous head are resolved in ee270be5: (1) shared ToS acceptance is now synchronously bound to the current chain/account even when the switch occurred while no guard was mounted; (2) CoW policy remains fail-closed through submitOrder, while post-acceptance completed progress always records the accepted UID instead of surfacing a false failure that can be retried. Added focused regressions for both. Validation: focused policy tests 9/9; full suite 185 files / 1777 tests; typecheck passed; lint 0 errors (6 existing warnings); production build passed. Please re-review this exact head.

composables/guards/useTosGuard.ts
dglowinski: Fixed in current head ee270be5. The guard now stores a global chain/account acceptance context and synchronously runs syncAcceptanceContext before the hasSigned early return and before blocker registration on every mount. A context mismatch invalidates hasSigned, sessionAccepted, load failure, and the async-check generation, so an A-to-B switch while no watcher exists cannot reuse A. tests/composables/useTosGuardRace.test.ts seeds signed/accepted state for A, mounts directly under B, and verifies the immediate pending blocker plus a fresh B contract read and required result. Focused policy tests pass 9/9; full suite 1777/1777.

composables/cowswap/useCowSwapExecutionCore.ts
dglowinski: Fixed in current head ee270be5. The installed SDK emits submitOrder before submitCowSwapOrder and emits completed only after executeCowSwapPlanItem returns its accepted UID. The core still asserts policy at submitOrder and every other pre-acceptance boundary, but completed is now observation-only and always reaches onProgress so the UID/status are retained. tests/composables/useCowSwapExecutionCore.test.ts flips policy after submitOrder and verifies terminal completed resolves to submitted with the accepted UID and no error. Focused policy tests pass 9/9; full suite 1777/1777.

LeonardEulerXYZ (approved):
Re-reviewed exact head `bf99e3c64ad981059453fd7b8380d7970fe4cace`. Both prior lifecycle blockers are resolved.

- ToS acceptance is now bound to a normalized account+chain context and synchronously invalidated on every guard mount, including account or network changes made while no operation guard was mounted. The new regressions cover both unobserved account and chain switches and require a fresh contract check.
- CoW policy remains fail-closed through the pre-submit `submitOrder` phase, while terminal `completed` progress is treated as post-acceptance observation so the accepted order UID and submitted status cannot be discarded after the irreversible orderbook boundary.

Focused boundary suite: 129/129 passed. Full suite: 185 files / 1,778 tests passed. Typecheck, lint, production build, diff check, current-`development` composition, hosted checks, and exact-head desktop shell smoke all passed. No further blocker found.

dglowinski:

PR #784, opened 2026-08-06
Bundle prepared execution context

dglowinski:
## Summary

- bind cached swaps, Permit2 signatures, vault registry entries, and Safe reconciliation to the exact prepared execution context
- prevent stale account, chain, TOS, or registry state from being reused across preparation and execution
- keep ambiguous Safe submissions locked until their actual terminal result is reconciled

## Issues and fixes

- **LITE-319 / EULE2-24 — Invalidate swap plans after TOS acceptance.** A quote prepared before TOS acceptance could be reused after the blocker cleared and omit the required TOS signing call. Prepared swap validity includes the TOS capability fingerprint across swap, collateral, and multiply quote consumers. Stale quote envelopes are rebuilt before review, and tracked envelopes are rejected again at both prepared-plan execution boundaries.
- **LITE-321 / EULE2-28 — Bind Permit2 signing to the reviewed wallet.** A wallet switch could expose a Permit2 prompt to a different account even though the later send guard rejected the drift. The prepared owner and chain are asserted immediately before signing, and the explicit prepared owner is supplied to the signing request.
- **LITE-310 / EULE2-29 — Reconcile Safe execution before retry.** A submitted Safe transaction with unknown receipt status could leave the same reviewed batch retryable. Submission identity, wallet context, and cart snapshot are persisted before the wallet send; unresolved batches remain locked until reconciliation confirms execution, cancellation, or failure.
- **LITE-299 / EULE2-33 — Scope vault registry entries by chain.** Address-only caching could reinsert chain-A vault metadata after switching to chain B and feed the wrong asset into an approval or deposit plan. Registry, pending-resolution, category, and escrow state use `(chainId, normalizedAddress)` identity, ignore stale async writes, and verify current-chain vault assets at the planner boundary.

## Changes

- fingerprint prepared execution with wallet, chain, account, TOS state, and reviewed plan identity
- bind every app-prepared or lazily prepared quote envelope to its TOS generation and reject it after that generation changes
- distinguish Safe off-chain cancellation from mined-but-reverted statuses and resolve an attached execution hash through its receipt
- restore the correct cart snapshot only for terminal non-execution; confirmed execution clears it exactly once
- record successful execution even if the user changed chain while it was pending, while suppressing stale-context redirects and cart cleanup
- retain ambiguous wallet-send failures as pending submissions so a post-broadcast transport error cannot make the batch immediately retryable
- re-resolve vault objects on chain changes and reject stale registry generations

## Test plan

- [x] full Lite suite: 184 files and 1,784 tests passed
- [x] focused prepared-context suite: 80 tests passed
- [x] `npm run typecheck`
- [x] `npm run lint` (six existing warnings only)
- [x] `git diff --check`
- [ ] submit a Safe approval prerequisite, reload, reconcile it, and confirm the cart is retained without false success
- [ ] switch account or chain before Permit2 signing and confirm no stale signature can proceed


LeonardEulerXYZ (changes requested):
Two execution-context blockers remain.

1. Ambiguous Safe submission state is only module memory (`pendingSafeSubmissions`). Reload loses the submitted hash and retry lock, allowing the same withdrawal/borrow/migration to be rebuilt and signed while the first Safe transaction may already have executed or remain executable. Persist and hydrate at least account, chain, submitted hash, and batch fingerprint before enabling retries; retain the lock until terminal reconciliation.

2. Batch review is not bound to the plan built at execution. The modal displays the fixed preview, while confirmation sends prerequisites, calls optional `buildExecutionPlan()` functions, merges fresh plans, and signs them without semantic comparison. Enforce a digest over targets, operation kinds, assets, recipients, amount bounds, and postconditions, or show the final rebuilt plan before terminal signing.

Reviewed exact head `47c87bd4895936c3020ea49fe59974df2f9cde84`; 19/19 changed files covered. Full suite: 1,606 passed, 1 skipped. Typecheck, lint, build, and diff-check passed.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `50f35b64e0874401112d385aac1c6da749f4ed63`.

The reviewed-envelope blocker is resolved: the final rebuilt prepared plan is now compared against the reviewed prepared envelope, permitting only the explicit Pyth freshness and flagged signature-placeholder deltas.

One blocking Safe-lifecycle gap remains. `persistPendingSafeSubmissionLocks()` catches and suppresses `localStorage` write failures. The first durable write occurs inside `onSafeSubmission`, after a Safe hash already exists. If `setItem()` fails (quota, unavailable/private storage, or a browser storage error), execution continues with only the in-memory lock; reload then loses the hash/context and reopens the same retry path while the first Safe transaction may remain executable. This is the original ambiguous-submission invariant under a storage-failure branch.

Require usable durable storage before allowing Safe submission, and add a discriminating test where `setItem()` throws: no Safe batch may be submitted without a reload-surviving lock. Throwing only after the hash exists is too late; the preflight/reservation must precede submission.

Validation:

- focused Safe persistence, reviewed-envelope, and batch tests: 53 passed
- typecheck, lint, production build, diff check, and clean-worktree check: passed
- current-base synthetic merge: conflicts in `tests/composables/useTxBatch.test.ts`; this branch also needs reconciliation with current `development`

Full 26/26 changed-file coverage and a fresh security/supply-chain pass found no other blocker.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `a87878ba76619957b23f967453a394fd0995e02c`.

The durable pre-submission reservation fixes the reported `localStorage.setItem()` fail-open branch: storage is reserved and read back before Safe receives a transaction request. However, one Safe lifecycle blocker remains.

`executePreparedPlan()` routes every Safe `sendTransaction` hash through `onSafeSubmission` (`composables/useEulerTx.ts:1490-1503`). This includes standalone SDK approval transactions before the terminal EVC batch. `executeBatch()` then sets `batchExecutionStarted = true` and persists that hash as the batch submission in `onSafeSubmission` (`composables/useTxBatch.ts:2700-2734`), without checking the current progress phase.

Concrete failure path:

1. The SDK sends a required approval through Safe before the EVC batch.
2. Safe returns the approval submission hash, so `onSafeSubmission` records it as the reviewed batch hash.
3. Approval status polling becomes indeterminate and throws `SafeTransactionStatusUnknownError`; the terminal EVC batch has not been sent.
4. Later reconciliation sees the approval executed and `reconcilePendingSafeSubmission()` treats that as final batch success: it revokes/restores, clears the cart, refreshes, and redirects.
5. The user is told the batch completed although only its approval executed.

The test named `does not misclassify an unresolved approval Safe transaction as the atomic batch` is non-discriminating for this path: it emits `onProgress({ status: 'approval' })` and throws, but never invokes `onSafeSubmission(approvalHash)`, whereas production invokes the hash callback from the shared Safe send closure.

Required fix: bind Safe hash capture/reconciliation to the terminal EVC send only. For example, track the SDK progress phase and accept `onSafeSubmission` as the batch hash only after the terminal batch boundary is armed; approval ambiguity needs its own fail-closed lifecycle and must never enter final-batch success handling. Add a regression that emits approval progress, invokes `onSafeSubmission(approvalHash)`, throws unknown status, then reconciles that approval as executed and proves the cart is not cleared or reported successful.

The focused tests, typecheck, lint, production build, diff check, and current-base composition otherwise pass. The branch inherits current `development`'s package/lock mismatch (`npm ci` reports missing `cac@6.7.14` and `crossws@0.4.10`) with zero package-file delta, so that is baseline repository debt rather than this PR's blocker.

LeonardEulerXYZ (changes requested):
## Exact-head review

Reviewed commit `ab4b8f09ef67287751b10bab61a77deceb017878`.

### [P1] Preserve the terminal Safe ambiguity if its hash update cannot be persisted

`onSafeSubmission` is called only after Safe has returned the terminal EVC batch hash. If this `setPendingSafeSubmission(submitted)` write fails, the callback throws and the outer catch retries the same durable update. If that also fails, it returns while the previously persisted record remains active.

That previous record can be the approval/prerequisite hash: preflight persistence succeeds, approval-hash persistence succeeds, then storage degrades before the terminal-hash write. The retained lock is therefore `{ submissionKind: 'prerequisite', submittedHash: approvalHash }`, even though Safe has already returned `batchHash` for the terminal batch. When reconciliation later confirms the approval executed, the prerequisite branch clears the lock and rebuilds the batch. The terminal batch is still ambiguous, so this re-enables a duplicate retry against a potentially live or executed EVC batch.

I reproduced this against the exact head with a reviewer-only regression that allowed the preflight and approval writes, rejected the terminal and catch-path writes, then reconciled the approval as executed. The pending lock became `null` while the terminal hash remained unresolved. The probe was removed after execution; the worktree is clean.

Please transition durable state to an unreconcilable terminal-submission reservation *before* the terminal Safe send (or otherwise guarantee that a failed terminal-hash write cannot leave an earlier prerequisite record eligible for automatic clearing). Add a regression covering this exact write sequence through reconciliation.

### Validation

- Focused committed suite: **81/81 passed** across `useEulerTx`, `useTxBatch`, and pending-Safe persistence.
- The new approval-vs-batch callback routing matches the installed Euler SDK 2.0.0 event order and fixes the prior ordinary-path hash misclassification.
- Hosted lint, typecheck, and tests are green.
- The PR currently conflicts with the advanced `development` branch in four touched files; that rebase/composition gate is separate from the blocker above.

LeonardEulerXYZ (changes requested):
## Exact-head review

Reviewed merge commit `1e6b5c5161bf0da70e2e2ff50466350f83c270f1` against current `development` (`b1ec117bf14b2dbbcbc08e58d72142e6954edb50`).

The new hashless terminal reservation fixes the previously reported degraded-storage failure on the sequential `executePreparedPlan` path.

### [P1] Retain an unresolved latched Safe bundle before allowing retry

The current-base merge adds the latched one-proposal migration path at `composables/useTxBatch.ts:2780-2815`. It calls `executePreparedPlanWithPlainCalls()` without creating any `PendingSafeBatchSubmission` or passing lifecycle callbacks.

That helper submits through `executePlanAsSafeBundle()`; `composables/useEulerTx.ts:1507-1524` obtains a Safe hash from `sendCalls()` and then waits for confirmation. If confirmation times out or is interrupted, it throws `SafeTransactionStatusUnknownError` after the proposal may already exist.

Because the latched branch never sets `safeSubmissionKind`, `batchExecutionContext`, or a pending lock, the catch at `useTxBatch.ts:2933-2980` treats this as an ordinary failure. The cart remains retryable with `pendingSafeSubmission === null`.

I reproduced the complete composable sequence with a disposable regression:

1. latch a bundled migration ceremony;
2. make `executePreparedPlanWithPlainCalls` return `SafeTransactionStatusUnknownError(safeHash, 'timeout')`;
3. observe `pendingSafeSubmission === null` and `canExecuteBatch === true`;
4. execute again;
5. observe the bundle executor called a second time.

This permits duplicate Safe proposals while the first proposal's execution status is unresolved. Please route the bundled path through the same durable pre-send reservation and submitted-hash persistence/reconciliation lifecycle as the sequential Safe path. A discriminating test should assert that an unresolved bundled proposal survives reload and blocks a second submission until reconciliation proves non-execution.

### Validation

- Exact `npm ci`: passed
- Focused Safe/review cohort: 119/119 passed
- Full typecheck: passed
- Targeted ESLint: passed
- Production build: passed
- `git diff --check`: passed
- Current-base merge composition: clean
- Disposable duplicate-bundle regression: reproduced; removed afterward

LeonardEulerXYZ (changes requested):
## Exact-head review

Reviewed `e4fcf8e88abef15001cccc4ad8dafcaec8d32acc` against current `development`.

The previous duplicate-proposal blocker is fixed: `executePreparedPlanWithPlainCalls()` now carries the Safe lifecycle hooks, persists a terminal reservation before `sendCalls()`, records the returned Safe hash before polling, and retains the lock across unknown confirmation/reload.

### Blocking: a late bundled-Safe success clears the newly active wallet's cart

`composables/useTxBatch.ts:2883-2887` handles successful latched bundle execution by unconditionally clearing `latchedBundledExecution`, calling `clearBatch()`, and redirecting. This differs from the ordinary prepared-plan success path at `useTxBatch.ts:2939-2943`, which guards cart clearing and redirect with `isExecutionContextActive(batchExecutionContext)`.

The bundle can remain in `waitForSafeTransactionExecution()` while the connected account/network changes. The owner/chain watcher then resets the old cart and permits the new context to build its own cart. When the original Safe later resolves successfully, the unconditional bundle success tail mutates that new context.

I reproduced this at the exact head with a disposable regression:

1. wallet A starts the latched bundled Safe execution and receives a submitted Safe hash;
2. before confirmation resolves, switch to wallet B / chain 8453;
3. add a new cart entry for B (`entryCount === 1`);
4. resolve A's original bundled execution successfully;
5. B's cart is cleared (`entryCount === 0`) and the success redirect runs.

Please mirror the context discipline already used by the sequential path: clear the exact `submittedSafeLock` globally, but only clear the active cart/latched ceremony and redirect when `isExecutionContextActive(batchExecutionContext)` is still true. Add the account/chain drift regression for the bundled success path.

### Validation

- Focused Safe lifecycle cohort: **121/121 passed**
- Disposable cross-context bundled-success regression: **failed as described**; removed afterward
- Typecheck: passed
- Targeted ESLint: passed
- Production build: passed
- `git diff --check`: passed
- Current-base composition: clean
- Exact-head worktree restored with no tracked changes

LeonardEulerXYZ (changes requested):
## Exact-head review

Reviewed `411495e3b39b3ad087e1964d1f1727e3a30de4b1` against current `development`.

The prior cross-wallet cart-loss blocker is fixed. The new regression is discriminating: it passes at this head, fails on `e4fcf8e8...` with `entryCount` dropping from 1 to 0, preserves wallet B's exact latched ceremony, and suppresses redirect.

One lifecycle blocker remains.

### [P1] Mark the completed execution successful even after chain drift

`redirectAfterBatchExecution(scope)` owns two independent effects:

1. `scope?.markSucceeded()` — execution-scoped terminal attribution used by detached Safe completion toasts;
2. router navigation — context-owned UI behavior.

The new bundled guard skips the entire helper when the active account/chain no longer matches. The sequential branch has the same shape. Account changes explicitly abandon the tracked execution, but network-only changes do not: `useSafeExecutionDetachment` watches account/connector, not chain. Therefore this reachable sequence misreports a confirmed transaction:

1. Submit a bundled Safe batch on chain A and close/detach the review modal.
2. Keep the same account but switch to chain B while confirmation is pending.
3. Let the chain-A Safe transaction confirm successfully.
4. The exact pending lock is correctly cleared and no redirect occurs, but `scope.markSucceeded()` is never called.
5. The detached tracker resolves with `succeeded === false` and warns: **“Safe transaction did not complete — check your Safe for details”** despite successful confirmation.

A disposable focused regression passed the tracked scope into `executeBatch`, changed only `activeChainId` after Safe hash capture, resolved the successful execution, and expected one `markSucceeded()` call with no router call. Result: no router call, but `markSucceeded` was called **0** times.

Please decouple success attribution from active-context UI effects: mark this exact execution successful once the terminal batch has finalized, while keeping cart/latch teardown and navigation guarded by `isExecutionContextActive(batchExecutionContext)`. Apply the same invariant to both bundled and sequential success tails, and add a chain-only detached-completion regression.

Validation performed:

- Focused Safe/batch cohort: 122/122 passed
- New cart-preservation regression: passes at this head and fails on the prior implementation
- Disposable chain-only detached-success regression: fails (`markSucceeded`: expected 1, received 0)
- Typecheck: passed
- Targeted ESLint: passed
- Production build: passed
- Hosted lint/typecheck/test/security checks: passed
- Current-base composition and `git diff --check`: clean

LeonardEulerXYZ (changes requested):
## Additional exact-head blockers

Still reviewing `411495e3b39b3ad087e1964d1f1727e3a30de4b1`. A delayed independent lifecycle pass exposed two additional duplicate-submission paths; I reproduced the first directly against this head.

### [P1] Retain the terminal reservation for ambiguous `sendCalls` failures

`executePlanAsSafeBundle()` calls `onSafeTerminalSubmissionStart()` before `sendCalls()`, correctly creating a durable hashless lock before the irreversible wallet boundary. But if `sendCalls()` accepts the proposal and then rejects before returning a usable ID—or returns a malformed ID—the resulting generic error reaches `useTxBatch.executeBatch()`'s ordinary catch branch, which calls `clearPendingSafeSubmission(submittedSafeLock)`.

That destroys the reservation precisely when execution status is unknowable and permits a duplicate proposal on retry.

A disposable regression invoked `onSafePreflight`, invoked `onSafeTerminalSubmissionStart`, then threw `Error('provider disconnected before response')` to model acknowledgement loss after acceptance. Expected: hashless pending lock retained and retry blocked. Actual: `pendingSafeSubmission === null`; the batch becomes retryable.

Once terminal submission is armed, absence of a returned hash is not evidence that the wallet rejected the proposal. Preserve the hashless lock for ambiguous post-arming failures; clear it only on a positively identified pre-submission rejection/cancellation. Add cases for provider rejection after terminal arming and malformed/missing bundle IDs.

### [P1] Give ordinary Safe bundle callers a durable lifecycle

The new general Safe bundling path is not limited to `useTxBatch`. `executePlan()` calls `executePlanAsSafeBundle()` with no `submissionLifecycle`, and many ordinary production callers invoke `executePreparedPlan()` without lifecycle callbacks (including lend, multiply, collateral, and swap flows).

Those flows can call `wallet_sendCalls` without any durable pre-send reservation or immediate Safe-hash persistence. Their detached tracker is in-memory only; it does not survive reload and cannot reconcile an accepted-but-unacknowledged Safe proposal. A provider disconnect/reload at the same boundary can therefore lead the user to retry and create a duplicate proposal outside the cart flow.

The lifecycle invariant must cover every production path that opts into `executePlanAsSafeBundle`, not only batch callers. Either make durable reservation/hash/reconciliation intrinsic to the general Safe bundle executor, or require and enforce a lifecycle implementation at every caller before bundling.

These are additional to the already-posted chain-drift success-attribution blocker.

composables/useSwapQuotesParallel.ts
```diff
@@ -368,6 +371,7 @@ export const useSwapQuotesParallel = (options: SwapQuotesParallelOptions) => {
       gasCostUsd,
       plan,
       preparedPlan: prepared,
+      tosContextVersion,
```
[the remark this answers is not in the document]
dglowinski: Fixed in 5a9c2bfb. Prepared envelopes are now bound to the Lite TOS context version and both execution entry points reject tracked envelopes after that context changes. Cached quote reuse in swap, collateral, and multiply flows now shares the same freshness guard, so those flows re-prepare and include the current TOS plugin data. Regression coverage is in tests/utils/sdk-tos.test.ts, tests/composables/useEulerTx.test.ts, and tests/composables/useSwapQuotesParallel.test.ts. Validated with 80 focused tests, the full 1,784-test suite, typecheck, changed-file ESLint, and a production build.

composables/useTxBatch.ts
```diff
@@ -2588,39 +2853,252 @@ export const useTxBatch = () => {
...
+          entries.value = [...batchEntriesSnapshot]
+          execError.value = errorMessage
+        }
+      }
+      else {
+        clearPendingSafeSubmission(submittedSafeLock ?? null)
+        // The batch never landed, so no granted authorization should be left
+        // standing. Entries stay in the cart for a retry, which re-grants.
+        await revokeAfterAbort(grantedRevokes)
+        execError.value = await describeExecError(error)
```
LeonardEulerXYZ: A terminal failure is still handled in whichever wallet context is active when it settles. Reproducer: submit a Safe batch under wallet/chain A, switch to B and build/latch B’s cart, then let A’s Safe proposal revert. This generic branch clears A’s persisted reservation and snapshot, unwinds A, and unconditionally writes `Transaction simulation failed: Safe execution reverted` into B’s `execError`. B’s cart survives, but switching back to A cannot restore A’s cart/error because its durable recovery record was deleted. Terminal outcomes need the same execution-context ownership and retained-state handling already used for unresolved statuses, success, reconciliation, cart cleanup, and redirects.

composables/useTxBatch.ts
```diff
@@ -2588,39 +2853,252 @@ export const useTxBatch = () => {
...
+        // The user may have switched away while Safe execution was pending.
+        // Only surface this context's error immediately; the snapshot and error
+        // are restored together when its owner/chain becomes active again.
+        if (isPendingSafeSubmissionForContext(pending, owner.value, chainId.value)) {
+          entries.value = [...batchEntriesSnapshot]
+          execError.value = errorMessage
+        }
+      }
+      else {
+        clearPendingSafeSubmission(submittedSafeLock ?? null)
```
LeonardEulerXYZ: This generic branch clears an already-armed terminal Safe reservation. In the sequential Safe fallback, `onSafeTerminalSubmissionStart` persists the hashless lock before `sendTransactionAsync`; if the provider accepts the request but the response/transport fails before the Safe hash reaches `onSafeSubmission`, the error remains generic and lands here. The lock is removed and `canExecuteBatch` becomes true, so the same proposal can be submitted again. After terminal arming, only an explicit wallet rejection may clear the lock; all other pre-hash failures must retain it as `SafeSubmissionStatusUnknownError` (or equivalent).

LeonardEulerXYZ (changes requested):
[SERANGED-TEST] Exact-head re-review of `5a9c2bfbad32dc650c279e6a257e476d60cf6394`.

The newer commits fix detached success attribution, intrinsic Safe bundle durability, ambiguous `sendCalls` responses, and TOS-generation invalidation. Two terminal-lifecycle complements remain reproducibly unsafe.

Validation: 168 focused tests passed; typecheck, scoped lint, production build, `git diff --check`, and merge-tree passed. Both findings below were reproduced with failing reviewer tests on this exact SHA and removed afterward.

composables/useEulerTx.ts
```diff
@@ -1404,13 +1446,141 @@ export const useEulerTx = () => {
...
+  }: {
+    storage: Storage
+    account: Address
+    chainId: number
+    safeWalletProvider: WalletProviderLike
+    provider: ReceiptClientLike
+  }) => {
+    const pending = findPendingSafeBundleSubmission(storage, account, chainId)
+    if (!pending) return
+    if (!pending.submittedHash) throw new SafeSubmissionStatusUnknownError(pending.errorMessage)
```
LeonardEulerXYZ, line 1498: Hashless intrinsic Safe locks have no usable recovery path after reload. This branch correctly refuses to retry when the persisted ordinary-bundle record has no trustworthy hash, but the only code that can clear that record is the execution-local `intrinsicLifecycle.clear()` closure; after reload that closure no longer exists, and no UI/composable API exposes the record for manual verification and clearing. Even after the user confirms in Safe that no proposal was created, every future ordinary `executePreparedPlan` bundle for this account/chain throws here forever. Preserve fail-closed behavior, but add an explicit recovery surface that shows the affected account/chain and requires confirmation before clearing the hashless record.

composables/useTxBatch.ts
dglowinski: Fixed in e432bba0. Once the terminal Safe fallback is armed, a pre-hash failure now clears only for an explicit wallet rejection; transport/provider failures become SafeSubmissionStatusUnknownError, preserving the durable fail-closed reservation. The useEulerTx regression tests cover both the ambiguous transport failure and explicit code-4001 rejection.

composables/useEulerTx.ts
dglowinski: Fixed in e432bba0. Wallet Settings now exposes persisted hashless Safe locks with the full account and chain, requires an explicit confirmation that no proposal exists in Safe, and only then enables the destructive clear. The storage helper independently refuses to clear any record that has a submitted hash. The new composable and storage regression tests cover both guards. Focused tests, full suite, typecheck, changed-file lint, and production build all pass.

composables/useTxBatch.ts
dglowinski: Fixed in e432bba0. A terminal Safe failure is now retained under the account/chain context that started it instead of clearing that context or writing into the currently active one. Switching back restores the original cart snapshot and its terminal error. The useTxBatch regression test “does not surface an earlier Safe failure in the active wallet context” exercises A → B → A.

dglowinski:
Addressed the cross-tab Safe retry-lock finding in `b7ffa6a0`.

- serialized intrinsic and batch Safe reservation lifecycles with a browser-wide Web Lock
- refresh durable batch reservations only after acquiring the mutex, preserving other account/chain contexts
- reconcile and manually clear hashless reservations under the same lock
- fail closed when cross-tab coordination is unavailable or another tab owns the lifecycle
- added regressions for concurrent submission, stale-tab storage, context preservation, and lock unavailability

Validation:
- focused Safe lifecycle/recovery cohort: 122/122
- remaining full suite: 1,808/1,808; the load-sensitive CoW Safe-backstop file passes 2/2 separately
- typecheck passed
- lint passed with 0 errors (6 pre-existing warnings)
- production build and `git diff --check` passed

@LeonardEulerXYZ ready for exact-head re-review.

utils/pending-safe-batch-submission.ts
```diff
@@ -0,0 +1,146 @@
...
+  try {
+    const raw = storage.getItem(PENDING_SAFE_BATCH_STORAGE_KEY)
+    if (!raw) return []
+    const values = parse(raw)
+    if (!Array.isArray(values)) return []
+    return values.map(normalizeRecord).filter((value): value is PersistedPendingSafeBatchSubmission => Boolean(value))
+  }
+  catch {
+    return []
+  }
```
LeonardEulerXYZ: **[P1] Fail closed when the durable batch registry is unreadable.** This loader maps a throwing `getItem`, malformed JSON, a non-array value, or invalid individual records to an empty/partially filtered registry. `syncPendingSafeSubmissionLocksFromStorage()` then trusts that result, and the next reservation persists over the existing unresolved lock. A transient read failure or one corrupt record can therefore make an ambiguous Safe proposal retryable. Distinguish an absent key from unreadable/invalid storage, reject the entire persisted set when any record is invalid, and do not write a replacement reservation after a failed read. A probe with an existing lock and a first-read failure reached the wallet boundary instead of blocking.

composables/useEulerTx.ts
```diff
@@ -1404,13 +1447,158 @@ export const useEulerTx = () => {
...
+    safeWalletProvider,
+    provider,
+  }: {
+    storage: Storage
+    account: Address
+    chainId: number
+    safeWalletProvider: WalletProviderLike
+    provider: ReceiptClientLike
+  }) => {
+    const pending = findPendingSafeBundleSubmission(storage, account, chainId)
```
LeonardEulerXYZ: **[P1] Interlock ordinary and batch Safe reservations.** This intrinsic path reads only `pending-safe-bundle-submissions:v1`, while `useTxBatch` reads only `pending-safe-batch-submissions:v1`. The browser mutex serializes the live storage mutation but is released after an ambiguous outcome, so it does not make the two durable registries mutually exclusive. An unresolved batch for account/chain A can be followed by the equivalent ordinary execution (and vice versa). A discriminating probe preseeded a batch reservation, called `executePreparedPlan()` for the same context, and observed one `sendCalls` instead of rejection. Use one authoritative context/fingerprint registry, or check both registries under the mutex before either path submits.

composables/useEulerTx.ts
```diff
@@ -1404,13 +1447,158 @@ export const useEulerTx = () => {
...
+    chainId: number
+    safeWalletProvider: WalletProviderLike
+    provider: ReceiptClientLike
+  }) => {
+    let storage: Storage | undefined
+    let reservation: PendingSafeBundleSubmission | undefined
+    let releaseCrossTabLock: (() => void) | undefined
+
+    const clear = () => {
+      if (storage && reservation) clearPendingSafeBundleSubmission(storage, reservation.reservationId)
```
LeonardEulerXYZ: **[P2] Always release the browser-wide lock if durable cleanup throws.** `clearPendingSafeBundleSubmission()` can throw from `getItem`, `setItem`, or `removeItem`; when it does, control never reaches `releaseCrossTabLock`. One cleanup/storage failure then holds the Web Lock for the lifetime of this page and blocks all Safe submission/recovery across tabs until reload. Release in `finally`; retaining the durable record for later reconciliation is fine, retaining the live mutex is not. The focused probe reproduced the leaked lock after a successful Safe confirmation plus a throwing `removeItem()`.

utils/reviewed-batch-execution.ts
```diff
@@ -0,0 +1,149 @@
...
+].join(':')
+
+const canonicalizePythUpdate = (item: Record<string, unknown>): Record<string, unknown> => {
+  const data = typeof item.data === 'string' ? item.data : ''
+  if (data.slice(0, 10).toLowerCase() !== PYTH_UPDATE_SELECTOR) return item
+  return { ...item, data: '__fresh_pyth_update__', value: '__fresh_pyth_fee__' }
+}
+
+const canonicalizeReviewedSignatures = (data: string): string => {
+  let result = data.toLowerCase()
```
LeonardEulerXYZ: **[P1] Canonicalize the signature encodings the migration flows actually use.** This searches only for a dynamic `bytes` length word `0x41` followed by 65 zero bytes. The migration calls marked as placeholder-bearing are Aave `permit` / `delegationWithSig` and Morpho `setAuthorizationWithSig`, whose ABIs encode static `(uint8 v, bytes32 r, bytes32 s)` fields. Their final signed calldata is therefore never canonicalized and `requireReviewedBatchPreparedExecution()` rejects it as a changed batch. The committed test uses synthetic dynamic bytes; a probe with real `permit(address,address,uint256,uint256,uint8,bytes32,bytes32)` calldata reproduced the rejection. Add selector/ABI-aware `(v,r,s)` canonicalization and real migration fixtures.

composables/borrow/useMultiplyForm.ts
```diff
@@ -1275,10 +1276,10 @@ export const useMultiplyForm = (options: UseMultiplyFormOptions) => {
             logWarn('multiply/review/preloadSubAccountSnapshot', e)
           }
         }
-        if (matchingCard?.preparedPlan) {
+        if (cachedPrepared) {
           // Lazy-prepared envelope is available — short-circuit prepare entirely.
-          multiplyPlan.value = matchingCard.plan ?? null
-          preparedMultiplyPlan.value = matchingCard.preparedPlan as TransactionPlanPrepared
+          multiplyPlan.value = cachedPrepared.plan
+          preparedMultiplyPlan.value = cachedPrepared.prepared as TransactionPlanPrepared
```
LeonardEulerXYZ: **[P1] Execute the envelope shown in the review.** This stores and displays/simulates `preparedMultiplyPlan`, but `sendMultiply()` later ignores it and calls `executePlan(multiplyPlan.value)`, rerunning every plugin and approval resolution without the reviewed-plan identity check. That is broader than the stated Pyth-freshness exception: TOS calls, approval/Permit2 mode, and other material calls can differ from the review. It also reaches the raw `executePlan` Permit2 callback, which signs without an explicit prepared owner or a current account/chain assertion, leaving the LITE-321 wallet-switch prompt defect in Multiply. Execute the reviewed prepared envelope and apply the existing narrowly scoped Pyth canonicalization/freshening rather than rebuilding unrestrictedly.

composables/useEulerTx.ts
```diff
@@ -528,10 +552,27 @@ export const useEulerTx = () => {
   const planDeposit = async (input: PlanDepositInput): Promise<TransactionPlan> => {
     const owner = requireOwner()
     const { sdk, account } = await freshPlanContext(input.account)
+    const vaultAddress = getAddress(input.vaultAddress)
+    const assetAddress = getAddress(input.assetAddress)
+    const fetchedVault = await profAsync('sdk', 'fetchDepositVault', () => sdk.vaultMetaService.fetchVault(
```
LeonardEulerXYZ: **[P1] Apply the vault/asset verification to every planner that couples approval token and deposit vault.** The new check protects standalone `planDeposit`, but wallet-collateral `planBorrow` still accepts independent caller-provided `{ vault, asset }` and forwards them without fetching current-chain vault metadata; the multiply planners expose the same shape. The SDK then uses `collateral.asset` for approval and `collateral.vault` for deposit. A stale cross-chain object can approve token A while the vault pulls token B, producing an internally inconsistent reviewed plan and an unnecessary authorization before failure. The sibling probe confirmed `planBorrow` never called `vaultMetaService.fetchVault` and did not reject the mismatch. Extract this check into a shared resolver and use it in borrow/multiply wallet-collateral planning too.

LeonardEulerXYZ (changes requested):
## Exact-head review

Reviewed `b7ffa6a0426e1db19d198a582a1c1bd15c16f878` against base `3c73cf214605e0358a997b2c54e271456648307d`.

The cross-tab refresh fix closes the prior stale-tab overwrite finding, and the chain-scoped registry keying/generation guards themselves are sound. Six blockers remain: five transaction-integrity/retry-safety defects and one browser-lock liveness defect. See inline findings.

Validation:
- exact-head hosted CI: lint, typecheck, and test succeeded
- Railway PR build succeeded
- focused exact-head Safe lifecycle/recovery cohort passed before adversarial probes; three new discriminating probes reproduced the Safe registry/read/cleanup failures
- focused exact-head vault/transaction cohorts passed; adversarial category and planner probes exposed uncovered sibling paths
- real ABI-encoded Aave `(v,r,s)` signature probe reproduced reviewed-plan rejection
- full 39-file intent, impact-graph, security, and fresh-challenge passes completed

Browser/wallet smoke: not run. These findings are at deterministic planning, persistence, and signing boundaries; no funded Safe submission was attempted.

Two additional races found during the challenge pass (`addEntry()` chain drift and category-cache reset generation) predate this PR and are intentionally excluded from this PR verdict.

composables/useEulerTx.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. Both paths now check the other durable registry under the shared Web Lock before reserving or opening Safe: ordinary execution checks the batch registry at composables/useEulerTx.ts:1607-1628, and batch execution checks the ordinary registry at composables/useTxBatch.ts:2838-2875. Bidirectional regressions prove the wallet boundary is not reached and the existing reservation remains intact (tests/composables/useEulerTx.test.ts:734-753; tests/composables/useTxBatch.test.ts:1464-1503).

utils/pending-safe-batch-submission.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. loadPendingSafeBatchSubmissions now distinguishes an absent key from getItem, parse, non-array, and invalid-record failures and rejects the entire invalid set (utils/pending-safe-batch-submission.ts:106-140). syncPendingSafeSubmissionLocksFromStorage propagates the read failure before any replacement write, with a sticky error preventing later overwrite (composables/useTxBatch.ts:314-421). Regressions cover malformed/throwing storage and an existing lock whose first read fails before the wallet boundary (tests/utils/pending-safe-batch-submission.test.ts:92-114; tests/composables/useTxBatch.test.ts:1505-1551).

composables/useEulerTx.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. Intrinsic Safe cleanup now releases the browser-wide lock in finally even when durable clear throws (composables/useEulerTx.ts:1590-1597). The regression forces removeItem to fail after confirmation, then proves a second execution acquires the lock and reaches reconciliation instead of hanging (tests/composables/useEulerTx.test.ts:755-780).

utils/reviewed-batch-execution.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. Reviewed execution now recognizes the exact Aave permit, Aave delegationWithSig, and Morpho setAuthorizationWithSig selectors and masks only their final static v,r,s words when the reviewed values are zero placeholders (utils/reviewed-batch-execution.ts:11-75). Real encodeFunctionData fixtures cover all three signatures, while a changed authorization amount still fails closed (tests/utils/reviewed-batch-execution.test.ts:45-100).

composables/useEulerTx.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. The shared resolver fetches current-chain vault metadata and requires the requested vault and asset to match (composables/useEulerTx.ts:553-578). It is now applied to wallet collateral in planBorrow (lines 632-659), planMultiplyWithSwap (840-867), and planMultiplySameAsset (870-897), while share-source collateral intentionally skips wallet-token approval validation. Focused regressions make the SDK return a mismatched asset and prove all three planners reject before their SDK planning call (tests/composables/useEulerTx.test.ts:265-328).

composables/borrow/useMultiplyForm.ts
dglowinski: Fixed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec. Multiply now re-prepares only to refresh time-sensitive plugin data, compares that candidate with the displayed prepared envelope via requireReviewedBatchPreparedExecution, and sends it through executePreparedPlan (composables/borrow/useMultiplyForm.ts:1343-1353). That preserves the reviewed owner and uses the prepared execution wallet-context assertions. Regressions prove a matching refresh executes and a material envelope delta executes nothing (tests/composables/useMultiplyForm.test.ts:482-550).

dglowinski:
@LeonardEulerXYZ All six exact-head findings are addressed in 57e94fc3c5600487258a3d0d6239c0e0ccac00ec, with evidence posted in each inline thread. Validation: 150 focused tests passed; full suite 1,827/1,827 passed; typecheck passed; production build passed; lint has 0 errors and the existing 6 no-explicit-any warnings. Hosted CI lint/typecheck/test and Railway build are also green. Please re-review this exact head.

utils/reviewed-batch-execution.ts
```diff
@@ -0,0 +1,172 @@
...
+  const data = typeof item.data === 'string' ? item.data : ''
+  if (data.slice(0, 10).toLowerCase() !== PYTH_UPDATE_SELECTOR) return item
+  return { ...item, data: '__fresh_pyth_update__', value: '__fresh_pyth_fee__' }
+}
+
+const getStaticPlaceholderSignatureStart = (data: string): number | undefined => {
+  const wordCount = STATIC_SIGNATURE_WORD_COUNTS.get(data.slice(0, 10))
+  if (!wordCount || data.length !== 10 + wordCount * 64) return undefined
+  const signatureStart = data.length - 3 * 64
+  return /^0+$/.test(data.slice(signatureStart)) ? signatureStart : undefined
```
LeonardEulerXYZ: [P1] Match the real SDK placeholder and wildcard only the ABI signature slot. `splitPermitSignature()` normalizes the 65-zero placeholder's `v` from `0` to `27`, so actual Aave/Morpho reviewed calldata ends in `(27,0,0)` and this all-zero test rejects the legitimate signed candidate. In the other direction, the dynamic scan below treats any `0x41` length word followed by 65 zero bytes anywhere in a flagged call as the signature, so an unrelated bytes argument can change without invalidating review. Decode the supported selectors and canonicalize only their exact signature argument; add real SDK-encoded `(27,0,0)` cases plus a sibling 65-byte argument that must remain immutable. The standalone probes currently reject the former and accept the latter.

utils/reviewed-batch-execution.ts
```diff
@@ -0,0 +1,172 @@
...
+const signatureCallKey = (call: ReviewedSignaturePlaceholderCall): string => [
+  call.targetContract.toLowerCase(),
+  call.onBehalfOfAccount.toLowerCase(),
+  call.value.toString(),
+  call.data.toLowerCase(),
+].join(':')
+
+const canonicalizePythUpdate = (item: Record<string, unknown>): Record<string, unknown> => {
+  const data = typeof item.data === 'string' ? item.data : ''
+  if (data.slice(0, 10).toLowerCase() !== PYTH_UPDATE_SELECTOR) return item
```
LeonardEulerXYZ: [P1] Scope the Pyth freshness exception to an authenticated Pyth call. This selector-only check replaces the entire calldata and ETH value for every target with the same four-byte selector; a non-Pyth target (or selector collision) can therefore mutate materially between review and execution. Require the target to be the configured Pyth contract for this chain and bind the exception to the exact reviewed call/argument slot. A standalone candidate changed both payload length and `value` on an arbitrary target and was accepted.

composables/borrow/useMultiplyForm.ts
```diff
@@ -1339,13 +1341,15 @@ export const useMultiplyForm = (options: UseMultiplyFormOptions) => {
...
-    if (!multiplyPlan.value) return
+    if (!multiplyPlan.value || !preparedMultiplyPlan.value) return
     isMultiplySubmitting.value = true
     try {
-      await executePlan(multiplyPlan.value)
+      // Refresh time-sensitive plugin data (notably Pyth), then require the
+      // final prepared envelope to match what the modal reviewed apart from
+      // that explicitly canonicalized freshness delta.
+      const candidatePrepared = await prepareTransactionPlan(multiplyPlan.value, { account: planAccount.value })
+      const prepared = requireReviewedBatchPreparedExecution(preparedMultiplyPlan.value, candidatePrepared)
```
LeonardEulerXYZ: [P1] Bind confirmation to this modal's reviewed envelope. `submitMultiply()` writes the reviewed prepared plan and raw plan into shared refs, while every modal callback invokes `sendMultiply()` with no captured session data. Opening review B overwrites those refs; confirming still-open modal A then prepares B and compares B against B, so the comparator passes and plan B executes even though the user confirmed A. Capture `{ reviewedPrepared, rawPlan, account/context }` in this modal's closure (or use an immutable review-session token) and reject superseded sessions. The disposable overlap regression observed modal A execute plan B.

composables/useEulerTx.ts
```diff
@@ -1641,41 +1956,152 @@ export const useEulerTx = () => {
...
+          onSafeSubmission: options?.onSafeSubmission,
+        },
       })
       if (bundled) {
         finalizeExecution(bundled)
         return bundled
       }
     }
 
+    if (safeWalletProvider) await options?.onSafePreflight?.()
```
LeonardEulerXYZ: [P1] Apply the durable Safe lifecycle before entering the sequential fallback. `executePlanAsSafeBundle()` returns `undefined` before creating its intrinsic lifecycle for single-call/non-bundleable plans; ordinary callers also provide no callbacks here, so this path neither reads the pending batch registry nor reserves durable ordinary state before the first Safe proposal. The complement regression seeded a same-context pending batch and executed a one-call prepared plan: it resolved and entered `executePreparedTransactionPlan` instead of rejecting. Either fail closed when a known Safe cannot use the atomic bundle path, or make one intrinsic durable lifecycle cover the entire sequential proposal sequence (including every prerequisite/terminal hash).

composables/useTxBatch.ts
```diff
@@ -2237,6 +2470,7 @@ export const useTxBatch = () => {
   }
 
   const addEntry = async (entry: BatchEntryInput) => {
+    if (pendingSafeSubmission.value) return
```
LeonardEulerXYZ: [P1] Do not publish an add that completed after its wallet context was reset. `buildPlan()` and slot-hint priming await external work, while the owner/chain watcher clears the cart and advances `resimToken`; this continuation never checks that epoch/context and appends the old-account/old-chain plan into the new cart. Capture an add generation plus normalized owner/chain before the first await and revalidate after every await before mutating shared snapshots, hints, metadata, or `entries`. A focused chain-switch regression expects an empty cart after reset but gets `entryCount === 1`.

utils/pending-safe-bundle-submission.ts
```diff
@@ -0,0 +1,139 @@
...
+  submission: Pick<PendingSafeBundleSubmission, 'account' | 'chainId'>,
+  account: Address,
+  chainId: number,
+) => submission.account.toLowerCase() === account.toLowerCase() && submission.chainId === chainId
+
+export const loadPendingSafeBundleSubmissions = (
+  storage: StorageLike,
+): PendingSafeBundleSubmission[] => {
+  const raw = storage.getItem(PENDING_SAFE_BUNDLE_STORAGE_KEY)
+  if (!raw) return []
```
LeonardEulerXYZ: [P2] Distinguish an absent key from a present empty/corrupt registry. `getItem()` returns `null` for absence, but `''` is an existing malformed value; treating both as `[]` lets the new batch-side interlock proceed when ordinary Safe state is ambiguous. The new batch loader correctly uses `raw === null` and rejects empty content. Apply the same fail-closed rule here and add an empty-string regression.

LeonardEulerXYZ (changes requested):
## Exact-head fixes review

Reviewed `57e94fc3c5600487258a3d0d6239c0e0ccac00ec` against both the prior reviewed head `b7ffa6a0426e1db19d198a582a1c1bd15c16f878` and merge base `3c73cf214605e0358a997b2c54e271456648307d`.

The pending-batch parser, bundled Safe interlock, lock-finally cleanup, and wallet-collateral vault/asset checks are fixed. The prepared-envelope comparison is directionally improved, but real migration calldata and overlapping Multiply reviews still violate the intended identity boundary. A fresh whole-head challenge also found two lifecycle races and one storage complement gap.

Blocking results:

1. Real Aave/Morpho static placeholders encode `(v=27,r=0,s=0)` after SDK normalization, but the comparator recognizes only three all-zero words, so legitimate signed migration execution is still rejected. The adjacent dynamic scanner also wildcards any 65-zero-byte region rather than the ABI signature argument.
2. The Pyth exception is selector-only, allowing arbitrary calldata and value changes for a non-Pyth target with the same selector.
3. Multiply modal A can execute plan B because both modal callbacks read shared mutable prepared/raw-plan refs overwritten by the latest review.
4. Single-call or otherwise non-bundleable ordinary Safe execution falls through without the intrinsic durable lifecycle and bypasses a same-context pending batch reservation.
5. An asynchronous batch add can publish an old-account/old-chain plan after the context-reset watcher has cleared the cart.
6. The ordinary Safe registry still treats a present empty value as absence, weakening the new cross-registry fail-closed invariant.

Validation:

- Focused current-head cohort: 150/150 passed.
- Full local suite: 1,823 passed, 1 skipped; ESLint and Nuxt typecheck passed.
- Hosted lint, typecheck, test, security, and Railway build checks passed.
- Independent comparator probes reproduced all three mismatches: real static placeholder rejected; unrelated dynamic bytes accepted; selector-only Pyth mutation accepted.
- Disposable overlap and context-drift probes reproduced Multiply plan-B execution and stale batch publication.
- Disposable Safe complement regression failed with the single-call path resolving despite a pending batch reservation.
- Local production build was not used as evidence because the shared dependency cache resolves Tailwind config relative to its seed worktree; the exact-head hosted build is green.

Requesting changes on the six current-head comments below.

composables/useTxBatch.ts
```diff
@@ -2556,21 +2881,123 @@ export const useTxBatch = () => {
...
+    }
     try {
       if (!await restorePendingBeforeRetry()) return
       // Final on-chain gas estimate before asking the user to sign. If the batch
       // would revert (against the current chain state, which may have moved since
       // the last simulation), surface the decoded reason and don't send.
-      const shouldRefreshExternalMigrationPositions = entries.value.some(entry => entry.refreshExternalMigrationPositions)
+      shouldRefreshExternalMigrationPositions = entries.value.some(entry => entry.refreshExternalMigrationPositions)
 
       if (latchedBundledExecution.value) {
```
LeonardEulerXYZ: [P1] Bind the latched Safe ceremony to the review/cart generation that produced it. `prepareBundledExecution()` awaits `collectBundledExecution()` and then writes the module-global latch without checking whether `clearBatch()`, an account/chain reset, or a newer review superseded it. I reproduced: start review A with a deferred `buildBundledExecution`, clear and build cart B, prepare B (latch = B), then resolve A; A overwrites the latch and this branch consumes A's grants/revokes/plans for B. The prepared-plan comparator does not authenticate the external `before`/`after` calls, so stale grants/revokes sit outside the identity check. Capture the batch context/generation plus entry identity before awaiting and refuse publication if it changed (or make the returned ceremony review-instance-owned and pass it explicitly to execution). Please keep the A→clear/B→A-resolves-last regression test.

utils/reviewed-batch-execution.ts
```diff
@@ -0,0 +1,242 @@
+import { isEVCBatchOperation, type EVCBatchItem, type TransactionPlan, type TransactionPlanPrepared } from '@eulerxyz/euler-v2-sdk'
+import { decodeFunctionData, encodeFunctionData, isHex, parseAbi, zeroHash, type Address, type Hex } from 'viem'
+import { PYTH_ABI } from '~/abis/pyth'
+
+export const REVIEWED_BATCH_EXECUTION_CHANGED_ERROR
+  = 'The batch changed after review. Close this review and review the final transaction again.'
+
+const OFFICIAL_PYTH_ADDRESSES_BY_CHAIN_ID = new Map<number, Address>([
```
LeonardEulerXYZ: [P1] Do not freeze the Pyth execution gate to this incomplete legacy-address table. Pyth's current official EVM table includes Base (`8453`, `0x8250...B1487a`), which is absent here, and publishes upgraded addresses for Ethereum, Arbitrum, BNB, Linea, Monad, and other supported chains effective 2026-08-18; those recommended targets are also absent. As soon as an Euler route points at one of those official contracts, both reviewed/candidate payloads remain uncanonicalized and every legitimate fresh Pyth update fails the reviewed-envelope comparison, blocking the batch/position action. Use a maintained source of trusted Pyth targets that supports multiple official addresses per chain (current + upgraded during migration), and add at least Base plus current/upgraded-address acceptance and unknown-target rejection tests. Source: https://docs.pyth.network/price-feeds/contract-addresses/evm

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `6d2f798a5059a8b3e3422e62bbfbc8a6a436661d`. The six prior findings are fixed, and local/hosted validation is green. Fresh whole-head challenge found two remaining execution-liveness/integrity blockers: an abandoned async Safe ceremony can overwrite the active cart's global latch, and the static Pyth trust table excludes official current/upgraded targets. Details inline.

Validation: focused fixes 155/155; full Vitest 1,835 passed / 1 skipped; typecheck clean; full lint 0 errors (6 existing warnings); Nuxt production build passed; GitHub lint/test/typecheck/security and Railway preview build all passed.

tests/composables/useMultiplyForm.test.ts
```diff
@@ -475,4 +482,120 @@ describe('useMultiplyForm cap validation', () => {
...
+    const scope = { markSucceeded: vi.fn() }
+    await modalOptions.props.onConfirm(scope)
+
+    expect(mocks.prepareTransactionPlan).toHaveBeenLastCalledWith(plan, { account: USER })
+    expect(mocks.executePreparedPlan).toHaveBeenCalledWith(refreshed)
+    expect(mocks.executePlan).not.toHaveBeenCalled()
+    expect(mocks.finalizeTxAndRedirect).toHaveBeenCalledWith({ scope })
+  })
+
+  it('keeps each open review modal bound to the plan and account it displayed', async () => {
```
LeonardEulerXYZ: [P1] Apply this review-session invariant to the active `/position/[number]/multiply` sibling too. That page still builds into module refs (`multiply.vue:943-975`), passes `preparedPlan.value` to the modal, closes `onConfirm` only over `sendMultiply(execution)`, and then executes whatever `preparedPlan.value` contains at confirmation (`:989-1020`). Opening review A, rebuilding B, then confirming A therefore executes B; it also never freshly prepares/canonical-compares Pyth/plugin data. This test proves only the `/borrow/...` composable path. Add the same captured-session + fresh-prepare + `requireReviewedBatchPreparedExecution` flow and an overlap regression for the position route.

composables/useEulerTx.ts
```diff
@@ -1605,16 +1929,19 @@ export const useEulerTx = () => {
...
+      throw new Error(SAFE_DURABLE_EXECUTION_REQUIRED_ERROR)
+    }
 
     let effectivePrepared = prepared
     if (isKnownSafe) {
       // A Safe never signs permit2 messages. If the envelope was prepared
       // before Safe detection resolved, re-resolve its approvals with
       // permit2 off — resolution overwrites `resolved` on each item, so the
-      // repaired plan is used for both the bundle and the fallback.
+      // repaired plan is the exact plan submitted through the Safe bundle.
```
LeonardEulerXYZ: [P1] Revalidate the whole reviewed ceremony when Safe detection changes from unresolved/false to true, not only this terminal prepared plan. `useSignaturePreference` disables signatures while detection is unresolved, but direct migration and batch reviews can still open with `bundledReview=false`/a null latch. If the same Safe connector resolves before confirm, their authorization grants take the sequential `sendPlainTransactions` path first; that helper detects Safe only for receipt polling and calls `sendTransactionAsync` without the durable reservation/interlock. The terminal plan reaches this repaired Safe path too late. Gate review on resolved classification or require re-review on any classification/connector change, and cover outgoing migration, inbound migration, and batch unresolved→Safe transitions.

utils/pending-safe-bundle-submission.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const reservePendingSafeBundleSubmission = (
+  storage: StorageLike,
+  reservation: PendingSafeBundleSubmission,
+) => {
+  const submissions = loadPendingSafeBundleSubmissions(storage)
+  if (submissions.some(submission => isSameContext(submission, reservation.account, reservation.chainId))) {
+    throw new Error('A previous Safe bundle submission is unresolved. Verify it in Safe before retrying.')
+  }
+  savePendingSafeBundleSubmissions(storage, [...submissions, reservation])
```
LeonardEulerXYZ: [P1] Read this reservation back and verify its exact identity before allowing `sendCalls`. Unlike the batch persistence path, this helper assumes `setItem` retained the record. A `StorageLike` that silently drops writes makes `reservePendingSafeBundleSubmission()` return successfully while a subsequent load is empty; `useEulerTx` then opens Safe with no durable retry lock. Reload/hash-loss can therefore permit a duplicate proposal. Reload and compare `reservationId`, account, chain, and the persisted envelope immediately after the write, failing closed on any mismatch; keep a dropped-write regression.

composables/useTxBatch.ts
```diff
@@ -2232,19 +2513,28 @@ export const useTxBatch = () => {
 
     logBatchDiag('getEntryPlanningAccount:base-snapshot-fetch', { owner: o, chainId: cid })
     const sdk = await getEulerSdkFresh()
-    baseAccountSnapshot = await fetchBaseAccountSnapshot(sdk, cid, ownerAddress)
+    assertContextCurrent()
+    const fetched = await fetchBaseAccountSnapshot(sdk, cid, ownerAddress)
```
LeonardEulerXYZ: [P1] The direct-add guards here do not cover the older resimulation publisher. `resimulate()` captures owner/chain, awaits `fetchBaseAccountSnapshot`, and assigns `baseAccountSnapshot` at `:2053-2056` before its first `resimToken` check at `:2088`. An owner/chain reset clears the snapshot, but the superseded continuation can write owner A back afterward; this empty-cart path then returns that non-null snapshot to a new owner-B entry without validating ownership. A deferred probe delivered A's account to B's builder with no second fetch. Check token/context before every resimulation publication (especially the base snapshot and `lastMerged`), and validate cached snapshot owner/chain before reuse.

LeonardEulerXYZ (changes requested):
Supplementary exact-head review after all three independent lanes completed. Four additional blockers survived source tracing: a missed position-multiply sibling, unresolved→Safe ceremony drift before terminal execution, an unverified ordinary Safe reservation write, and a stale resimulation publisher that can restore another owner's planning snapshot. These are additive to review 4952435706; details inline.

dglowinski:

PR #796, opened 2026-08-10
feat: Safe multisig badge on governance addresses + oracle router governor row

kasperpawlowski:
## Summary

- Detect whether governance-related addresses are Safe multisigs and show a small Safe icon with the signer configuration, e.g. **(3/7)** = threshold / owners (LITE-94)
- Add an **Oracle governor** row to the vault addresses section, read on-chain from the EulerRouter (`governor()`); renounced governance renders as "None"
- Extract the copy-pasted address row markup from the four overview blocks into a shared `VaultOverviewAddressValue` component

## Implementation

- Detection is fully on-chain with no Safe API dependency, so it works on every enabled chain: `masterCopy()` (special-cased in every Safe proxy since v1.1.1) is validated against the canonical singleton list from [safe-deployments](https://github.com/safe-global/safe-deployments) (v1.1.1–v1.5.0 incl. eip155 variants), then `getThreshold()`/`getOwners()` provide the badge data; Safe threshold/owner invariants are enforced so lookalikes are rejected
- `useSafeAddressInfo` / `useOracleRouterGovernor` composables cache per `chainId:address` (5-min TTL, in-flight dedup, transient RPC failures not cached) via a new reusable `createOnchainLookupCache` util; the three probe reads fire concurrently and coalesce into one RPC request via transport batching
- Probes run client-side only and only for governance rows (`check-safe` prop): risk manager, fee receivers, oracle governor, Earn owner/curator/guardian, Securitize governor
- Badge tooltip shows the Safe version, an explainer, and the shortened owner list
- The same detection primitive was added to the SDK as `safeAccountService` (euler-xyz/euler-sdks#94); this app-side probe is intentionally self-contained and can swap to `sdk.safeAccountService.fetchSafeAccount` once a release ships with it

## Test plan

- [x] Unit tests: singleton recognition + Safe invariant validation (`tests/utils/safe-account.test.ts`)
- [x] Unit tests: cache TTL / dedup / failure semantics (`tests/utils/onchain-lookup-cache.test.ts`)
- [x] Composable tests: detection, non-Safe/EOA nulls, sentinel-address skip, cross-instance caching (`tests/composables/useSafeAddressInfo.test.ts`, `tests/composables/useOracleRouterGovernor.test.ts`)
- [x] Full suite: 1635 passing incl. golden/parity; `nuxt typecheck` and ESLint clean
- [ ] Visual check on a vault with a Safe governor (badge placement, tooltip, owners list)
- [ ] Check a vault with renounced router governance shows "None"

## Summary by CodeRabbit

* **New Features**
  * Added Safe multisig badges showing Safe version, signature threshold, and owner count.
  * Added explorer links, shortened labels, special address labels, and copy-to-clipboard actions for vault addresses.
  * Added Oracle governor visibility for applicable borrowable vaults.
  * Added Safe address detection for relevant vault roles.
  * Standardized address presentation across vault overview sections.

* **Bug Fixes**
  * Improved handling of unavailable, invalid, or unsupported on-chain address data.


tests/utils/safe-account.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    expect(resolveSafeAccountInfo(SAFE_141_SINGLETON, 4n, OWNERS)).toBeNull()
+    expect(resolveSafeAccountInfo(SAFE_141_SINGLETON, BigInt(Number.MAX_SAFE_INTEGER) + 1n, OWNERS)).toBeNull()
+  })
+
+  it('rejects owner lists a Safe cannot have', () => {
+    const zeroOwner = '0x0000000000000000000000000000000000000000'
+    const sentinelOwner = '0x0000000000000000000000000000000000000001'
+    expect(resolveSafeAccountInfo(SAFE_141_SINGLETON, 1n, [...OWNERS, zeroOwner])).toBeNull()
+    expect(resolveSafeAccountInfo(SAFE_141_SINGLETON, 1n, [...OWNERS, sentinelOwner])).toBeNull()
+    expect(resolveSafeAccountInfo(SAFE_141_SINGLETON, 1n, [OWNERS[0], OWNERS[0].toUpperCase().replace('0X', '0x')])).toBeNull()
```
LeonardEulerXYZ: **Merge blocker:** this expression is inferred as `string`, so the exact PR head fails `npm run typecheck` with `TS2322: Type 'string' is not assignable to type `0x${string}``. Preserve the `Address` type for the mixed-case duplicate fixture (for example, assign/cast the transformed value as `Address`) and rerun typecheck.

LeonardEulerXYZ (changes requested):
The visual change is clean in the Railway preview, and the earlier transport/owner-invariant findings are fixed on this head. One merge blocker remains: `npm run typecheck` fails at `tests/utils/safe-account.test.ts:58` because `toUpperCase().replace(...)` widens the second owner from `Address` to `string` (`TS2322`). The focused feature suite passes (26/26), lint has no errors, and `git diff --check` is clean.

Deployment note, separate from the source blocker: GitHub marks Railway deployment `5832107876` successful for this exact Lite SHA, but the currently served SDK bundle still contains the pre-fix `safeAccountService` probe (immediate null on rejected `masterCopy`, without the transport-failure and owner-invariant handling now present at SDK PR #94 head `5ebab4bc`). Lite itself does not reference `safeAccountService`, so the screenshots validate the local Lite implementation and visual delta, not integration with the latest SDK head.

components/entities/safe/SafeAccountBadge.vue
```diff
@@ -0,0 +1,27 @@
+<script setup lang="ts">
+const { address } = defineProps<{ address: string }>()
+
+const { safeInfo } = useSafeAddressInfo(() => address)
+
+const tooltipText = computed(() => {
+  if (!safeInfo.value) return ''
+  const { threshold, owners, version } = safeInfo.value
+  return `This address is a Safe smart account (v${version}). Executing a transaction from it requires ${threshold} of its ${owners.length} owner signatures.`
```
LeonardEulerXYZ: This overstates what the owner threshold controls. An enabled Safe module can execute via `execTransactionFromModule` without owner confirmations; Safe v1.4.1 explicitly describes that path as execution “without further confirmations.” Since the badge does not inspect modules or guards, please say that the Safe is configured with an X-of-Y **owner threshold**, rather than claiming every transaction requires those signatures.

components/entities/safe/SafeAccountBadge.vue
```diff
@@ -0,0 +1,27 @@
...
+})
+</script>
+
+<template>
+  <UiHoverPreviewTooltip
+    v-if="safeInfo"
+    title="Safe multisig"
```
LeonardEulerXYZ, line 24: Non-blocking accessibility point: `UiHoverPreviewTooltip` derives the trigger's `aria-label` from this fixed title, and that label overrides the visible `(threshold/owner-count)` text in the accessible name. Pass a dynamic `aria-label` such as `Safe multisig: 2 of 5 owner threshold` so screen-reader users receive the same status as sighted users.

LeonardEulerXYZ (commented):
Two non-CI points from the earlier-head review remain on this exact head: one control-boundary wording correction and one accessibility improvement.

LeonardEulerXYZ (changes requested):
The CI/typecheck blocker is resolved on this exact head: local `npm run typecheck`, the focused owner-list test, and all three GitHub CI jobs pass. I am not re-approving yet because the separate control-boundary wording finding remains unchanged at `SafeAccountBadge.vue:9`: Safe modules can execute through `execTransactionFromModule` without owner confirmations, so the tooltip must describe the configured X-of-Y **owner threshold** rather than claim every transaction requires those signatures. The accessibility aria-label point remains non-blocking.

LeonardEulerXYZ (approved):
Reviewed exact head `b5d8cd38306b99479f0cb0673a91f1af6f715614`.

The previous control-boundary blocker is resolved. The tooltip now describes only the configured X-of-Y owner threshold, explicitly avoids claiming that every Safe execution requires owner confirmations, and the badge exposes the same status through a dynamic accessible name.

Validation:
- `npx vitest run tests/utils/safe-account.test.ts` — 7/7 passed
- `npm run build` — passed
- GitHub lint, test, and typecheck checks — green on this head
- `git diff --check` — clean

No remaining blockers found.

LeonardEulerXYZ (approved):
Reviewed exact head `6adf4f999d1059122ab4a8d7164892ffb8f814b9`.

The self-ownership parity fix is correct: the local Safe probe now rejects a recognized-singleton lookalike that lists its own address as an owner, matching Safe `OwnerManager`'s `GS203` invariant. The regression test covers it alongside zero, sentinel, duplicate-owner, threshold, and malformed-return cases.

Validation:
- `npx vitest run tests/utils/safe-account.test.ts`: 8/8 passed
- `npm run typecheck`: passed
- `npm run build`: passed
- `git diff --check origin/development...HEAD`: clean
- GitHub lint/typecheck/test checks: green

Non-blocking: the internal `SafeAccountInfo.threshold` JSDoc still says "required to execute a transaction"; it would be more exact to mirror the badge/SDK wording and scope this to owner-authorized transactions, since modules can execute without owner signatures. This does not affect the rendered badge copy.

No blockers.

kanvgupta (dismissed):

LeonardEulerXYZ (approved):
Re-reviewed the delta since the prior approved head and verified current head `0b1bd5015b7119bde6c1b235d030beb4eaaab978`.

Local verification:
- clean `npm ci`
- `npm run test:run`: 1,639 passed, 1 skipped
- `npm run typecheck`
- `npm run build`
- built-server smoke: `GET /` returned 200 with the Euler Lite page

No blocking findings.

Seranged (approved):

kanvgupta (approved):

PR #797, opened 2026-08-10
feat: disable Permit2 and batch approval+EVC into one tx for Safe wallets

kasperpawlowski:
## Summary

When the connected wallet is a Safe multisig (LITE-221):

1. **Permit2/signatures force-disabled** — `useSignaturePreference().signaturesEnabled` becomes an effective computed pinned to `false` while a Safe is connected. The stored user preference is untouched and returns when a regular wallet reconnects. The settings toggle is disabled with an explanation. Because every flow (including the call sites that pin `usePermit2: useSignatures` in migrate/swap pages) reads this composable, the approval flow is forced everywhere with no per-flow changes.
2. **approve + EVC `batch()` submitted as ONE Safe transaction** — instead of the SDK's sequential send-and-wait-receipt loop, the resolved plan is converted to raw calls and submitted via EIP-5792 `wallet_sendCalls` (`forceAtomic: true`). Safe's provider turns the bundle into a single MultiSend proposal, so signers run one signature ceremony instead of one per transaction. The existing `waitForSafeTransactionExecution` poller resolves the returned `safeTxHash` bundle id to the executed transaction unchanged.

## Implementation

- `composables/useSafeWallet.ts` — reactive app-wide "connected wallet is a Safe" signal built on `getSafeWalletProvider` (iframe connector id, Safe wallet name, or WalletConnect peer metadata), kept current via `watchAccount` with a race guard
- `utils/transaction-plan-calls.ts` — `transactionPlanToCalls(plan, sdk, chainId)`: resolved plan → ordered `{ to, data, value }` calls; throws `PlanNotBundleableError` for unresolved approvals, permit2 signatures, and CoW items
- `useEulerTx` — both `executePlan` and `executePreparedPlan` try the bundle first when a Safe provider is detected (raw plans run plugin processing + approval resolution before converting); wallet-context drift is asserted before submission; sequential execution remains the fallback for single-call plans and non-bundleable plans, and for all non-Safe wallets. Results keep the `{ plan, hashes, receipts }` shape so post-tx invalidation/subgraph sync work unchanged
- No wagmi upgrade needed — `sendCalls` ships in the installed `@wagmi/core` 3.6.4 via `@wagmi/vue/actions`
- UI: approve steps lose the "Separate tx" badge when bundled (`bundledApprovals` in `StepDecodingContext`); batch review approval rows say "bundled in batch" only for Safe (previously claimed it for everyone); settings toggle disabled state + copy

## Not in scope

- Migration authorization grants/revokes keep their existing sequential Safe-aware path
- Gating on `wallet_getCapabilities` — Safe's provider reports the legacy `atomicBatch` shape, so detection stays connector-based and `sendCalls` failures surface directly

## Test plan

- [x] `useSafeWallet`: iframe connector, regular connector, connector switching, stale-detection race, single watcher
- [x] `useSignaturePreference`: forced-off while Safe connected, preference preserved across Safe sessions, stray writes ignored
- [x] `transactionPlanToCalls`: ordering + value summing, permit2/unresolved/CoW/missing-EVC rejections, contractCall encoding
- [x] `useEulerTx`: bundle submission shape (`forceAtomic`, calls), receipts from Safe execution, sequential fallback for single-call and permit2 plans, account-drift guard, malformed bundle id, raw-plan resolution path
- [x] Full suite 1633 passing; typecheck + lint clean
- [ ] E2E with a real Safe (iframe app + WalletConnect): lend deposit, supply, borrow, multiply, repay, earn deposit each produce a single proposal
- [ ] Non-Safe wallet regression pass (Permit2 unchanged)

## Summary by CodeRabbit

* **New Features**
  * Safe wallets can execute supported approval and transaction plans as atomic bundles.
  * Unsupported plans automatically fall back to sequential execution.
  * Safe executions can continue in the background while awaiting co-signers, with completion or failure shown in notifications.

* **Improvements**
  * Transaction reviews distinguish bundled approvals, signatures, and separate transactions.
  * Signatures are automatically disabled for Safe wallets while saved preferences are preserved.
  * Wallet detection responds reliably to account and provider changes.

* **Bug Fixes**
  * Improved handling of failed, invalid, or incomplete bundle executions.


LeonardEulerXYZ (changes requested):
Reviewed exact head `9eeb3007e674d4594ac41df04b1222d34d719b6e`.

The bundle encoding/order and account/chain assertion are sound, and the ordinary test/build gates are green. Three independent blockers remain:

1. **A reverted Safe bundle is finalized as success** (`composables/useEulerTx.ts:1440-1445`). `waitForSafeTransactionExecution()` can return a mined receipt with `status: 'reverted'`; the new bundle path returns it without the success check already used by `sendPlainTransactions`. Callers then run `finalizeExecution`, clear the batch, redirect, and report success although no protocol state changed. A direct exact-head runtime probe returned `{ status: "reverted" }`. Require `execution.receipt.status === 'success'` before returning and add a reverted-receipt regression test.

2. **Safe detection is not fail-closed during plan preparation** (`composables/useSafeWallet.ts:7-23`, `composables/useEulerTx.ts:1141-1149`). `false` means both "ordinary wallet" and "Safe detection is still pending/failed". A plan can therefore be prepared with `usePermit2: true` before delayed WalletConnect/Safe detection completes; `executePreparedPlan()` cannot repair that resolved envelope and falls back to the sequential Permit2/signature path. Expose pending/resolved detection state and prevent preparation/submission until the current connector generation resolves, with a delayed-detection regression test.

3. **The definitive Safe execution branch still trusts the lagging reactive preference** (`composables/useEulerTx.ts:1469-1511`). After `getSafeWalletProvider(connector)` has positively identified a Safe, both approval resolution and the sequential fallback still use `signaturesEnabled.value`. Force `usePermit2: false` from the authoritative provider result; do not let an asynchronous UI ref re-enable Permit2 for a known Safe.

There is also a review-display mismatch: `bundledApprovals` is derived from `isSafeWallet` alone, while actual execution falls back for Permit2, unsupported items, and fewer than two calls. The review UI can therefore label an approval as bundled when the user will receive sequential submissions. Reuse actual bundle eligibility from the resolved plan rather than duplicating or approximating it in the display layer.

Validation:
- focused Safe/transaction tests — 46/46 passed
- full Lite suite — 1,637/1,637 passed across 172 files
- `npm run lint` — 0 errors (6 existing warnings outside the changed cohort)
- `npm run build` — passed
- `git diff --check` — clean

The current tests cover happy-path bundling, unsupported-plan fallback, account drift, and hash shape, but they stub signatures off and omit reverted receipts and delayed Safe detection, so they do not discriminate these regressions.

LeonardEulerXYZ (changes requested):
Reviewed exact head `55a9e1fbf55563efd009347a5ead865859a82a10`.

The original execution findings are fixed correctly: reverted receipts throw, unresolved Safe detection forces Permit2 off, a tainted prepared envelope is repaired before both bundle and fallback execution, the definitive Safe path hardcodes `usePermit2: false`, `sendCalls` is pinned to the detected connector, and settled-state bundle labels mirror the execution gate.

One review-state invariant remains.

**Blocker — connector changes retain the previous positive Safe result while the new probe is unresolved.**

`detectForConnector()` sets `isSafeWalletResolved = false` but does not clear `isSafeWallet`. Both `OperationReviewModal.vue` and `BatchReviewModal.vue` consume `isSafeWallet` directly when deciding whether to label approvals as bundled. On a Safe → slow regular connector transition, the new EOA plan is correctly forced to direct approvals, but the review UI can still say those approvals are "bundled in batch" until detection settles. Execution is fail-closed; the transaction-count representation is not.

I added a temporary adversarial test, ran it, then removed it. It reproduced the stale state:

```text
onChange(slowInjectedConnector)
isSafeWalletResolved === false
expected isSafeWallet === false; received true
```

Fix either the state invariant (clear `isSafeWallet` when a new current-connector detection starts) or gate every display consumer on `isSafeWalletResolved && isSafeWallet`. Please retain a Safe → unresolved EOA regression test.

Validation:
- focused Safe/Permit2/bundle tests: 40/40 passed
- full suite: 1,642/1,642 passed
- typecheck, lint, and production build: passed
- `git diff --check origin/development...HEAD`: clean
- GitHub lint/typecheck/test checks: green

LeonardEulerXYZ (changes requested):
Reviewed exact head `e68a55a24858722fe25cb04599ad4ff9fb304c90`.

The stale Safe→EOA display state is fixed, known-Safe identity now keeps reactive signature gating off when provider acquisition fails, and successful-provider prepared Safe envelopes are normalized to `usePermit2:false`. One degraded execution path remains.

**Blocker — provider-unavailable prepared Safe envelopes still reach sequential execution with `usePermit2:true`.**

`executePreparedPlan()` only normalizes `effectivePrepared` inside `if (safeWalletProvider && connector)`. For the exact supplement case—connector identity is `safe`, but `getProvider()` rejects—`safeWalletProvider` is undefined, so that block is skipped. The original prepared envelope is then passed unchanged to `executePreparedTransactionPlan()`.

I added a temporary adversarial regression, ran it, and removed it:

```text
connector: { id: 'safe', name: 'Safe', getProvider: rejects }
prepared.usePermit2: true
expected SDK sequential input usePermit2: false
received: true
```

Use `isSafeConnectorIdentity(connector)` in the execution path as well: derive an authoritative-Safe boolean from provider detection OR connector identity, normalize every prepared envelope under that boolean, but only attempt bundling/polling when the provider exists. Keep the provider-rejection prepared-envelope test.

Validation:
- focused Safe/Permit2/bundle tests: 42/42 passed
- full suite: 1,644/1,644 passed
- typecheck and lint: passed (six unrelated warnings)
- production build: passed
- GitHub lint/typecheck/test: green
- temporary degraded-mode challenge: failed as described
- `git diff --check origin/development...HEAD`: clean

LeonardEulerXYZ (approved):
Reviewed exact head `4598b568a60143989b4dabad2a8189959fa917ee`.

The remaining degraded-mode blocker is resolved.

`isKnownSafe = Boolean(safeWalletProvider) || isSafeConnectorIdentity(connector)` now governs every Permit2 decision in raw and prepared execution, while `safeWalletProvider` continues to govern only provider-dependent bundling and Safe status polling. Therefore an identity-known Safe with unavailable provider degrades to sequential plain approvals without carrying or signing Permit2 state.

Both adversarial prepared-envelope cases are covered:

- `usePermit2:true` with no resolved Permit2 item is normalized before sequential execution.
- A Permit2-resolved envelope is re-resolved with `usePermit2:false`; the repaired plan executes sequentially with no typed-data signature.

I mutation-tested the invariant by temporarily removing connector identity from both `isKnownSafe` expressions. The two new degraded-mode tests failed exactly as intended, then the source was restored cleanly.

Validation:
- focused Safe/Permit2/bundle tests: 44/44 passed
- full suite: 1,646/1,646 passed
- typecheck: passed
- lint: 0 errors, six unrelated warnings
- production build: passed
- GitHub lint/typecheck/test checks: green
- mutation challenge: 2 targeted failures as expected
- `git diff --check origin/development...HEAD`: clean

No blockers.

LeonardEulerXYZ (changes requested):
Re-reviewed the delta since the prior approved head and verified exact head `aebebcf9cfed032abe6927b4e6441b3ce793257c`.

There are three blocking correctness gaps:

1. **A detached failure can be reported as confirmed.** `OperationReviewModal` hands the raw `onConfirm()` promise to `detach()`, and `detach()` treats every resolved promise as success. However the existing operation handlers are not rejection-safe contracts: for example both migration `sendMigration` functions catch execution errors, call `showError`, and return normally. If the user closes while such a Safe proposal is pending and it later fails, the callback resolves and this code emits `Safe transaction confirmed`. Please make the detached result explicit (or require and enforce rejection) and cover a swallowed execution failure.

2. **Navigation suppression is neither complete nor execution-scoped.** Only `useTxFinalization.finalizeTxAndRedirect()` consults the new guard. Existing migration handlers bypass it and call `router.replace(...)` directly after success, so a detached migration still redirects whatever page the user moved to. The global count also cannot identify which execution is finalizing: if operation A is detached and operation B is attended, A completing while B is attached makes `shouldSuppressPostTxNavigation()` return false and A can hijack B. Please bind suppression to an execution token/context and apply it to every redirecting completion path, with the detached-A/attached-B interleaving covered.

3. **The inherited ToS-layer fix shifts state layers but not failure indices.** The SDK includes the loose ToS operation in `failedBatchItems.operationIndex`. After stripping that state layer, `useTxBatch` still consumes the raw index, so a failure in the first cart operation is unmarked and `hasFailedOps` stays false; later operations mark the following row. That can let `executeBatch()` reach prerequisite broadcasts before gas estimation catches the revert. Normalize failure indices by the same leading-plugin count, while leaving plugin-only failures at batch level.

Verification of the committed paths is green (57 focused tests and Nuxt typecheck), but the above interleavings are not covered.

LeonardEulerXYZ (approved):
Re-reviewed the complete fix delta from aebebcf9cfed032abe6927b4e6441b3ce793257c at exact head bf9eacb54dcc89bbb3b96ea40e7a31db42c03ef7. The execution-scoped single-slot design closes the prior false-confirmation and global-attribution failures: Safe classification is latched at submit, detached completion requires an explicit flow success marker, batch and operation redirects respect the detached execution, and all sequential send paths pin the captured connector. New submissions are gated while the detached slot is occupied. I also checked the direct migration success markers inherited by dependent branches. Verification: clean npm ci; 72 focused tests; full suite 1656 passed / 1 skipped; typecheck, lint, and production build all passed.

LeonardEulerXYZ (changes requested):
The slot-exclusivity fix and #802 operation map are correct, and the three edited success tails now mark/suppress correctly. The direct-flow audit is still incomplete, however.

Several unchanged `OperationReviewModal` callers remain Safe-detachable and still execute `modal.close()` plus delayed `router.replace()` directly without `markTrackedExecutionSucceeded()` or `shouldSuppressPostTxNavigation()`. Confirmed examples on this exact head:

- Earn supply: `pages/earn/[vault]/index.vue` after `executePlan()`
- Earn withdraw: `pages/earn/[vault]/[subAccount]/withdraw.vue` after `executePlan()`
- Lend supply: `pages/lend/[vault]/index.vue` after `executePreparedPlan()`
- Lend withdraw/swap-withdraw: `pages/lend/[vault]/[subAccount]/withdraw.vue`

A successful Safe proposal detached from any of these modals still redirects the user away from their current flow and resolves without the explicit success marker, producing the “did not complete” warning. Please complete the inventory across every `OperationReviewModal` submit callback, route all direct success tails through one shared finalize contract (or equivalent marker/suppression gate), and add real-callback regressions for at least Earn and Lend. The CoW completion watchers can remain separate because #798 prevents Safe submission there.

Verification on this exact head: clean install, 58 focused detachment/batch tests, and typecheck passed. The focused suite contains no integration test covering these unchanged callbacks.

LeonardEulerXYZ (changes requested):
The eight edited page flows now satisfy both success marking and redirect suppression, and the slot/map fixes remain sound. The complement is still not empty because the audit predicate requires success attribution independently of navigation.

Two additional Safe-detachable `OperationReviewModal` callbacks execute successfully and close without `markTrackedExecutionSucceeded()`:

- `components/entities/reward/RewardUnlockItem.vue`: `executePlan(unlockPlan)` → `modal.close()`
- `components/entities/portfolio/PortfolioSdkRewardItem.vue`: `executePlan(plan.value)` → refresh → `modal.close()`

They do not navigate, so suppression is irrelevant; however, after detachment their promises resolve with no success marker and the tracker emits “Safe transaction did not complete” for confirmed reward unlock/claim proposals.

Please include all detachable submit callbacks in the invariant: every successful `OperationReviewModal` execution must mark the same tracked execution, whether or not it later navigates. Add focused real-callback regressions for reward unlock and reward claim. My structural scan found these two unmarked detachable execution/close paths; the other page/composable candidates are gated or use the shared finalizer.

LeonardEulerXYZ (changes requested):
The two success markers fix false incomplete toasts, but both detached reward success tails still mutate the global modal stack.

`useModal.close()` without an ID executes `list.pop()`. After the user detaches the reward review and opens another modal while Safe signatures are pending, the confirmed reward flow calls unscoped `modal.close()` and closes that newer modal. Thus “neither navigates” is not sufficient: modal-stack teardown is itself a post-detachment UI side effect.

Please gate the success-path close in both `RewardUnlockItem.vue` and `PortfolioSdkRewardItem.vue` with `shouldSuppressPostTxNavigation()` (the existing name is broader in practice), or capture and close only the original modal ID. Keep the success marker unconditional. Add a regression that detaches, opens another modal, resolves the reward execution, and asserts the newer modal remains open while the success toast fires.

The structural invariant should be: (1) every successful detachable execution marks success; independently, (2) every subsequent global UI teardown or route change is suppressed when detached. The eight page flows satisfy both; these two reward flows currently satisfy only clause 1.

LeonardEulerXYZ (changes requested):
The two successful reward exits now satisfy both clauses. One bounded gap remains in the same `RewardUnlockItem.unlock()` detachable callback: its non-success exits still perform unscoped global closes while the async review submission may already be detached.

Specifically:

- first `isBatchActive` recheck after `refreshLocks(true)` → `modal.close()`
- second recheck after `buildUnlockREULPlan(...)` → `modal.close()`
- `changed | missing | unavailable` result → `modal.close()`

A user can submit, detach while refresh/build is pending, open another modal, and then hit any of these exits; `close()` still executes `list.pop()` and removes the newer modal. Gate these three closes with `shouldSuppressPostTxNavigation()` as well, or scope every close in this callback to the original review modal ID.

I widened the structural scan from successful exits to every unscoped `modal.close()` in files containing detachable review flows. These three are the only remaining applicable calls. The CoW completion watcher is Safe-gated and remains a definitional non-member; all other detachable closes are suppressed or routed through the shared finalizer.

LeonardEulerXYZ (approved):
Approved at exact head `08ccb3a7`.

The detachment invariant now holds across the applicable surface:

1. every successful detachable execution marks the tracked execution;
2. every subsequent route change or unscoped global modal teardown is suppressed after detachment.

The three remaining `RewardUnlockItem.unlock()` exits are correctly gated. A widened scan of detachable review files leaves only the CoW completion watcher, which is Safe-gated and cannot detach. The direct page flows, reward flows, batch path, shared finalizer, single-slot ownership, and plugin-aware operation attribution are consistent.

Verification: GitHub lint/typecheck/test green; clean local install; 58 focused detachment/batch tests and typecheck pass.

LeonardEulerXYZ (dismissed):
Re-reviewed exact head `38ee37ac2197f150b891d0945891c120ca63103e` since approved head `08ccb3a796db1b5946bccf8eebb4e541a4948c1b`.

Requesting changes for one execution-attribution blocker:

**[P1] Bind the success marker to the execution that produced it** — `composables/useSafeExecutionDetachment.ts:31`

The account watcher correctly abandons execution A and clears the module-scoped slot, which allows execution B to start under the new wallet. However, A's asynchronous flow can still reach `markTrackedExecutionSucceeded()` afterward. That function marks whichever record currently occupies the slot, so A can set B's `succeeded` flag. If B is detached and resolves without reaching its own finalize point, B then emits the false `Safe transaction confirmed` toast instead of the warning.

I reproduced the interleaving at this exact head:

1. begin and detach A;
2. abandon A on the wallet transition;
3. begin and detach B;
4. let A's late success tail call `markTrackedExecutionSucceeded()`;
5. resolve B without its own success marker.

The regression fails because `toast.success('Safe transaction confirmed')` is called once. The marker needs an execution-bound token/handle (or equivalent identity check), so a continuation can mutate only its own record, including after abandonment and slot reuse. Please keep a regression for this overlap.

Validation: the existing focused detachment/Safe/finalization suites pass (20 tests), and hosted lint/typecheck/test are green. The additional discriminating overlap test fails as above.

Seranged (dismissed):

Seranged (dismissed):

Seranged (approved):

kanvgupta (approved):

PR #798, opened 2026-08-10
feat: disable CoW swap quotes for Safe wallets

kasperpawlowski:
## Summary

Removes CoW Swap from the quote pipeline for Safe multisigs, routing them to on-chain swap providers (LITE-322). **Stacked on #797** — depends on `useSafeWallet`; retarget to `development` once #797 merges.

## Why

CoW orders require a recoverable 65-byte ECDSA signature: the SDK's `cowExecutor` funnels every order (and EVC permit) signature through `normalizeCowSignature`, which throws `Unsupported signature length` for anything else — and it does so **after** the user has already sent ERC-20 approval transaction(s). On close-position flows a burned EVC permit nonce compounds the damage. A Safe cannot produce such a signature, so for Safe wallets the CoW route is unusable-by-construction today. Proper Safe+CoW support (presign / EIP-1271 order placement) requires SDK `cowExecutor` changes and is out of scope — this PR makes the failure impossible instead.

## Implementation

No SDK changes needed — CoW is opt-in app-side (`includeCowSwap`, filtered in `useSwapApi.getSwapProviders` before any quote request), and `useSwapQuotesParallel` already evicts on-screen CoW cards and clears a manual CoW selection when the gate flips mid-session, auto-falling through to the next-best provider.

- `composables/useCowSwapEligibility.ts` — `cowSwapForcedOff = isSafeWallet || !isSafeWalletResolved` (fail-closed while detection is pending, mirroring the signature-preference gate from #797)
- ANDed into the three `includeCowSwap` call sites (`useMultiplyForm`, `useCollateralSwapRepay`, `borrow/swap.vue`), preserving a literal `false` return so the eviction watcher fires
- Defence-in-depth: `useCowSwapExecutionCore.assertTransactionsEnabled` throws `CoW Swap is not available with Safe wallets` before any transaction is sent, should a stale CoW quote ever survive to a submit branch

## Test plan

- [x] `useCowSwapEligibility`: regular wallet, Safe, detection pending (fail-closed), mid-session Safe connect
- [x] `useCowSwapExecutionCore`: execution rejects for Safe and for pending detection before any transaction
- [x] Existing quote-pipeline behavior locked by `useSwapQuotesParallel` tests (callable `includeCowSwap` re-evaluated per sweep; eviction on flip to `false`)
- [x] Full suite 1648 passing; typecheck + lint clean
- [ ] Manual: Safe connected on mainnet → multiply/repay/collateral-swap quotes show only on-chain providers; EOA unaffected

### DIFF
--- a/composables/borrow/useMultiplyForm.ts
+++ b/composables/borrow/useMultiplyForm.ts
@@ -103,6 +103,7 @@ export const useMultiplyForm = (options: UseMultiplyFormOptions) => {
   const { getBalance } = useWallets()
   const { finalizeTxAndRedirect } = useTxFinalization()
   const { entryCount: batchEntryCount } = useTxBatch()
+  const { cowSwapForcedOff } = useCowSwapEligibility()
   const {
     version: rewardsVersion,
     getSupplyRewardApy,
@@ -167,7 +168,7 @@ export const useMultiplyForm = (options: UseMultiplyFormOptions) => {
   } = useSwapQuotesParallel({
     amountField: 'amountOut',
     compare: 'max',
-    includeCowSwap: () => batchEntryCount.value === 0 && !isMultiplySavingCollateral.value,
+    includeCowSwap: () => !cowSwapForcedOff.value && batchEntryCount.value === 0 && !isMultiplySavingCollateral.value,
     buildTxPlanForQuote: (quote, _provider, context) => buildMultiplyPlanFromQuote(quote, context.account),
     getStateOverrideOptions: () => buildMultiplyStateOverrideOptions(),
     // First quote in each sweep computes plugin prefetch (Pyth Hermes updates
--- a/composables/cowswap/useCowSwapExecutionCore.ts
+++ b/composables/cowswap/useCowSwapExecutionCore.ts
@@ -46,6 +46,7 @@ export const useCowSwapExecutionCore = () => {
   const { signTypedDataAsync } = useSignTypedData()
   const { sendTransactionAsync } = useSendTransaction()
   const { isSpyMode } = useSpyMode()
+  const { cowSwapForcedOff } = useCowSwapEligibility()
   const { triggerPortfolioRefresh } = usePortfolioRefresh()
 
   const status = ref<CowSwapExecutionStatus>('idle')
@@ -69,6 +70,10 @@ export const useCowSwapExecutionCore = () => {
 
   const assertTransactionsEnabled = () => {
     if (isSpyMode.value) throw new Error('Transactions are disabled in spy mode')
+    // Backstop behind the quote-level gate: a Safe cannot produce the ECDSA
+    // order signature the CoW executor requires, and failing here is cheaper
+    // than failing after the approval transactions have been sent.
+    if (cowSwapForcedOff.value) throw new Error('CoW Swap is not available with Safe wallets')
   }
 
   const requireWallet = () => {
--- a/composables/repay/useCollateralSwapRepay.ts
+++ b/composables/repay/useCollateralSwapRepay.ts
@@ -89,6 +89,7 @@ export const useCollateralSwapRepay = (options: UseCollateralSwapRepayOptions) =
   const { account: planAccount } = usePlanAccount()
   const { client: rpcClient } = useRpcClient()
   const { entryCount: batchEntryCount, getMergedPlan } = useTxBatch()
+  const { cowSwapForcedOff } = useCowSwapEligibility()
   const { settings } = useUserSettings()
   const enableIntrinsicApy = computed(() => settings.value.enableIntrinsicApy)
   const { getSupplyRewardApy, getBorrowRewardApyForCollaterals } = useRewardsApy()
@@ -157,7 +158,7 @@ export const useCollateralSwapRepay = (options: UseCollateralSwapRepayOptions) =
     slippage,
     clearSimulationError,
     getCurrentDebt,
-    includeCowSwap: () => batchEntryCount.value === 0,
+    includeCowSwap: () => !cowSwapForcedOff.value && batchEntryCount.value === 0,
     buildTxPlanForQuote: (quote, _provider, context) => buildRepayPlan(quote, context.account),
     buildGasEstimatePlan: buildBatchAwareGasEstimatePlan,
     prefetchPluginData: (plan, account) => prefetchPluginData(plan, { account }),
--- a/composables/useCowSwapEligibility.ts
+++ b/composables/useCowSwapEligibility.ts
@@ -1,0 +1,22 @@
+import { computed } from 'vue'
+
+/**
+ * Wallet-level CoW Swap eligibility.
+ *
+ * CoW orders require a recoverable ECDSA signature — the SDK's cowExecutor
+ * rejects anything else (`normalizeCowSignature`), and it does so only
+ * AFTER the user has already sent ERC-20 approval transactions. Safe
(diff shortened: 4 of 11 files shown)


composables/borrow/useMultiplyForm.ts
```diff
@@ -167,7 +168,7 @@ export const useMultiplyForm = (options: UseMultiplyFormOptions) => {
   } = useSwapQuotesParallel({
     amountField: 'amountOut',
     compare: 'max',
-    includeCowSwap: () => batchEntryCount.value === 0 && !isMultiplySavingCollateral.value,
+    includeCowSwap: () => !cowSwapForcedOff.value && batchEntryCount.value === 0 && !isMultiplySavingCollateral.value,
```
LeonardEulerXYZ, line 172: When this callable flips to `false`, `useSwapQuotesParallel` removes CoW cards that are already present, but an in-flight CoW request remains on the current sweep generation. After its await completes, the pipeline checks only `guard.isStale(gen)` and then upserts the CoW card, so it resurfaces for the Safe and can be selected until the execution backstop rejects submission. I reproduced this by delaying only the CoW response, flipping the gate, then resolving it: the providers became `['cow', 'other']` instead of `['other']`. Please re-check the live eligibility immediately before accepting/upserting a CoW result (including after async enrichment), or invalidate/abort and restart the sweep, and add that delayed-response regression to `useSwapQuotesParallel.test.ts`. The shared fix should cover multiply, collateral repay, and collateral refinance.

LeonardEulerXYZ (changes requested):
Reviewed exact head `f1923d05b7b5fe79eee0a82201ce27df9f023dee`, stacked directly on approved #797 head `4598b568a60143989b4dabad2a8189959fa917ee`.

The fail-closed Safe eligibility and pre-execution backstop are sound, but one quote-pipeline race still violates the PR's core behavior: a CoW request already in flight when Safe detection flips can resolve after the eviction watcher and reinsert an unusable CoW card. See the inline finding.

The execution backstop still prevents transaction/signature side effects, so this is bounded to stale route presentation and failed submission rather than stranded approvals. It should nevertheless be fixed before merge because Safe users can still be routed back onto the path this PR is meant to remove.

Validation:
- Focused changed/affected suites: 47/47 passed
- Full suite: 1,648 passed, 1 skipped
- `npm run typecheck`: passed
- `npm run lint`: passed with 6 existing warnings in untouched files
- `npm run build`: passed
- `git diff --check`: clean
- Delayed-response adversarial test: failed as described (`['cow', 'other']` after the gate became false; expected `['other']`)
- Cross-repo check: installed SDK 1.2.5 confirms approvals occur before CoW signature normalization, matching the stated threat model

No screenshot attached: the defect is a timing race, and I do not have an approved Safe wallet fixture for a public preview capture. The deterministic delayed-response test is the discriminating evidence here.

LeonardEulerXYZ (approved):
Reviewed exact head `48019526bc6b81c53fdc0fc40fd0261ebd18fa77`, based directly on approved #797 head `4598b568a60143989b4dabad2a8189959fa917ee`.

The prior late-response blocker is fixed at the final acceptance point: `upsertQuote` now re-checks live CoW eligibility after quote retrieval and any asynchronous enrichment, so an in-flight CoW response cannot repopulate the list after the Safe gate flips.

The committed regression is discriminating:
- previous head `f1923d05`: fails with `['cow', 'other']` instead of `['other']`;
- current head `48019526`: passes.

Fresh full-diff review found no additional blockers. Quote generation, provider selection, and all three CoW execution surfaces consume the same fail-closed Safe gate, while execution retains a final pre-side-effect guard.

Validation:
- focused CoW suites: 13/13 passed
- full suite: 1,653/1,653 passed
- typecheck: passed
- lint: passed with six existing warnings in untouched files
- production build: passed
- `git diff --check`: clean
- GitHub checks: green

Approved at this exact head.

LeonardEulerXYZ (approved):
Re-reviewed exact head 60d3e14be5a919bba8502b3048b1f3c005940193. This merge contains the approved execution-scoped detachment fix at bf9eacb54dcc89bbb3b96ea40e7a31db42c03ef7, and the PR-specific Safe/CowSwap behavior remains correct after integration. Verification: clean npm ci; 21 focused tests; full suite 1663 passed / 1 skipped; typecheck, lint, and production build all passed.

LeonardEulerXYZ (changes requested):
The PR-specific CoW work remains clean, and this head correctly carries the slot and operation-map fixes. I cannot approve the exact tree yet because the merged #797 base still has Safe-detachable Earn/Lend `OperationReviewModal` success callbacks that bypass the tracked-execution success marker and navigation suppression. Those flows can still redirect after detachment and warn that a successful proposal did not complete. Please refresh this stack after the complete direct-flow finalization fix lands in #797; no additional CoW-specific change is requested.

LeonardEulerXYZ (changes requested):
The PR-specific CoW work remains clean and this merge contains the eight direct-flow fixes. The exact tree is still blocked by two inherited detachable reward callbacks (`RewardUnlockItem.vue` and `PortfolioSdkRewardItem.vue`) that execute and close without marking tracked success, causing confirmed detached Safe proposals to warn that they did not complete. Please refresh after those final #797 success-attribution gaps are fixed; no CoW-specific change is requested.

LeonardEulerXYZ (changes requested):
The CoW-specific work remains clean. This exact tree still inherits the two reward success tails that call unscoped global `modal.close()` after detached confirmation. Because `close()` pops the current modal, they can close a newer modal the user opened while signatures were pending. Please refresh after #797 gates or scopes those closes; no CoW-specific change is requested.

LeonardEulerXYZ (changes requested):
The CoW-specific work remains clean. This exact tree inherits three remaining unscoped `modal.close()` calls in the pending `RewardUnlockItem.unlock()` callback (two batch rechecks and the changed/missing/unavailable exit). They can still pop a newer modal after detachment. Please refresh after those bounded #797 exits are suppressed or scoped; no CoW-specific change is requested.

LeonardEulerXYZ (approved):
Approved at exact head `6d84739d`.

The CoW-specific work remains clean, and this merge head contains the completed #797 detachment invariant fixes. Safe wallets cannot enter CoW order flows; ordinary-wallet behavior is preserved. GitHub lint, typecheck, and test are green.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `f0035dde5df526768e212c3a117b6684ccf3ab38` since the last approved head `6d84739d`.

The PR-specific Safe/CowSwap behavior remains coherent, and the merged display-summary work correctly surfaces the post-execution ceremony. Requesting changes because the merged-forward wallet-switch change introduces one execution-attribution blocker at this exact head:

**[P1] Bind the success marker to the execution that produced it** — `composables/useSafeExecutionDetachment.ts:31`

After execution A is abandoned on an account/connector change, the slot is cleared and execution B may start. A's old asynchronous success tail can still call the parameterless `markTrackedExecutionSucceeded()`, which mutates the current slot rather than A's record. It can therefore mark B successful. If B later resolves without reaching its own finalize point, its detached continuation emits a false `Safe transaction confirmed` toast.

A discriminating test at this exact composed head—detach A, abandon A, detach B, deliver A's late success marker, then resolve B—fails because the success toast is called once. Please make success attribution execution-bound and retain this overlap regression.

Validation: existing focused detachment/Safe/finalization tests pass (20 tests); hosted lint/typecheck/test are green. The new overlap probe fails and is the reason for this disposition.

PR #799, opened 2026-08-11
feat: bundle migration authorizations + EVC batch into one Safe proposal

kasperpawlowski:
## Summary

Transaction-kind migrations for Safe wallets now submit **authorization grants + the migration EVC batch + revocations as one atomic `wallet_sendCalls` bundle** — a single Safe proposal instead of two to three sequential ones (LITE-323). **Stacked on #797**; retarget to `development` once it merges.

## Why atomicity is also a safety win

Today, if the migration batch fails after a grant landed, the authorization stays live on-chain until the revoke unwind succeeds — and receipt-ambiguous failures leave revokes queued in `pendingRevokes` bookkeeping. In one atomic bundle a failed migration reverts its grants with it, and the revocations ride in the same proposal, so **no standing-authorization window exists on success or failure** and the bundled path needs no unwind bookkeeping at all.

## The key design question: plan correctness before grants mine

The sequential flow is ordered the way it is because `planCrossProtocolMigration` does a **live allowance read** inside the connector and throws when the grant is not yet on-chain. The bundle path instead builds its plan from the SDK's **simulation variant** (`planCrossProtocolMigrationSimulation`): for transaction-kind authorization requests this plan is **byte-identical to the execution plan** (`plan === previewPlan`, a fact the codebase already relies on for calldata previews) — the connector validates the grant's target/spender/amount instead of reading live state. So nothing needs to mine before the bundle is assembled, and the pre-submission blocking simulation runs with the SDK's authorization `stateOverrides` threaded through a new `extraStateOverrides` pass-through on `simulatePreparedPlan` / `runPreparedSimulation` (the same SDK field `useTxBatch` already uses).

## Implementation

- `useEulerTx.executePreparedPlanWithPlainCalls(prepared, { before, after })` — wraps the Safe bundle with plain calls (grants before the plan, revocations after); returns `undefined` when there is no Safe bundle context, keeping the caller in charge of the sequential fallback (which must mine grants first)
- `executePlanAsSafeBundle` gains the `extraCalls` wrapper; all existing guards apply unchanged (context assertion, connector pinning, hash-shaped bundle id, reverted-receipt throw)
- `migrate.vue` (outgoing) and `borrow/swap.vue` (inbound external): a bundle-first branch — simulation-variant plan → prepare (`usePermit2: false`) → blocking simulation with authorization overrides → bundle. `'aborted'` (simulation rejected, nothing on-chain) returns cleanly with no unwind; `'unavailable'` falls through to today's sequential grant flow untouched
- Review display: `buildMigrationAuthorizationTxSteps` gains a `bundled` option — grant/revoke rows drop the "Separate tx" badge when a Safe is connected, matching the `bundledApprovals` precedent

## Test plan

- [x] `executePreparedPlanWithPlainCalls`: exact bundle ordering `[grant, batch, revoke]` with connector pinning + `forceAtomic`; `undefined` for non-Safe connectors and provider-less Safes; reverted bundle throws before finalizing
- [x] `simulatePreparedPlan`: `extraStateOverrides` passed through when present, omitted otherwise
- [x] `buildMigrationAuthorizationTxSteps`: bundled flag flips `isSeparateTx`, default unchanged
- [x] Existing migration cleanup/drift/revoke suites unchanged and green; full suite 1648 passing; typecheck + lint clean
- [ ] Manual with a real Safe: Aave outgoing + inbound migration produce ONE proposal each; simulation rejection leaves nothing on-chain; EOA migrations unchanged
- [ ] Manual: degraded Safe (provider unavailable) falls back to sequential grants

### DIFF
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -1,4 +1,4 @@
-import { getAddress, type Address, type Hash, type Hex, type TransactionReceipt } from 'viem'
+import { getAddress, type Address, type Hash, type Hex, type StateOverride, type TransactionReceipt } from 'viem'
 import type {
   Account,
   CollateralShareSource,
@@ -1184,12 +1184,20 @@ export const useEulerTx = () => {
     })
   }
 
-  const simulatePreparedPlan = async (prepared: TransactionPlanPrepared, stateOverrideOptions?: SimulationStateOverrideOptions) => {
+  const simulatePreparedPlan = async (
+    prepared: TransactionPlanPrepared,
+    stateOverrideOptions?: SimulationStateOverrideOptions,
+    extraStateOverrides?: StateOverride,
+  ) => {
     return profAsync('sdk', 'simulatePreparedTransactionPlan', async () => {
       const sdk = await getEulerSdkForChain(prepared.chainId)
       return sdk.executionService.simulatePreparedTransactionPlan(prepared, {
         stateOverrides: true,
         stateOverrideOptions,
+        // Caller-supplied overrides for state the plan assumes but which is
+        // not on-chain yet (e.g. migration authorizations that will be
+        // granted inside the same Safe bundle).
+        ...(extraStateOverrides?.length ? { extraStateOverrides } : {}),
       })
     })
   }
@@ -1397,11 +1405,13 @@ export const useEulerTx = () => {
    * Submit every plan transaction as one EIP-5792 call bundle. Safe turns
    * the bundle into a single MultiSend proposal, so signers approve once
    * instead of once per transaction (approve + EVC batch collapse into one
-   * proposal). Returns undefined when the plan cannot be bundled (permit2 /
-   * CoW swap items) or when bundling brings no benefit (fewer than two
-   * calls) — callers fall back to sequential execution.
+   * proposal). `extraCalls` wrap the plan inside the same bundle (migration
+   * authorization grants before it, revocations after it). Returns undefined
+   * when the plan cannot be bundled (permit2 / CoW swap items) or when
+   * bundling brings no benefit (fewer than two calls) — callers fall back to
+   * sequential execution.
    */
-  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk }: {
+  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls }: {
     plan: TransactionPlan
     chainId: number
     owner: Address
@@ -1410,15 +1420,25 @@ export const useEulerTx = () => {
     connector: NonNullable<ReturnType<typeof getAccount>['connector']>
     safeWalletProvider: WalletProviderLike
     sdk: PlanEncodingSdk
+    extraCalls?: {
+      before?: readonly PlainTxRequest[]
+      after?: readonly PlainTxRequest[]
+    }
   }) => {
-    let calls
+    let planCalls
     try {
-      calls = transactionPlanToCalls(plan, sdk, chainId)
+      planCalls = transactionPlanToCalls(plan, sdk, chainId)
     }
     catch (err) {
       if (err instanceof PlanNotBundleableError) return undefined
       throw err
     }
+    const toPlanCall = (tx: PlainTxRequest) => ({ to: tx.to, data: tx.data, value: tx.value ?? 0n })
+    const calls = [
+      ...(extraCalls?.before ?? []).map(toPlanCall),
+      ...planCalls,
+      ...(extraCalls?.after ?? []).map(toPlanCall),
+    ]
     if (calls.length < 2) return undefined
 
     const currentAccount = getAccount(config)
@@ -1629,6 +1649,72 @@ export const useEulerTx = () => {
   }
 
   /**
+   * Execute a prepared plan as one Safe call bundle wrapped by extra plain
+   * calls: migration authorization grants before the plan, revocations after
+   * it. Atomicity is the point — a failed migration reverts its grants with
+   * it, and the revocations land in the same proposal, so no standing
+   * authorization ever needs unwind bookkeeping.
+   *
+   * Returns undefined when no Safe bundle context is available (regular
+   * wallet, or a Safe whose provider could not be acquired) — the caller
+   * owns the sequential fallback, which must broadcast the grants and wait
+   * for them to mine before the migration plan is rebuilt.
+   */
+  const executePreparedPlanWithPlainCalls = async (
+    prepared: TransactionPlanPrepared,
+    extraCalls: {
+      before?: readonly PlainTxRequest[]
+      after?: readonly PlainTxRequest[]
+    },
+  ) => {
+    if (isSpyMode.value) {
+      throw new Error('Transactions are disabled in spy mode')
+    }
+    const sdk = await getEulerSdkFresh()
+    const provider = sdk.providerService?.getProvider(prepared.chainId)
+    if (!provider) {
+      throw new Error('No provider available to confirm the transaction')
+    }
+    const connector = getAccount(config).connector
+    const safeWalletProvider = await getSafeWalletProvider(connector)
+    if (!safeWalletProvider || !connector) return undefined
+
+    const preparedOwner = typeof prepared.account === 'string'
+      ? getAddress(prepared.account)
+      : getAddress(prepared.account.owner)
+
+    // Same invariant as executePreparedPlan: an envelope executed for a Safe
+    // never carries permit2.
+    let effectivePrepared = prepared
+    if (hasPermit2Signature(prepared.plan)) {
+      const repairedPlan = await sdk.executionService.resolveRequiredApprovals({
+        plan: prepared.plan,
+        chainId: prepared.chainId,
+        account: preparedOwner,
+        usePermit2: false,
+      })
+      effectivePrepared = { ...prepared, plan: repairedPlan, usePermit2: false }
+    }
+    else if (prepared.usePermit2) {
+      effectivePrepared = { ...prepared, usePermit2: false }
+    }
+
+    const bundled = await executePlanAsSafeBundle({
+      plan: effectivePrepared.plan,
+      chainId: prepared.chainId,
+      owner: preparedOwner,
+      provider,
+      connector,
+      safeWalletProvider,
+      sdk,
+      extraCalls,
+    })
+    if (!bundled) return undefined
+    finalizeExecution(bundled)
+    return bundled
+  }
+
+  /**
    * Attach a fresh snapshot of a sub-account onto a pre-loaded Account. Call
    * once per form interaction (after the receiver/borrow sub-account is known
    * and before fanning out per-quote plan builds) to amortise what would
@@ -1688,6 +1774,7 @@ export const useEulerTx = () => {
     estimateGasForPlan,
     prefetchPluginData,
     simulatePreparedPlan,
+    executePreparedPlanWithPlainCalls,
     executePlan,
     executePreparedPlan,
   }
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -1433,6 +1433,16 @@ export const useEulerTx = () => {
       if (err instanceof PlanNotBundleableError) return undefined
       throw err
     }
+    if (planCalls.length === 0) {
+      // Wrapper calls must never satisfy the bundle on their own: submitting
+      // [grant, revoke] around an empty plan would finalize a no-op
+      // migration as success. Without wrappers an empty plan simply has
+      // nothing to bundle and falls back to sequential execution.
+      if (extraCalls?.before?.length || extraCalls?.after?.length) {
+        throw new Error('Transaction plan produced no calls to bundle')
+      }
+      return undefined
+    }
     const toPlanCall = (tx: PlainTxRequest) => ({ to: tx.to, data: tx.data, value: tx.value ?? 0n })
     const calls = [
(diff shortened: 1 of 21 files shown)


composables/useEulerTx.ts
```diff
@@ -1410,15 +1420,25 @@ export const useEulerTx = () => {
...
       if (err instanceof PlanNotBundleableError) return undefined
       throw err
     }
+    const toPlanCall = (tx: PlainTxRequest) => ({ to: tx.to, data: tx.data, value: tx.value ?? 0n })
+    const calls = [
+      ...(extraCalls?.before ?? []).map(toPlanCall),
+      ...planCalls,
+      ...(extraCalls?.after ?? []).map(toPlanCall),
+    ]
     if (calls.length < 2) return undefined
```
LeonardEulerXYZ: `calls.length` includes `extraCalls.before` and `extraCalls.after`, so it no longer proves that the migration plan contributed a call. A reviewer probe passed an empty prepared plan plus one grant and one revoke; this branch submitted `[grant, revoke]`, finalized the successful Safe receipt, and returned a successful result with `plan: []`. Both migration callers then take their success close/refresh/redirect path despite no migration call being present. Please guard `planCalls.length === 0` before composing the wrappers (ideally throw for this migration path) and cover it with a regression.

LeonardEulerXYZ (changes requested):
Reviewed exact head `827b5385a14614fedafc561807862a900106ff61`, based directly on approved #797 head `4598b568a60143989b4dabad2a8189959fa917ee`. This is a sibling of #798, not stacked on it.

The main atomic path is otherwise well-formed: transaction-form SDK grants are encoded before the prepared migration call(s), restorations are reversed after them, `forceAtomic` and connector pinning are preserved, simulation receives the SDK authorization overrides, and unavailable Safe-provider paths retain the mined sequential fallback.

One success-invariant blocker remains: `executePlanAsSafeBundle` checks the combined wrapper + plan call count, not that the migration plan itself encoded at least one call. A temporary adversarial test passed an empty prepared plan with one grant and one revoke. The helper submitted `[grant, revoke]`, received a successful Safe receipt, finalized it, and returned `{ plan: [], hashes, receipts }` instead of rejecting or falling back. The migration pages would then run their normal success close/refresh/redirect path even though no migration call was included.

Please require `planCalls.length > 0` before adding wrapper calls (preferably fail loudly for this migration-only path), and add the empty-plan regression. Wrapper calls may surround a migration; they must not substitute for one.

Validation:
- focused Safe/migration suites: 46/46 passed
- full suite: 1,648 passed, 1 skipped
- typecheck: passed
- lint: passed with six existing warnings in untouched files
- production build: passed
- `git diff --check`: clean
- GitHub checks: green

Changes requested at this exact head.

LeonardEulerXYZ (approved):
Re-reviewed exact head `ce4ef5e49cec45fca753c103d751f2312cd40c35`, based directly on approved #797 head `4598b568a60143989b4dabad2a8189959fa917ee`.

The empty-plan blocker is fixed with the correct invariant and failure mode:
- `planCalls.length === 0` is checked before wrapper composition;
- wrapper calls present: execution fails loudly with `Transaction plan produced no calls to bundle`;
- no wrapper calls: the existing `undefined` / sequential-fallback semantics remain unchanged.

The committed regression reproduces the original adversarial case and asserts no `sendCalls` and no prepared sequential execution. It is discriminating: deleting the new guard makes it fail by resolving successfully with `{ plan: [], hashes, receipts }`, exactly the prior false-success path. A separate reviewer probe confirmed the wrapper-free empty-plan branch still returns `undefined` without submission.

Fresh delta review found no new issues. The broader atomic ordering, simulation overrides, connector pinning, receipt-status checks, Permit2 repair, and provider-unavailable sequential fallback remain intact.

Validation:
- full suite: 1,653/1,653 passed
- typecheck: passed
- lint: passed with six existing warnings in untouched files
- production build: passed
- `git diff --check`: clean
- GitHub checks: green

Approved at this exact head.

LeonardEulerXYZ (changes requested):
Re-reviewed the delta since the prior approved head and verified exact head `417e24ca71859f843041d9aada9bff6f0bf3e40f`.

There are three blocking correctness gaps:

1. **Detached migration failures can produce a false confirmation toast.** `OperationReviewModal` detaches the raw `onConfirm()` promise, while `useSafeExecutionDetachment.detach()` treats any resolution as confirmation. Both migration handlers in this PR catch execution errors, call `showError`, and return normally. Closing the modal while a Safe proposal is pending therefore turns a later failed proposal into a resolved promise and emits `Safe transaction confirmed`. Please propagate a structured success/failure result or reject on failed execution, and add a regression for the detached failure path.

2. **The Safe migration paths still redirect after detachment.** The new successful bundle paths eventually call `modal.close()` and `router.replace(...)` directly. They do not pass through `useTxFinalization`, which is the only place the new suppression guard is consulted. Thus closing the review and navigating elsewhere does not prevent a later Safe migration confirmation from redirecting the user back to the migrated position. The global count is also ambiguous under concurrency: detached A completing while attended B is pending sees an attached count and is allowed to redirect. Please make detachment execution-scoped and apply it to these direct migration completion paths, with a detached-A/attached-B interleaving covered.

3. **The inherited ToS-layer normalization leaves failure attribution shifted.** SDK `operationIndex` includes the prepended loose ToS operation, but the patch strips only the account/vault/wallet state layers. A revert in cart operation 0 arrives as index 1, marks no row in a one-entry cart, and leaves `hasFailedOps` false; later operations mark the following row. `executeBatch()` can then reach prerequisite broadcasts before its final gas estimate rejects. Normalize failure indices by the same leading-plugin count and keep plugin-only failures at batch level.

The committed paths are otherwise green locally (104 focused tests and Nuxt typecheck), but those tests do not cover these failure/interleaving cases.

LeonardEulerXYZ (changes requested):
The provider-unavailable case now aborts correctly, but the outbound migration still has a live-classification fallthrough that can silently restore the multi-proposal ceremony.

The review snapshot records `useSignatures = false`, so it promised the atomic Safe path. At confirmation, `executeOutboundMigrationAsSafeBundle()` checks the live `isSafeWallet`; if that value is false it returns `not-safe`. The caller handles only `aborted` and `executed`, then falls through into the sequential authorization grant → migration → revoke path. This is reachable during a connector transition because `useSafeWallet` deliberately sets `isSafeWallet = false` while the new probe is unresolved, and the execution guard binds account/chain rather than the reviewed connector/mode.

Please make the reviewed mode authoritative: when `useSignatures === false`, any inability to execute the reviewed atomic Safe proposal — including false/unresolved live classification or connector drift — must abort before `authorizationRequest` or any sequential send. The inbound path already has the required fail-closed shape. Please add the corresponding outbound regression: prepare/review as Safe, make live Safe classification false or unresolved at confirm with account/chain unchanged, and assert no grant, migration, or revoke transaction is submitted.

I independently reran clean install, 89 focused tests, the full 1663-test suite (1 skipped), typecheck, lint, and build; they pass, but none exercise this transition.

LeonardEulerXYZ (changes requested):
The execution-side `bundledReview` latch fixes the previously reported inbound/outbound fallthrough, but review/copy presentation is still keyed from live Safe state rather than that latch.

Both modal openings call `buildCalldataWrapCalls(...)` / `buildInboundExternalCalldataWrapCalls(...)`, and those helpers still consult live `isSafeWallet.value`. If Safe detection changes between preview creation and modal opening:

- `bundledReview === true` but live Safe is false: the review/copy omits grant/revoke wrappers, while confirmation later executes the latched atomic bundle;
- `bundledReview === false` but live Safe is true: the review/copy includes atomic wrappers, while confirmation preserves the latched sequential mode.

Make the latched preview mode authoritative for presentation too: pass `preview.bundledReview` into the wrapper builder (rather than reading live wallet state), and add page-level regressions that vary live detection after preview creation while asserting copied/displayed calls and submitted ceremony remain identical.

This exact tree also inherits #797’s still-unfixed direct Earn/Lend finalization callbacks, so it must be refreshed after that shared blocker is closed.

Verification on this exact head: clean install, 94 focused tests, and typecheck passed. No test currently references `bundledReview` or the wallet-transition review/copy cases.

LeonardEulerXYZ (changes requested):
The migration-specific fix is correct: `bundledReview` now consistently drives step rows, copied wrapper calls, and confirmation execution, and batch rows correctly declare standalone prerequisites. No remaining migration-mode mismatch found.

This exact tree is still blocked by the two inherited #797 reward callbacks (`RewardUnlockItem.vue` and `PortfolioSdkRewardItem.vue`) that execute through `OperationReviewModal` and close without marking tracked success. Please refresh after that shared success-attribution fix lands; no further #799-specific change is requested.

LeonardEulerXYZ (changes requested):
The migration-mode work remains clean. This exact tree is still blocked only by the inherited reward success tails: after detached confirmation they call unscoped global `modal.close()`, whose implementation pops whichever modal is currently open. Please refresh after #797 suppresses or scopes those closes; no further #799-specific change is requested.

LeonardEulerXYZ (changes requested):
The migration-specific work remains clean. This exact tree inherits three remaining unscoped `modal.close()` calls in the pending `RewardUnlockItem.unlock()` callback (two batch rechecks and the changed/missing/unavailable exit). They can still pop a newer modal after detachment. Please refresh after those bounded #797 exits are suppressed or scoped; no further #799-specific change is requested.

LeonardEulerXYZ (approved):
Approved at exact head `06659865`.

Migration review presentation, copied calldata, and confirmation execution share the same latched ceremony mode in both directions; batch prerequisites are honestly represented as standalone; and this merge head contains the completed #797 detachment invariant fixes. GitHub lint, typecheck, and test are green.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `11c4a3d22fc869b50c1104287d8c1d7909cc9764` since reviewed head `06659865a2ae9ff7ef18e57585e312502edf1238`, including the full composed stack.

The ordinary bundled path has the right grant → merged plan → reverse revoke ordering, and its non-empty prerequisite branch fails closed when the Safe provider disappears. Two blockers remain:

**[P1] Keep success attribution bound to the originating execution** — `composables/useSafeExecutionDetachment.ts:31`

The inherited wallet-switch fix clears abandoned execution A from the module slot, allowing execution B to start. A can later reach the global, parameterless `markTrackedExecutionSucceeded()` and mark B's record. A focused overlap probe—detach A, abandon it, detach B, deliver A's late success marker, then resolve B without B finalizing—fails because B emits `Safe transaction confirmed`. The success marker needs an execution identity/handle so stale continuations cannot mutate a replacement record.

**[P1] Do not bypass the fail-closed Safe path when fresh grants are empty** — `composables/useTxBatch.ts:2554`

The modal derives the one-proposal ceremony from `willBundlePrerequisites`. At execution, if `collectBundledExecution()` returns no grant/revoke calls because the grants became live since add-time, this branch falls through to `executePreparedPlan()`. That is not guaranteed to remain one proposal: if Safe provider acquisition is degraded at that point, `executePreparedPlan()` recognizes the Safe connector but continues through `executePreparedTransactionPlan`, i.e. the sequential fallback. The user reviewed one proposal, but execution can silently become multiple submissions.

The existing test codifies this fall-through. I replaced it with the required invariant—call `executePreparedPlanWithPlainCalls(prepared, { before: [], after: [] })`, fail if it returns `undefined`, never call the classic executor—and the test fails because the fail-closed wrapper is never called. Empty calls should still take the same provider-bound Safe path (or perform an equivalent provider check) before honoring the reviewed ceremony.

Validation at the exact head: 108 focused tests pass; typecheck and changed-file lint pass; hosted lint/typecheck/test are green. Both additional discriminating probes fail as described.

LeonardEulerXYZ (changes requested):
Supplemental exact-head review at `11c4a3d22fc869b50c1104287d8c1d7909cc9764` after an independent adversarial pass. The previously requested changes remain valid; two additional reviewed-vs-executed parity blockers are concrete:

**[P1] Latch or revalidate the reviewed ceremony before broadcasting** — `composables/useTxBatch.ts:2432`, `components/BatchReviewModal.vue:441`

`willBundlePrerequisites` is live and drives the modal's “bundled in proposal” description. `handleExecute()` does not latch that reviewed mode; `executeBatch()` reads the predicate again. If Safe classification or connector mode changes after review, execution can fall through to `sendExecutionPrerequisites()` and broadcast standalone grants. Account/chain checks do not bind connector identity or the reviewed bundled/sequential mode. A connector/mode transition must either invalidate/reopen review or fail before the first broadcast; add a Safe → non-Safe/degraded transition test between review and execution.

**[P1] Review and copy the same fresh wrapper calls that execution submits** — `components/BatchReviewModal.vue:344`, `pages/position/[number]/migrate.vue:1096`, `pages/position/[number]/borrow/swap.vue:3612`

The modal rows and copied calldata come from add-time review/prepared state. At confirmation, both migration builders re-fetch authorization and build fresh grant/revoke calls. Concrete sequence: authorization is live at add time, so the review and copied calldata contain no wrappers; it is revoked before confirmation; execution then submits a newly discovered grant and revoke around a freshly built plan. The reverse transition leaves stale rows while execution omits the calls.

The exact fresh payload therefore differs from both the displayed ceremony and “Copy calldata”; it also bypasses an exact final gas estimate/simulation. Either freeze the reviewed authorization payload with appropriate expiry/state guards, or rebuild and present the exact fresh grant → plan → revoke payload before allowing confirmation. Please cover live → revoked and revoked → live transitions, asserting displayed rows, copied calldata, and submitted calls are equivalent.

Independent validation found no additional issue in grant ordering, reverse-entry revoke ordering, atomic revert behavior, or the mixed-cart all-or-nothing predicate.

PR #802, opened 2026-08-11
fix: batch simulation dies with 'not loaded' when ToS registration rides in the batch

kasperpawlowski:
## Summary

Fixes the "Batch simulation not loaded" failure reported while testing the Safe stack (LITE-326) — but the bug is **pre-existing and not Safe-specific**: it hits any account whose ToS acceptance hasn't been registered on-chain yet whose first action is a batch. Fresh Safes reproduce it reliably because every Safe is a new account; long-standing EOA testers never see it because their registration landed long ago.

## Root cause

The lite ToS plugin prepends the on-chain `signTermsOfUse` registration as a **loose `EVCBatchItem`** ahead of the entry operations. The SDK's layered batch simulation emits one account layer per top-level batch unit, so the loose item produces its own layer: `simulatedAccounts.length = entries + 2`. The version guard in `useTxBatch` only catches *fewer* layers than expected; `getCurrentFinalLayer` requires **exactly** `entries + 1`, so the final-layer lookup returns undefined forever and `awaitFinalPlanningLayer` exhausts its 8 attempts into the generic error.

## Fix

`stripLeadingPluginLayers`: compute `extra = simAccounts − (plans + 1)` and drop the plugin layers between the pre-batch snapshot (kept as layer 0) and the first entry's layer — applied in lockstep to the account, vault (`simulatedVaultsLayers`, pre-normalization), and wallet (`simulatedWalletBalances`) layer arrays so all layer indices keep mapping to cart entries. Failure attribution is deliberately untouched: `operationIndex` semantics for loose items are SDK-internal and the ToS registration cannot realistically fail; noted in a comment.

## Test plan

- [x] Regression test: one entry + ToS-extra simulation shape (`[base, tos, afterOp]`) → 2 layers, no simError, active layer settles (previously: retry exhaustion)
- [x] Unit tests for `stripLeadingPluginLayers` (base-inclusive, base-less, no-op)
- [x] Full suite 1616 passing; typecheck + lint clean
- [ ] Manual: fresh account (unsigned ToS) + supply-to-Earn + withdraw batch simulates and executes

## Rollout

Merges to `development` directly, then gets merged forward into #797 → #798/#799 so Safe testing works on those heads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction batch simulation when plugin operations add extra layers.
  * Preserved base account, vault, and wallet states while keeping simulation results aligned with batch entries.
  * Prevented simulation errors and ensured final batch states are calculated correctly.
  * Improved error reporting by identifying the affected batch entry, while reporting plugin failures at the batch level.

### DIFF
--- a/composables/useTxBatch.ts
+++ b/composables/useTxBatch.ts
@@ -534,6 +534,28 @@ export const normalizeSimulatedVaultLayers = <T>(
   return null
 }
 
+/**
+ * Strip layers produced by plugin-prepended loose batch items (e.g. the lite
+ * ToS registration for an account whose acceptance hasn't landed on-chain).
+ * The layered simulation emits one layer per top-level batch unit, so those
+ * items surface as extra layers between the pre-batch snapshot and the first
+ * entry's layer, breaking the layer[i] ↔ entry[i-1] contract that
+ * getCurrentFinalLayer enforces (`layers.length === entries.length + 1`).
+ *
+ * `hasBaseLayer` arrays keep index 0 (the pre-batch snapshot) and drop the
+ * following `extra` plugin layers; base-less arrays drop the first `extra`.
+ */
+export const stripLeadingPluginLayers = <T>(
+  layerArray: T[],
+  extra: number,
+  hasBaseLayer: boolean,
+): T[] => {
+  if (extra <= 0) return layerArray
+  return hasBaseLayer
+    ? [...layerArray.slice(0, 1), ...layerArray.slice(1 + extra)]
+    : layerArray.slice(extra)
+}
+
 type BatchSimulationWalletBalances = {
   simulatedWalletBalances?: Record<string, bigint>[]
 }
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (changes requested):
Reviewed exact head `fdf7bcfba39b3cb569c9d40e2310223895082f1e`.

**Blocking — shift failure attribution together with the simulation layers**

`stripLeadingPluginLayers` removes the prepended ToS operation from the account/vault/wallet layer arrays, but `failedBatchItems.operationIndex` is still consumed unchanged at `composables/useTxBatch.ts:1942-1945`. The SDK numbers operations across the whole merged plan, including loose items, so with `[ToS, cartOp0]` a revert in `cartOp0` arrives as `operationIndex: 1`. After this patch the rendered cart has only entry index 0, so that failure marks no layer. With multiple entries it marks the following row instead.

This is also an execution-gating issue, not only display: `hasFailedOps` is derived from those layer flags. When the only cart operation fails after a prepended ToS item, `executeBatch()` can pass its early guard and reach `sendExecutionPrerequisites()` before the later gas estimate catches the revert. That unnecessarily broadcasts migration grants and relies on best-effort revoke cleanup.

Please normalize `operationIndex` by the same detected leading-plugin count (and keep plugin-only failures as batch-level failures rather than assigning them to an arbitrary cart row), with regressions for failures in the first and later cart operations.

I verified this with an adversarial test using three account snapshots `[base, afterToS, afterCartOp0]` and a failed item at `operationIndex: 1`: the normalized two-layer result reports `layers[1].failed === false` instead of `true`. The committed `useTxBatch` suite still passes 44/44 and typecheck passes, so the current coverage does not exercise this shifted failure axis.

LeonardEulerXYZ (approved):
Re-reviewed the complete fix delta from fdf7bcfba39b3cb569c9d40e2310223895082f1e at exact head aaca08d0a3db5eb17be8628ba205af3bfbd4189d. The explicit operation-to-entry map matches SDK 1.2.5 collectOperations semantics: loose batch items each contribute one operation, named grouped items contribute one operation, plugin operations remain an unmapped prefix, and each entry selects the layer after its final operation. Failed plugin operations fail the batch closed; failed cart operations resolve to the correct row, including multi-operation entries. The original ToS-prefix adversarial case, multi-operation mapping, and plugin-failure path are now discriminating regressions. Verification: clean npm ci; 50 focused batch tests; full suite 1622 passed / 1 skipped; typecheck, lint, and production build all passed.

Seranged (approved):

PR #803, opened 2026-08-12
fix: discover Aave supply-only migrations

Seranged:
## Summary

- Discover Aave supplies even when they are not enabled as collateral.
- Keep non-collateral supplies separate from debt-backed migration positions.

## Changes

- Scan aToken balances across every configured Aave reserve before identifying supplied assets.
- Read debt balances only for reserves marked as borrowed, then limit metadata and pricing calls to active assets.
- Add regression coverage for configuration-zero supplies and mixed supply/debt accounts.

## Test plan

- [x] `npm run test:run -- tests/composables/useExternalMigrationPositions.test.ts` (11 tests)
- [x] `npm run test:run` (1,618 tests)
- [x] Targeted ESLint on both changed files
- [x] `npm run build`
- [x] Local Base smoke: `0x7FD92603BEDB13fe74e4A60eb2738d86680A3d25` shows the 4.000001 USDC Aave supply and 1 USDC Morpho position
- [ ] `npm run typecheck` currently fails on unrelated SDK/type drift already present on `development` (activity event types, `eulerInterfacesBranch`, and `unitOfAccountValuation`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Aave position discovery to include supplied assets even when collateral is disabled.
  - Excluded zero-balance supplies and debts from detected positions.
  - Ensured debt candidates use only assets with positive borrowed balances.
  - Preserved valid positions when unrelated reserve, metadata, or balance data is unavailable.
  - Continued to surface errors for configured reserve data failures.
  - Kept non-collateral supplies eligible for supply-only migration candidates.

- **Tests**
  - Added coverage for reserve-read failures, supply-only positions, and separating non-collateral supplies from debt-backed positions.

### DIFF
--- a/composables/useExternalMigrationPositions.ts
+++ b/composables/useExternalMigrationPositions.ts
@@ -403,9 +403,9 @@ const readContractsAllowFailure = async (
   }
 }
 
-// Failed reads must reject the whole Aave fetch (mirroring the Morpho fetch)
-// rather than decode as "no balance" — otherwise an RPC blip silently makes a
-// real position disappear from discovery with zero observability.
+// Failed root and candidate reads reject the whole Aave fetch rather than
+// decode as "no balance". Individual unconfigured reserve probes may be skipped
+// so an unrelated reserve failure cannot discard already-confirmed positions.
 const requireReadResult = (result: ContractReadResult | undefined, label: string): unknown => {
   if (result?.status === 'success') return result.result
   const cause = result?.status === 'failure' ? result.error : undefined
@@ -686,9 +686,27 @@ export const useExternalMigrationPositions = (options: {
       'externalMigration/aaveReserveMulticall',
     )
     const reserveTokensByAsset = new Map<Address, AaveReserveTokens>()
+    const configuredAssets = new Set(
+      [...collateralAssets, ...debtAssets].map(asset => asset.toLowerCase()),
+    )
     reserves.forEach((asset, index) => {
-      const tokens = parseAaveReserveTokens(requireReadResult(reserveDataResults[index], `getReserveData(${asset})`))
-      if (tokens) reserveTokensByAsset.set(asset, tokens)
+      const result = reserveDataResults[index]
+      const isConfiguredAsset = configuredAssets.has(asset.toLowerCase())
+      if (result?.status !== 'success') {
+        if (isConfiguredAsset) requireReadResult(result, `getReserveData(${asset})`)
+        const cause = result?.status === 'failure' ? result.error : new Error('Aave reserve data result missing')
+        logWarn('externalMigration/aaveReserveDataSkipped', cause, { data: { asset } })
+        return
+      }
+
+      const tokens = parseAaveReserveTokens(result.result)
+      if (!tokens) {
+        const invalidResult = new Error(`Aave discovery read returned invalid reserve data: ${asset}`)
+        if (isConfiguredAsset) throw invalidResult
+        logWarn('externalMigration/aaveReserveDataSkipped', invalidResult, { data: { asset } })
+        return
+      }
+      reserveTokensByAsset.set(asset, tokens)
     })
 
     const balanceReadEntries: { kind: 'supply' | 'variableDebt' | 'stableDebt', asset: Address }[] = [
--- a/composables/useExternalMigrationPositions.ts
+++ b/composables/useExternalMigrationPositions.ts
@@ -403,9 +403,9 @@ const readContractsAllowFailure = async (
   }
 }
 
-// Failed root and candidate reads reject the whole Aave fetch rather than
-// decode as "no balance". Unconfigured reserve-data and supply-balance probes
-// may be skipped so an unrelated failure cannot discard confirmed positions.
+// Failed root and configured-position reads reject the whole Aave fetch rather
+// than decode as "no balance". Unconfigured reserve probes and per-asset
+// metadata failures are isolated so they cannot discard confirmed positions.
 const requireReadResult = (result: ContractReadResult | undefined, label: string): unknown => {
   if (result?.status === 'success') return result.result
   const cause = result?.status === 'failure' ? result.error : undefined
@@ -793,14 +793,26 @@ export const useExternalMigrationPositions = (options: {
     )
     const assetsByAddress = new Map<Address, ExternalMigrationAsset>()
     activeAssets.forEach((asset, index) => {
+      const symbolResult = metadataResults[index * 2]
+      const decimalsResult = metadataResults[index * 2 + 1]
+      if (symbolResult?.status !== 'success' || decimalsResult?.status !== 'success') {
+        const field = symbolResult?.status !== 'success' ? 'symbol' : 'decimals'
+        const result = field === 'symbol' ? symbolResult : decimalsResult
+        const cause = result?.status === 'failure' ? result.error : new Error(`Aave ${field} result missing`)
+        logWarn('externalMigration/aaveMetadataSkipped', cause, {
+          data: { asset, field },
+        })
+        return
+      }
+
       assetsByAddress.set(asset, parseAaveAsset(
         asset,
-        requireReadResult(metadataResults[index * 2], `symbol(${asset})`),
-        requireReadResult(metadataResults[index * 2 + 1], `decimals(${asset})`),
+        symbolResult.result,
+        decimalsResult.result,
       ))
     })
 
-    const usdPricesByAsset = await fetchAaveAssetUsdPrices(targetChainId, activeAssets)
+    const usdPricesByAsset = await fetchAaveAssetUsdPrices(targetChainId, [...assetsByAddress.keys()])
 
     const candidates: AaveMigrationCandidate[] = []
     for (const collateralAsset of suppliedAssets) {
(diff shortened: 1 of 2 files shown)


LeonardEulerXYZ (approved):
## Review summary

Approved at exact head `11e272dd312ec3c7956548592c13410c68510038`.

The implementation fixes the underlying Aave invariant correctly: the user-configuration bitmap records borrow/collateral flags, not every positive supply, so discovery now reads aToken balances across the pool reserve list and only uses the bitmap to decide whether a supplied asset may be paired with live debt. Non-collateral supplies remain independent supply-only candidates.

### Validation

- `npm exec vitest run tests/composables/useExternalMigrationPositions.test.ts` — 11/11 passed
- `npm exec eslint composables/useExternalMigrationPositions.ts tests/composables/useExternalMigrationPositions.test.ts` — passed
- `npm run build` — passed (existing build warnings only)
- GitHub CI at this head — lint, typecheck, build/test all passed
- Railway PR preview — deployed successfully
- Headed Chromium preview smoke on Base using `0x7FD92603BEDB13fe74e4A60eb2738d86680A3d25`:
  - Aave V3 USDC supply rendered as `$4.00` / ~4 USDC
  - the separate Morpho Vaults USDC position remained present
  - Aave row routed to `/position/external/borrow/swap` with the supply-only source id and rendered `USDC supply`, `Lend vault`, and `$4.00` collateral value
- Direct Base chain check:
  - Aave pool user configuration is `0`
  - the account nevertheless holds approximately 4 USDC in the configured aToken
  - the pool's listed reserve-data calls resolve successfully

### Risk review

- No dependency, lockfile, workflow, server, signing, or transaction-building changes.
- Multicall/result indexing remains aligned when reserve token parsing skips an entry.
- Metadata and price reads are narrowed to positive supplied/borrowed assets.
- The selected position is re-read by the SDK before migration preparation; discovery does not become authoritative for transferred amounts.

No blocking findings.

dglowinski (dismissed):

composables/useExternalMigrationPositions.ts
```diff
@@ -688,38 +686,13 @@ export const useExternalMigrationPositions = (options: {
       'externalMigration/aaveReserveMulticall',
     )
     const reserveTokensByAsset = new Map<Address, AaveReserveTokens>()
-    activeAssets.forEach((asset, index) => {
+    reserves.forEach((asset, index) => {
       const tokens = parseAaveReserveTokens(requireReadResult(reserveDataResults[index], `getReserveData(${asset})`))
```
LeonardEulerXYZ: **Blocking:** `requireReadResult` over every reserve widens one reserve failure into a total Aave-discovery failure. I reproduced this with a valid WETH supply and an unrelated USDC `getReserveData` failure: `development` still returns the WETH Aave row, while this head drops every Aave row. If Morpho returns anything, `load()` then leaves `error` empty, so the UI silently presents an incomplete scan as successful. Please tolerate/skip failures for irrelevant reserves (while preserving fail-closed handling for candidate reads), or surface connector-level partial failure, and cover the mixed Aave-failure/Morpho-success case.

LeonardEulerXYZ (changes requested):
## Follow-up adversarial review

Superseding my earlier approval after the delayed adversarial pass produced a concrete regression.

A single failure while reading an otherwise irrelevant reserve now aborts the entire Aave discovery. Because connector results are combined with `Promise.allSettled`, any valid Morpho row then suppresses the error and the UI reports a successful scan containing only Morpho positions. The user is not told that all Aave positions were omitted.

I reproduced this at exact head with a collateral-enabled WETH supply plus an unrelated USDC reserve-data failure:

- `development`: returns both the valid Aave WETH row and Morpho row
- PR head: returns only the Morpho row; `error === ''`; `hasLoaded === true`

Please isolate failures for reserves that are not needed for a candidate (or otherwise surface connector-level partial failure), and add regression coverage for an unrelated reserve failure plus mixed Aave/Morpho success. The duplicate-reserve hardening and additional edge cases from the adversarial pass are non-blocking; this silent omission is the blocker.

composables/useExternalMigrationPositions.ts
```diff
@@ -746,14 +737,14 @@ export const useExternalMigrationPositions = (options: {
     )
 
     let resultIndex = 0
-    const collateralAmounts = new Map<Address, bigint>()
+    const supplyAmounts = new Map<Address, bigint>()
     const debtAmounts = new Map<Address, { variableDebt: bigint, stableDebt: bigint }>()
     for (const entry of balanceReadEntries) {
       if (!reserveTokensByAsset.has(entry.asset)) continue
       const amount = parseBigIntAmount(requireReadResult(balanceResults[resultIndex], `balanceOf(${entry.kind}:${entry.asset})`))
```
LeonardEulerXYZ: **Blocking:** Reserve-data failures are now isolated for unconfigured reserves, but balance failures are still unconditionally fatal here. Because `balanceReadEntries` contains a supply read for every reserve, one failed aToken `balanceOf` on an unrelated unconfigured reserve rejects all Aave discovery. With Morpho succeeding, the caller shows only Morpho and no error. I reproduced this with valid configured WETH supply plus a failed unconfigured USDC balance: the same fixture returns WETH + Morpho on `development`, but only Morpho with `error === ''` on this head. Please skip/record failed supply probes for unconfigured reserves, while keeping configured/candidate reads fail-closed, and add this balance-stage mixed-source test.

LeonardEulerXYZ (changes requested):
## Fresh-head re-review

The pushed commit fixes the exact `getReserveData` reproducer, and its new tests pass. However, the same PR-expanded failure domain remains at the next per-reserve stage, so I cannot renew approval yet.

With configured WETH collateral/supply and an unrelated unconfigured USDC reserve:

- WETH reserve data and aToken balance succeed; the wallet has a valid WETH Aave row.
- USDC reserve data succeeds, but its aToken `balanceOf` fails.
- `development` does not query that unrelated balance and returns both the WETH Aave and Morpho rows.
- Current head queries it through `...reserves.map(...)`; `requireReadResult` rejects the entire Aave fetch, while `Promise.allSettled` retains Morpho and leaves `error === ''`. The UI therefore again looks successful while omitting all Aave positions.

Please apply the same relevance-aware isolation to per-reserve supply balance failures (while retaining fail-closed reads for bitmap-configured/candidate assets), and add the mixed Aave/Morpho regression fixture at the balance stage.

Validation at this head: committed focused suite 13/13 passed; changed-file ESLint, typecheck, build, CI, and preview smoke passed. The independent complement probe passes on `development` and fails on this head for the omission above.

LeonardEulerXYZ (changes requested):
## Fresh-head re-review

The pushed commit closes the unrelated `balanceOf` failure seam, and its new regression test discriminates correctly against the previous head. One final PR-expanded strict stage still has the same silent all-Aave omission: token metadata.

**Exact-head discriminator**

I added a disposable test with:

- configured WETH collateral and a valid WETH aToken balance;
- unconfigured USDC with a positive aToken balance;
- USDC `symbol()` failure; and
- one valid Morpho position.

Results:

- `development` (`c00800ba`): passes; WETH Aave + Morpho are retained.
- PR head (`300ea119`): fails; the WETH Aave row is removed and only Morpho remains, while the aggregate composable still resolves with `error === ''`.

So `requireReadResult` over all metadata calls still lets a metadata failure for one newly discovered unconfigured supply discard an independently confirmed configured Aave position. This can silently hide a user's migratable position.

Please isolate metadata failures per discovered asset/leg (or otherwise preserve independently confirmed Aave rows while dropping/surfacing only the failed asset), retain fail-closed behavior where that metadata is required for the same candidate, and add the mixed Aave/Morpho regression above. Reserve-data and balance failure isolation now look correct.

Other fresh-head validation passed: committed focused suite 14/14; changed-file ESLint; typecheck; production build; GitHub CI; Railway deployment/HTTP preview and normal migration-page rendering.

LeonardEulerXYZ (approved):
## Fresh-head approval

Approved at exact head `9146147d4bc3b898937bebc4c34f5dbbb5255773`, superseding my prior changes request.

The metadata-stage blocker is resolved at the correct unit: failed metadata drops/logs only that asset, independently complete Aave and Morpho candidates survive, and price lookup is restricted to assets with valid metadata. The earlier reserve-data and balance-stage fixes remain intact.

### Discrimination and validation

- Committed focused suite: 15/15 passed.
- Metadata mixed-source regression passes on this head and fails on prior head `300ea119` for the original silent Aave omission.
- Reviewer-only price-call assertion passed: WETH remained priced while failed-metadata USDC was never sent to `fetchAssetUsdPriceByAddress`; reviewer mutation was then restored and the exact-head tree verified clean.
- Full local suite: 1,618 passed, 1 skipped, 0 failed in this clean checkout.
- Local typecheck: passed; the reported SDK/activity drift did not reproduce in this exact-head worktree.
- Production build: passed with existing warnings only.
- GitHub CI: lint, typecheck, test, CodeRabbit, and Railway deployment all green.
- Railway preview HTTP check: 200.
- Browser smoke on Base using `0x7FD92603BEDB13fe74e4A60eb2738d86680A3d25`: Aave V3 USDC `$4.00` and Morpho Vaults USDC `$1.00` both rendered.

### Review scope

- Full base-to-head diff re-read across the composable and regression tests.
- No dependency, lockfile, workflow, server, signing, or transaction-building changes.
- No remaining blocking findings.

dglowinski (approved):

PR #795, opened 2026-08-10
feat: badge unknown risk-manager vaults in discovery graph

kasperpawlowski:
## Summary

The expanded discovery graph already badges unknown *external* collateral (via `market.unknownCollateral`), but product member vaults with an unverified governor showed no indicator — even though the pair cards below render the red "Unknown" risk-manager pill for them. A market like Alterscope LRT could have every vault flagged "Unknown" in the Lend/Borrow cards while the graph looked perfectly clean.

This adds a per-node check in `DiscoveryMarketGraph.vue` that runs the same verification predicates the cards use:

- `isVaultGovernorVerified` for EVK vaults
- `isSecuritizeGovernorVerified` for Securitize vaults
- `isEarnVaultOwnerVerified` for Earn vaults

and renders the existing red "!" badge when verification fails. The check is wired into the same badge chain as the unknown-collateral badge, so precedence over the deprecated/ramping/keyring/cyclical badges is unchanged. Vaults hydrated without governance data (`governorAdmin` not yet fetched) are not flagged, mirroring the guard in `useMarketGroups`, so lazily-loaded collateral doesn't flash the badge.

Out of scope (left as-is): the collapsed market card's "X unknown" counter still counts only external collateral, and the Matrix view has no unknown indicator.

## Test plan

- [x] `npm run typecheck`
- [x] `npx eslint components/entities/vault/discovery/DiscoveryMarketGraph.vue`
- [x] `npx vitest run tests/utils/discovery-calculations.test.ts` (22 passing)
- [ ] Visual check: expand a market whose member vaults show the "Unknown" risk-manager pill (e.g. Alterscope LRT) and confirm the red badge renders on those graph nodes
- [ ] Confirm verified markets render no new badges

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved risk-manager verification indicators in the vault discovery graph.
  * Added badges for unverified governors or owners, including external vaults and resolved group members.
  * Prevented incorrect warnings for vaults whose details load progressively.
  * Improved detection of verified governor administrators and reduced false verification results.

PR #806, opened 2026-08-13
chore: promote development to master

Seranged:
## Features
- Add on-chain Safe multisig badges with threshold and owner counts to vault governance roles, expose the EulerRouter oracle governor, and standardize vault overview address rows (#796) @kasperpawlowski
- Detect connected Safe wallets, force signature and Permit2 flows off, bundle supported approvals plus EVC operations into one atomic Safe proposal, and let users close review modals while long-running proposals finish with completion feedback (#797) @kasperpawlowski
- Exclude CoW Swap quotes for Safe wallets before approvals are sent, routing those flows to compatible on-chain swap providers instead (#798) @kasperpawlowski
- Bundle migration authorization grants, the migration EVC batch, and revocations into one atomic Safe proposal, eliminating the standing-authorization window on the bundled path (#799) @kasperpawlowski
- Badge discovery-graph member vaults whose risk manager or owner is unverified, while avoiding warnings during incomplete governance hydration (#795) @kasperpawlowski
- Expand Morpho migration discovery to Unichain, Monad, HyperEVM, and Arbitrum, gated by end-to-end migration support (#804) @Seranged
- Keep deployment manifests and runtime ABIs available through upstream outages by routing browser and server SDK reads through same-origin cached proxies, warming them, supporting emergency mirror overrides, and serving validated data stale for up to seven days (#814) @kasperpawlowski

## Fixes
- Decode Euler Earn Public Allocator flow-cap updates into readable per-strategy inflow and outflow limits with linked vault details and safe raw fallbacks (#801) @Seranged
- Keep batch simulation layers and failure attribution aligned when Terms of Use or other plugin operations are prepended, preventing fresh accounts from getting stuck on `Batch simulation not loaded` (#802) @kasperpawlowski
- Discover Aave supply-only positions even when they are not enabled as collateral, while keeping them separate from debt-backed migration positions (#803) @Seranged
- Surface post-execution restore and revoke transactions in the collapsed batch review summary, in execution order with duplicate transactions consolidated (#805) @kasperpawlowski
- Keep Earn vault verification documentation and rEUL address test fixtures aligned with the active label and SDK address surfaces (#807) @Seranged
- Pin Safe batch authorization and restoration labels to the latched review ceremony, separating in-proposal restorations from standalone transactions sent after execution (#808) @Seranged
- Preserve successful external migration results with ordered scans and retryable source warnings, resolve EulerRouter governance consistently across client and server vault hydration, and keep bundled Safe restoration summaries in one-to-one parity with proposal calls (#809) @Seranged
- Hide collateral-only, non-explorable product groups from Explore and its filters while preserving direct routes and external-collateral resolution (#816) @kasperpawlowski
- Include Securitize vault intrinsic APY on lend forms and keep the headline, projected estimate, and breakdown provider metadata sourced consistently (#819) @Seranged

## Documentation
- Align chain-scoped label caching, stale-response suppression, wrap-pair guards, SDK cache-policy examples, and migration-query freshness guidance with current behavior (#739) @cursor
- Document Earn uncovered losses, gasless-signature preferences, and transaction-based migration authorization behavior (#785) @cursor
- Document Safe wallet detection and atomic submissions, external migration discovery, batch plugin layers, and governance hydration rules (#811) @cursor
- Point transaction-building, Keyring, and Terms of Use guidance at the SDK plugin pipeline and the guard registry's current responsibilities (#815) @Seranged

## Dependencies
- Update `@eulerxyz/euler-v2-sdk` from 1.2.5 to 2.0.0 across the Safe and governance work (#796, #797) @kasperpawlowski


LeonardEulerXYZ (approved):
Reviewed the complete development → master promotion at `b1ec117bf14b2dbbcbc08e58d72142e6954edb50`.

No blocking findings.

Validation performed:
- verified the PR head exactly matches `development`
- inspected all 79 changed files / 94 promoted commits, with focused review of Safe transaction bundling and detachment, Permit2/signature policy, CoW exclusion, migration authorization lifecycle, Aave/Morpho discovery, governor/Safe probes, activity decoding, and UI summaries
- verified the 12 master-only commits compose cleanly; the only overlapping source path is byte-identical, and the synthetic merge preserves the two master-only docs/test changes
- verified `@eulerxyz/euler-v2-sdk@2.0.0` lock integrity against the npm artifact and tagged upstream release; no additional package graph changes or suspicious install hooks
- Node `v24.14.0`, npm `11.9.0`: clean lock install, typecheck, lint (0 errors), production build
- focused funds-path suite: 235/235 passed
- full suite: 1,733 passed, 1 skipped
- ran the built Nitro artifact and browser-smoked onboarding plus live Ethereum lend-market rendering/navigation with the documented local environment
- current GitHub lint, typecheck, test, Cursor review, and Railway preview checks are green

The local smoke emitted expected warnings where optional preview credentials/services were intentionally absent (Reown/Pyth/subgraph); no promotion-specific runtime failure was observed.

LeonardEulerXYZ (changes requested):
Superseding my earlier approval on `b1ec117bf14b2dbbcbc08e58d72142e6954edb50`: the promotion head does not include the two direct follow-ups now reviewed separately.

Blocking promotion item:
- #808 (`2bbd1322b5c56590430b665ff9fe1b8fac35ea4a`) fixes Safe batch-review ceremony labels so migration authorizations/restorations follow the execution mode latched when review opened rather than later live Safe detection. This signing-ceremony correction should land before production promotion.

Also expected before promotion:
- #807 (`53d7f86a6a46265305ad0c5ef2ac34a5b18e2c6a`) restores development/master parity for Earn verification docs and the SDK 2.0 rEUL fixture, preventing the next promotion from carrying stale source-branch content.

I reviewed and approved both follow-up heads independently. They are sibling one-commit branches from this PR head and touch disjoint files. Merge them into `development`, let #806 advance, then I will re-run the exact-head promotion composition/check guard and replace this review.

LeonardEulerXYZ (changes requested):
Additional exact-head findings from the independent challenge pass, verified against `b1ec117bf14b2dbbcbc08e58d72142e6954edb50`.

## Blocking

### Partial migration-source failure is presented as a complete result

`composables/useExternalMigrationPositions.ts:960-982` retains positions returned by one source when the other source rejects, but sets neither `error` nor any partial-result indicator whenever `nextPositions` is non-empty.

Failure path: Aave discovery rejects while Morpho returns at least one position (or vice versa) → the UI publishes the successful source's rows, sets `hasLoaded`, and presents the list indistinguishably from a complete discovery result. A user can therefore overlook a migratable position from the failed source.

Please retain the successful rows but expose a source-specific partial-discovery state/warning. Add a discriminating test where one whole source rejects while the sibling source returns a position; assert the retained row and visible incomplete-result state.

### EulerRouter governor verification fails open when router metadata is undecodable

`utils/vault/governor-verification.ts:90-97` applies the router-governor entity gate only when `getEulerRouterGovernor()` returns an address. `entities/oracle.ts:174-192` returns `null` for malformed/schema-incompatible EulerRouter metadata, so an object explicitly identified as `EulerRouter` can remain verified solely from `governorAdmin` when the router governor could not be verified.

That verified result feeds entity branding, market-group warnings, and the server verified-vault set. Undecodable governance metadata must not be equivalent to “no router-governor constraint.” Please fail closed for an identified EulerRouter whose governor cannot be decoded, or resolve `governor()` independently and model unresolved verification explicitly. The existing test at `tests/utils/vault/governor-verification.test.ts:154-169` currently codifies the fail-open result and should be inverted/replaced with a real malformed-data case.

## Material warnings

### Detached Safe tracking expires after five minutes

`utils/safeWalletTransactions.ts:3-4,237-248` terminates confirmation polling after five minutes. `composables/useSafeExecutionDetachment.ts:140-157` then reports the rejection and clears the tracked slot. A successfully proposed multisig waiting on ordinary co-signers can therefore become untracked even though the UI describes Safe waits as lasting “minutes to days.”

This need not block if bounded polling is intentional, because the error clearly says status is unknown and instructs the user to verify in Safe before retrying. It should nevertheless retain a resumable Safe identifier/manual status path rather than discard tracking. A fake-timer test should cover pending beyond five minutes followed by eventual success.

### Bundled Safe summary can hide duplicate restoration calls that remain in the proposal

`components/BatchReviewModal.vue:198-219` deduplicates restoration rows by `txKey`, relying on the sequential prerequisite invariant. The bundled path at `composables/useTxBatch.ts:2489-2508` independently collects and concatenates every entry's revokes without deduplication. Two bundled entries can therefore place two identical restoration calls in the Safe proposal while the collapsed review summary shows one.

Please either deduplicate the actual bundled calls under a proved-safe invariant or display every call the proposal contains. Add a two-entry test with identical grant/revoke transactions and assert display/execution parity.

These findings do not alter my approvals on #807 or #808: the affected behavior is inherited from their #806 base rather than introduced by either follow-up.

LeonardEulerXYZ (changes requested):
Reviewed the final `development` → `master` composition at exact head `3c73cf214605e0358a997b2c54e271456648307d`.

## Blocking — **Copy calldata** can encode a different core Safe plan from the proposal that executes

The modal correctly latches fresh Safe grant/revoke rows and execution plans when review opens, but it then prepares the older simulated `lastMerged` plan for the calldata export:

- `components/BatchReviewModal.vue:375-387` calls `prepareBundledExecution()` and then separately calls `prepareBatchPlan()`.
- `composables/useTxBatch.ts:2410-2419` shows `prepareBatchPlan()` always prepares `lastMerged`.
- `components/BatchReviewModal.vue:414-459` wraps that prepared preview plan with the *latched* grants/revokes and labels the result as matching the actual submission.
- By contrast, `composables/useTxBatch.ts:2575-2595` executes a fresh merge of `latchedBundledExecution.plans`.

Those core plans are not invariantly identical: migration entries intentionally rebuild their bundled simulation/execution plan from the authorization state resolved when review opens. If that state drifted since add-time simulation—the case the #808/#809 ceremony latch is designed to handle—the exported JSON can contain:

`latched grants + stale preview core plan + latched revokes`

while confirmation submits:

`latched grants + fresh latched core plan + latched revokes`.

I reproduced this at this exact head with a discriminating `useTxBatch` probe: distinct preview and latched execution plans were supplied; after `prepareBundledExecution()`, `prepareBatchPlan()` still called `prepareTransactionPlan(previewPlan)`, while execution retained `latchedExecutionPlan`. The focused file remained green at **58/58 tests**, demonstrating the uncovered path rather than a broad harness failure.

This does not make execution fail open, but it breaks the funds-sensitive one-to-one invariant between the review/export artifact and the proposal a Safe owner is asked to approve.

**Required:** when a bundled ceremony is latched, prepare and encode the merged `latchedBundledExecution.plans` for **Copy calldata**, and ensure execution consumes that same prepared core envelope (or otherwise prove byte identity). Add a regression where preview and latched plans differ and assert exported `before/core/after` calls equal the calls submitted to the Safe.

## Non-blocking warning retained from the prior promotion review

Safe status polling still expires after five minutes (`utils/safeWalletTransactions.ts:3-4,206-241`). A detached pending proposal then resolves as status-unknown and releases its tracking slot (`composables/useSafeExecutionDetachment.ts:137-158`) even though later Safe confirmation remains possible. This can produce a false terminal error and no later completion toast for ordinary multi-signer delays. It is not the blocker above, but it remains a release caveat worth fixing with durable/restartable Safe transaction tracking.

## What did compose correctly

- #807 head `53d7f86a6a46265305ad0c5ef2ac34a5b18e2c6a`, #809 fixed head `48aa55ccdf3104daa79556d5e9c51c78279bee96`, and #808’s merged resolution are all ancestors of this promotion head.
- The final #808/#809 Safe conflict resolution preserves latched ceremony grouping and duplicate bundled restoration rows one-to-one with proposal calls.
- Migration partial-source observability and latest-generation ownership survived integration.
- Real-SDK EulerRouter governance hydration and fail-closed verification survived integration.
- Exact-head local validation passed: focused suites, full test suite, ESLint, typecheck, production build, and whitespace checks.
- Headed Chromium smoke of the exact built artifact passed across home, Explore, Lend, Borrow, Earn, and migration portfolio routes; bounded chain/token/vault API checks returned valid JSON.
- GitHub CI, security review, and the exact-head Railway deployment are successful.

Please push the calldata-parity fix and request a fresh review; this review is bound only to `3c73cf214605e0358a997b2c54e271456648307d`.

LeonardEulerXYZ (approved):
Reviewed the final `development` → `master` promotion at exact head `9c487df5c265b90df5786381733b8b509f31a7d5`.

## Verdict

Approved for production promotion with the explicitly accepted P2 transaction-lifecycle residuals below. I found no P0/P1 blocker on the current head.

## Promotion shape and validation

- PR head exactly matches `origin/development`; `origin/master` is now an ancestor, with no master-only composition branch remaining.
- The master integration changes only `package.json` / `package-lock.json` relative to the prior reviewed head and preserves `@eulerxyz/euler-v2-sdk@2.0.0` while carrying the master audit overrides.
- Clean install with the repo-pinned Node `v24.14.0` and npm `11.10.0` succeeded; npm audit reported **0 vulnerabilities**.
- Focused current-head lifecycle/server/recent-change suite: **202/202 passed** across Safe batch execution/detachment/status polling, external migration partial results, EulerRouter governor verification, market-group visibility, intrinsic APY, ABI proxy, and EulerChains proxy.
- Exact-head GitHub CI lint, typecheck, build/full test job, Cursor security review, and Railway preview are green.
- Headed Chromium under managed Xvfb loaded `/`, `/explore`, `/lend`, `/borrow`, `/earn`, `/portfolio`, and `/portfolio/migrate` from the exact-head dev preview with HTTP 200, no Cloudflare challenge, and no console errors.
- Deployed internal API probes returned valid JSON for EulerChains, AccountLens/UtilsLens/VaultLens ABIs, V3 token/EVault/Earn paths, and oracle adapters/routers.
- Whole-diff review covered the Safe/Permit2/CoW/migration execution paths, transaction-plan encoding and review summaries, cached manifest/ABI routes and override precedence, SDK 2.0 lock integrity, external migration discovery, governor verification, market visibility, intrinsic APY, env/CSP changes, and dependency/supply-chain surfaces.

## Accepted P2 residual risk

- Safe confirmation tracking times out after five minutes and reports status as unknown. The local tracking gate is then released, so a user who retries without checking Safe could create another proposal. The warning is explicit and this is treated as a recovery/retry edge case, not a P0/P1 promotion blocker.
- In the bundled migration edge case where authorization state changes between add-time preview and review-open latching, **Copy calldata** can still encode the older preview core plan while Confirm executes the freshly latched core plan. Execution remains fail-closed and consumes the latched ceremony; the mismatch is confined to the optional export/review artifact and is accepted here as P2.
- Browser-storage cross-tab exactly-once/replay coordination remains deliberately outside this parity-scoped release.

No additional blocking finding was identified in the current-head challenge pass.

dglowinski (approved):

LeonardEulerXYZ:
Follow-up after the delayed independent lifecycle/security passes completed, rechecked against current exact head `9c487df5c265b90df5786381733b8b509f31a7d5`:

- **No new P0/P1 blocker. Approval stands.**
- The deployment-manifest cache-admission concern raised by one pass does **not** reproduce on either the prior immutable head or the current head. `server/api/internal/euler-chains.get.ts` already validates positive chain IDs, required `coreAddrs` / `lensAddrs`, and EVM address shape before cache admission; an invalid enabled-chain entry throws so stale last-known-good data remains available. The focused EulerChains route suite passed earlier in the final-head run.
- One additional transaction-lifecycle residual is worth recording: the private pending-add queue is not generation-owned by the review ceremony and is not included in `canExecuteBatch`. With an existing cart, a user could begin a slow second add and open review during that narrow in-flight window; an older bundled preparation can republish after cart invalidation. The stale proposal would omit the late operation (rather than introduce unseen calldata) and successful execution would then clear the current cart. Production add handlers await `addBatchEntry()` before redirect/reset, so reaching this requires an existing cart plus overlapping user actions and slow asynchronous planning. Under the stated promotion policy, I classify this as an accepted **P2 timing edge case**, alongside the already-recorded calldata-export and detached-Safe retry residuals.

Browser-storage cross-tab replay coordination remains explicitly outside this parity-scoped release.

PR #810, opened 2026-08-13
fix: preserve batch review and execution parity

Seranged:
## Summary
- Use one immutable, generation-scoped ceremony for batch review, calldata export, gas estimation, and wallet submission.
- Reject stale work across cart, account, chain, connector-session, and Safe-classification changes, including at asynchronous wallet boundaries.
- Request migration authorization signatures only after confirmation and allow them to change only the exact placeholder representation reviewed by the user.

## Background and rationale

Batch execution has evolved in stages. The original flow captured operation plans, merged them for review, and then executed them. As support expanded to migrations, Safe proposals, dynamic authorization, connector changes, plugins, and confirmation-time signatures, more asynchronous preparation points were introduced. Earlier fixes pinned individual pieces—such as reviewed core plans and Safe/migration authorization payloads—but the complete transaction ceremony could still combine state captured at different times.

This PR completes that design: one immutable ceremony, bound to the current cart and wallet session, is the source of truth for review, copying, gas estimation, supported simulation, and submission. Material changes invalidate the ceremony and require a fresh review; confirmation-time signing can only complete the exact authorization placeholders represented during review. This preserves the core transaction-stack invariant that the plan approved by the user is the plan ultimately sent to the wallet.

Beyond closing the current race conditions, this creates a safer extension point for future connectors, authorization methods, plugins, migrations, and wallet execution modes. New functionality can contribute to the same ceremony instead of introducing another parallel preparation path. That reduces the combinations of mutable state and timing we need to reason about, makes parity testable at the final wallet boundary, and provides a fail-closed foundation as the batch stack grows.

## Changes
- Deduplicate preparation per cart generation and bind each ceremony to the normalized owner, chain, connector ID and UID, and resolved Safe classification.
- Derive authorization calls and rows from one resolution, preserving exact grant → prepared core → reverse restoration ordering and every duplicate wallet prompt.
- Compare confirm-time migrations with the reviewed prepared envelope, supporting both raw signatures and the ABI-encoded `v`/`r`/`s` representation used by Aave, Morpho, and MetaMorpho while rejecting all other plan drift.
- Estimate the frozen prepared envelope directly so plugins and approval resolution cannot produce a different call vector before broadcast.
- Scope asynchronous adds, resimulation data, snapshots, execution completion, errors, cart clearing, and redirects to their originating ceremony and wallet context.
- Derive expanded rows, unverified-vault disclosure, copied calldata, and Safe submission from the ceremony plan and shared encoder.
- Keep whole-batch third-party simulation unavailable because a single-transaction simulation cannot represent the complete reviewed ceremony. Per-operation simulation remains available.

## Test plan
- [x] Run targeted ESLint across all changed files.
- [x] Run `npm run typecheck`.
- [x] Run the production build.
- [x] Run `npm run test:run -- --silent --reporter=dot`: 1,797 tests passed across 183 files.
- [x] Run the focused ABI-parity and prepared-estimation cohort: 122 tests passed across 3 files.
- [x] Cover overlapping preparation, pending adds, wallet-context changes, stale snapshots, final wallet boundaries, late completion, deferred migration signatures, raw and ABI-encoded signature parity, prepared-envelope gas estimation, duplicate authorization prompts, unverified-vault disclosure, calldata/submission parity, and exact execution ordering.

### DIFF
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -1005,6 +1005,7 @@ export const useEulerTx = () => {
 
   const signMigrationAuthorization = async (
     request: MigrationAuthorizationRequest,
+    options?: { beforeSignature?: () => void },
   ): Promise<SignedMigrationAuthorization> => {
     if (isSpyMode.value) {
       throw new Error('Authorization signatures are disabled in spy mode')
@@ -1019,9 +1020,10 @@ export const useEulerTx = () => {
       currentAccount: currentAccount.address,
       currentChainId: currentAccount.chainId,
     })
+    options?.beforeSignature?.()
     const signature = await signTypedDataAsync(request.typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
     const postMigrationAuthorization = request.postMigrationAuthorization
-      ? await signMigrationAuthorization(request.postMigrationAuthorization)
+      ? await signMigrationAuthorization(request.postMigrationAuthorization, options)
       : undefined
     return {
       request,
@@ -1208,6 +1210,7 @@ export const useEulerTx = () => {
     expectedChainId,
     connector,
     resolveHash,
+    beforeBroadcast,
   }: {
     isOkx: boolean
     expectedAccount: Address
@@ -1220,6 +1223,8 @@ export const useEulerTx = () => {
      */
     connector?: ReturnType<typeof getAccount>['connector']
     resolveHash?: (hash: Hash) => Promise<Hash>
+    /** Final caller-owned freshness check at the wallet submission boundary. */
+    beforeBroadcast?: () => void
   }) => {
     let okxDelayPending = false
     const send = async ({ to, data, value }: { to: Address, data: Hex, value?: bigint }) => {
@@ -1234,6 +1239,7 @@ export const useEulerTx = () => {
         currentAccount: currentAccount.address,
         currentChainId: currentAccount.chainId,
       })
+      beforeBroadcast?.()
       const hash = await sendTransactionAsync({
         account: expectedAccount,
         chainId: expectedChainId,
@@ -1264,6 +1270,7 @@ export const useEulerTx = () => {
     options?: {
       onBroadcast?: (index: number, walletContext: WalletExecutionContext) => void
       walletContext?: WalletExecutionContext
+      beforeBroadcast?: () => void
     },
   ): Promise<TransactionReceipt[]> => {
     if (isSpyMode.value) {
@@ -1293,6 +1300,7 @@ export const useEulerTx = () => {
       expectedAccount: owner,
       expectedChainId: cid,
       connector,
+      beforeBroadcast: options?.beforeBroadcast,
     })
 
     const receipts: TransactionReceipt[] = []
@@ -1421,7 +1429,7 @@ export const useEulerTx = () => {
    * bundling brings no benefit (fewer than two calls) — callers fall back to
    * sequential execution.
    */
-  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall }: {
+  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall, beforeBroadcast }: {
     plan: TransactionPlan
     chainId: number
     owner: Address
@@ -1440,6 +1448,8 @@ export const useEulerTx = () => {
      * "no Safe context" — with it set, undefined strictly means the latter.
      */
     allowSingleCall?: boolean
+    /** Final caller-owned freshness check at the Safe submission boundary. */
+    beforeBroadcast?: () => void
   }) => {
     let planCalls
     try {
@@ -1474,6 +1484,7 @@ export const useEulerTx = () => {
       currentAccount: currentAccount.address,
       currentChainId: currentAccount.chainId,
     })
+    beforeBroadcast?.()
 
     // Pin submission to the connector whose provider was identified as Safe.
     // Without it, wagmi resolves the currently-active connector, and a
@@ -1587,7 +1598,10 @@ export const useEulerTx = () => {
     return result
   }
 
-  const executePreparedPlan = async (prepared: TransactionPlanPrepared) => {
+  const executePreparedPlan = async (
+    prepared: TransactionPlanPrepared,
+    options?: { beforeBroadcast?: () => void },
+  ) => {
     if (isSpyMode.value) {
       throw new Error('Transactions are disabled in spy mode')
     }
@@ -1641,6 +1655,7 @@ export const useEulerTx = () => {
         connector,
         safeWalletProvider,
         sdk,
+        beforeBroadcast: options?.beforeBroadcast,
       })
       if (bundled) {
         finalizeExecution(bundled)
@@ -1653,6 +1668,7 @@ export const useEulerTx = () => {
       expectedAccount: preparedOwner,
       expectedChainId: prepared.chainId,
       connector,
+      beforeBroadcast: options?.beforeBroadcast,
       resolveHash: safeWalletProvider
         ? async submittedHash => (await waitForSafeTransactionExecution({
           submittedHash,
@@ -1666,6 +1682,7 @@ export const useEulerTx = () => {
       prepared: effectivePrepared,
       sendTransaction,
       signTypedData: async (typedData) => {
+        options?.beforeBroadcast?.()
         const signature = await signTypedDataAsync(typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
         return signature as Hex
       },
@@ -1694,7 +1711,7 @@ export const useEulerTx = () => {
       before?: readonly PlainTxRequest[]
       after?: readonly PlainTxRequest[]
     },
-    options?: { allowSingleCall?: boolean },
+    options?: { allowSingleCall?: boolean, beforeBroadcast?: () => void },
   ) => {
     if (isSpyMode.value) {
       throw new Error('Transactions are disabled in spy mode')
@@ -1738,6 +1755,7 @@ export const useEulerTx = () => {
       sdk,
       extraCalls,
       allowSingleCall: options?.allowSingleCall,
+      beforeBroadcast: options?.beforeBroadcast,
     })
     if (!bundled) return undefined
     finalizeExecution(bundled)
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -70,10 +70,10 @@ import {
   assertWalletExecutionContext,
   type WalletExecutionContext,
 } from '~/utils/walletExecutionContext'
+import { PLACEHOLDER_MIGRATION_AUTHORIZATION_SIGNATURE } from '~/utils/migrationAuthorizationSignatures'
 
 const OKX_POST_APPROVE_DELAY_MS = 3000
 const ERC20_APPROVE_SELECTOR = '0x095ea7b3'
-const PLACEHOLDER_AUTHORIZATION_SIGNATURE = `0x${'00'.repeat(65)}` as Hex
 const SUB_ACCOUNT_SNAPSHOT_FETCH_OPTIONS = {
   populateVaults: false,
   populateMarketPrices: false,
@@ -1041,7 +1041,7 @@ export const useEulerTx = () => {
 
     return {
       request,
-      signature: PLACEHOLDER_AUTHORIZATION_SIGNATURE,
+      signature: PLACEHOLDER_MIGRATION_AUTHORIZATION_SIGNATURE,
       ...(postMigrationAuthorization ? { postMigrationAuthorization } : {}),
     }
   }
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -99,6 +99,16 @@ const isOkxWallet = async (connector?: { id?: string, name?: string, getProvider
   return false
 }
 
+/**
+ * A plan submission the wallet has accepted but whose confirmation is still
+ * pending: an on-chain transaction hash for standard wallets, or a safeTxHash
+ * proposal id for Safes (resolvable only through the Safe wallet provider).
+ */
+export interface PreparedPlanBroadcast {
+  kind: 'transaction' | 'proposal'
+  hash: Hash
+}
+
 export interface PlanDepositInput {
   vaultAddress: Address
   assetAddress: Address
@@ -1217,6 +1227,7 @@ export const useEulerTx = () => {
     connector,
     resolveHash,
     beforeBroadcast,
+    onBroadcast,
   }: {
     isOkx: boolean
     expectedAccount: Address
@@ -1231,6 +1242,13 @@ export const useEulerTx = () => {
     resolveHash?: (hash: Hash) => Promise<Hash>
     /** Final caller-owned freshness check at the wallet submission boundary. */
     beforeBroadcast?: () => void
+    /**
+     * Fires as soon as the wallet accepted the submission, before any
+     * confirmation wait — from here on the transaction may land even if
+     * everything after this point fails, so callers use it to stop treating
+     * a later error as "never sent".
+     */
+    onBroadcast?: (hash: Hash) => void
   }) => {
     let okxDelayPending = false
     const send = async ({ to, data, value }: { to: Address, data: Hex, value?: bigint }) => {
@@ -1258,6 +1276,7 @@ export const useEulerTx = () => {
         okxDelayPending = true
       }
       const submittedHash = hash as Hash
+      onBroadcast?.(submittedHash)
       return resolveHash ? resolveHash(submittedHash) : submittedHash
     }
     return send
@@ -1354,12 +1373,14 @@ export const useEulerTx = () => {
   const executeMigrationAuthorizationGrants = async (
     request: MigrationAuthorizationRequest,
     broadcastRevokes: MigrationAuthorizationRevoke[] = [],
+    options?: { beforeBroadcast?: () => void },
   ): Promise<MigrationAuthorizationRevoke[]> => {
     const { grants, revokesByGrant } = encodeMigrationAuthorizationTxs(request)
     await sendPlainTransactions(grants, {
       // The request was prepared for this exact owner/network. Do not let a
       // wallet switch during the preceding SDK reads retarget the grant.
       walletContext: { account: request.owner, chainId: request.chainId },
+      beforeBroadcast: options?.beforeBroadcast,
       onBroadcast: (index, walletContext) => {
         const revoke = revokesByGrant[index]
         if (revoke) {
@@ -1435,7 +1456,7 @@ export const useEulerTx = () => {
    * bundling brings no benefit (fewer than two calls) — callers fall back to
    * sequential execution.
    */
-  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall, beforeBroadcast }: {
+  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall, beforeBroadcast, onBroadcast }: {
     plan: TransactionPlan
     chainId: number
     owner: Address
@@ -1456,6 +1477,13 @@ export const useEulerTx = () => {
     allowSingleCall?: boolean
     /** Final caller-owned freshness check at the Safe submission boundary. */
     beforeBroadcast?: () => void
+    /**
+     * Fires once the Safe accepted the proposal (safeTxHash exists), before
+     * the execution wait — from here on the proposal may execute even if
+     * status polling fails, so callers use it to stop treating a later error
+     * as "never submitted".
+     */
+    onBroadcast?: (safeTxHash: Hash) => void
   }) => {
     let planCalls
     try {
@@ -1508,6 +1536,7 @@ export const useEulerTx = () => {
     if (!/^0x[0-9a-f]{64}$/i.test(id)) {
       throw new Error('Safe wallet returned an unexpected call bundle id')
     }
+    onBroadcast?.(id as Hash)
 
     const execution = await waitForSafeTransactionExecution({
       submittedHash: id as Hash,
@@ -1606,7 +1635,17 @@ export const useEulerTx = () => {
 
   const executePreparedPlan = async (
     prepared: TransactionPlanPrepared,
-    options?: { beforeBroadcast?: () => void },
+    options?: {
+      beforeBroadcast?: () => void
+      /**
+       * Fires as soon as the wallet accepted a plan submission (a transaction
+       * hash for standard wallets, a safeTxHash proposal id for Safes),
+       * before any confirmation wait. From that point the submission may land
+       * even when a later step throws, so callers use it to distinguish
+       * "never sent" from "sent but unconfirmed".
+       */
+      onBroadcast?: (broadcast: PreparedPlanBroadcast) => void
+    },
   ) => {
     if (isSpyMode.value) {
       throw new Error('Transactions are disabled in spy mode')
@@ -1662,6 +1701,7 @@ export const useEulerTx = () => {
         safeWalletProvider,
         sdk,
         beforeBroadcast: options?.beforeBroadcast,
+        onBroadcast: hash => options?.onBroadcast?.({ kind: 'proposal', hash }),
       })
       if (bundled) {
         finalizeExecution(bundled)
@@ -1675,6 +1715,12 @@ export const useEulerTx = () => {
       expectedChainId: prepared.chainId,
       connector,
       beforeBroadcast: options?.beforeBroadcast,
+      // A Safe's sequential fallback submits through the Safe app, so the
+      // wallet returns a safeTxHash proposal id, not an on-chain hash.
+      onBroadcast: hash => options?.onBroadcast?.({
+        kind: safeWalletProvider ? 'proposal' : 'transaction',
+        hash,
+      }),
       resolveHash: safeWalletProvider
         ? async submittedHash => (await waitForSafeTransactionExecution({
           submittedHash,
@@ -1717,7 +1763,12 @@ export const useEulerTx = () => {
       before?: readonly PlainTxRequest[]
       after?: readonly PlainTxRequest[]
     },
-    options?: { allowSingleCall?: boolean, beforeBroadcast?: () => void },
+    options?: {
+      allowSingleCall?: boolean
+      beforeBroadcast?: () => void
+      /** See executePreparedPlan — fires with the safeTxHash proposal id. */
+      onBroadcast?: (broadcast: PreparedPlanBroadcast) => void
+    },
   ) => {
     if (isSpyMode.value) {
       throw new Error('Transactions are disabled in spy mode')
@@ -1762,6 +1813,7 @@ export const useEulerTx = () => {
       extraCalls,
       allowSingleCall: options?.allowSingleCall,
       beforeBroadcast: options?.beforeBroadcast,
+      onBroadcast: hash => options?.onBroadcast?.({ kind: 'proposal', hash }),
     })
     if (!bundled) return undefined
     finalizeExecution(bundled)
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -100,13 +100,29 @@ const isOkxWallet = async (connector?: { id?: string, name?: string, getProvider
 }
 
 /**
- * A plan submission the wallet has accepted but whose confirmation is still
- * pending: an on-chain transaction hash for standard wallets, or a safeTxHash
- * proposal id for Safes (resolvable only through the Safe wallet provider).
+ * A plan submission the wallet has accepted: an on-chain transaction hash for
+ * standard wallets, or a safeTxHash proposal id for Safes (resolvable only
+ * through the Safe wallet provider).
+ *
+ * The executor waits for each submission's receipt before sending the next,
+ * so it re-fires the same broadcast with status 'confirmed' once the receipt
+ * succeeded — at any failure, at most the single latest 'submitted' broadcast
+ * without a matching 'confirmed' has an unknown outcome.
  */
 export interface PreparedPlanBroadcast {
   kind: 'transaction' | 'proposal'
   hash: Hash
+  /**
+   * Which plan step produced the submission. 'bundle' (one Safe proposal for
+   * the whole plan) and 'evcBatch' move value; 'approval' and 'pluginCall'
+   * (TOS/oracle prerequisites) are idempotent and safe to re-run.
+   */
+  item: 'bundle' | 'evcBatch' | 'approval' | 'pluginCall'
+  /** Zero-based position in this attempt's send order. */
+  index: number
+  /** True when no value-moving submission follows this one in the plan. */
+  completesPlan: boolean
+  status: 'submitted' | 'confirmed'
 }
 
 export interface PlanDepositInput {
@@ -1031,7 +1047,14 @@ export const useEulerTx = () => {
       currentChainId: currentAccount.chainId,
     })
     options?.beforeSignature?.()
-    const signature = await signTypedDataAsync(request.typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
+    // Pin the signature request to the connector that just passed the wallet
+    // context check. Without it wagmi resolves the currently-active connector
+    // at request time, so a connector switched during the check-to-sign gap
+    // would receive the request despite never being validated.
+    const signature = await signTypedDataAsync({
+      ...(request.typedData as unknown as Parameters<typeof signTypedDataAsync>[0]),
+      ...(currentAccount.connector ? { connector: currentAccount.connector } : {}),
+    })
     const postMigrationAuthorization = request.postMigrationAuthorization
       ? await signMigrationAuthorization(request.postMigrationAuthorization, options)
       : undefined
@@ -1623,7 +1646,12 @@ export const useEulerTx = () => {
       usePermit2: isKnownSafe ? false : signaturesEnabled.value,
       sendTransaction,
       signTypedData: async (typedData) => {
-        const signature = await signTypedDataAsync(typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
+        // Same pinning as sendTransaction: sign through the captured
+        // connector, never the currently-active one.
+        const signature = await signTypedDataAsync({
+          ...(typedData as unknown as Parameters<typeof signTypedDataAsync>[0]),
+          ...(connector ? { connector } : {}),
+        })
         return signature as Hex
       },
       onProgress: (_progress: TransactionPlanExecutionProgress) => {},
@@ -1692,6 +1720,7 @@ export const useEulerTx = () => {
 
     if (safeWalletProvider && connector) {
       // Prepared plans already ran plugins and approval resolution.
+      let bundleBroadcast: PreparedPlanBroadcast | undefined
       const bundled = await executePlanAsSafeBundle({
         plan: effectivePrepared.plan,
         chainId: prepared.chainId,
@@ -1701,13 +1730,40 @@ export const useEulerTx = () => {
         safeWalletProvider,
         sdk,
         beforeBroadcast: options?.beforeBroadcast,
-        onBroadcast: hash => options?.onBroadcast?.({ kind: 'proposal', hash }),
+        onBroadcast: (hash) => {
+          bundleBroadcast = {
+            kind: 'proposal',
+            hash,
+            item: 'bundle',
+            index: 0,
+            completesPlan: true,
+            status: 'submitted',
+          }
+          options?.onBroadcast?.(bundleBroadcast)
+        },
       })
       if (bundled) {
+        // A truthy bundle result means the proposal executed with a
+        // successful receipt — the submission's outcome is no longer unknown.
+        if (bundleBroadcast) {
+          options?.onBroadcast?.({ ...bundleBroadcast, status: 'confirmed' })
+        }
         finalizeExecution(bundled)
         return bundled
       }
     }
+
+    // Ordered execution tracking for the sequential path. The executor emits
+    // a pre-send progress event (no hash) naming the item about to be sent,
+    // then waits for the receipt and re-emits with the confirmed hash — so
+    // the item announced by the latest pre-send event classifies the next
+    // wallet submission, and a hash-bearing event confirms it.
+    const planItems = effectivePrepared.plan
+    const lastValueMovingItem = [...planItems].reverse().find(item =>
+      item.type !== 'requiredApproval' && item.type !== 'cowSwap')
+    let pendingSend: { item: PreparedPlanBroadcast['item'], completesPlan: boolean } | undefined
+    let lastSubmitted: PreparedPlanBroadcast | undefined
+    let sendIndex = 0
 
     const sendTransaction = buildSendTransaction({
       isOkx,
@@ -1717,10 +1773,19 @@ export const useEulerTx = () => {
       beforeBroadcast: options?.beforeBroadcast,
       // A Safe's sequential fallback submits through the Safe app, so the
       // wallet returns a safeTxHash proposal id, not an on-chain hash.
-      onBroadcast: hash => options?.onBroadcast?.({
-        kind: safeWalletProvider ? 'proposal' : 'transaction',
-        hash,
-      }),
+      onBroadcast: (hash) => {
+        lastSubmitted = {
+          kind: safeWalletProvider ? 'proposal' : 'transaction',
+          hash,
+          // Fail toward the strict classification: an unattributed send is
+          // treated as the value-moving batch, never as a re-runnable step.
+          item: pendingSend?.item ?? 'evcBatch',
+          index: sendIndex++,
+          completesPlan: pendingSend?.completesPlan ?? true,
+          status: 'submitted',
+        }
+        options?.onBroadcast?.(lastSubmitted)
+      },
       resolveHash: safeWalletProvider
         ? async submittedHash => (await waitForSafeTransactionExecution({
           submittedHash,
@@ -1735,10 +1800,33 @@ export const useEulerTx = () => {
       sendTransaction,
       signTypedData: async (typedData) => {
         options?.beforeBroadcast?.()
-        const signature = await signTypedDataAsync(typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
+        // Same pinning as sendTransaction: sign through the captured
+        // connector, never the currently-active one.
+        const signature = await signTypedDataAsync({
+          ...(typedData as unknown as Parameters<typeof signTypedDataAsync>[0]),
+          ...(connector ? { connector } : {}),
+        })
         return signature as Hex
       },
-      onProgress: (_progress: TransactionPlanExecutionProgress) => {},
+      onProgress: (progress: TransactionPlanExecutionProgress) => {
+        if (progress.hash) {
+          // Receipt confirmed for the outstanding submission.
+          if (lastSubmitted) {
+            options?.onBroadcast?.({ ...lastSubmitted, status: 'confirmed' })
+            lastSubmitted = undefined
+          }
+          return
+        }
+        if (progress.status === 'approval') {
+          pendingSend = { item: 'approval', completesPlan: false }
+        }
+        else if (progress.status === 'evcBatch' || progress.status === 'contractCall') {
+          pendingSend = {
+            item: progress.status === 'evcBatch' ? 'evcBatch' : 'pluginCall',
+            completesPlan: !!progress.item && progress.item === lastValueMovingItem,
+          }
+        }
+      },
     })
 
     finalizeExecution(result)
@@ -1802,6 +1890,7 @@ export const useEulerTx = () => {
       effectivePrepared = { ...prepared, usePermit2: false }
     }
 
+    let bundleBroadcast: PreparedPlanBroadcast | undefined
     const bundled = await executePlanAsSafeBundle({
       plan: effectivePrepared.plan,
       chainId: prepared.chainId,
@@ -1813,9 +1902,24 @@ export const useEulerTx = () => {
       extraCalls,
       allowSingleCall: options?.allowSingleCall,
       beforeBroadcast: options?.beforeBroadcast,
-      onBroadcast: hash => options?.onBroadcast?.({ kind: 'proposal', hash }),
+      onBroadcast: (hash) => {
+        bundleBroadcast = {
+          kind: 'proposal',
+          hash,
+          item: 'bundle',
+          index: 0,
+          completesPlan: true,
+          status: 'submitted',
+        }
+        options?.onBroadcast?.(bundleBroadcast)
+      },
     })
     if (!bundled) return undefined
+    // A truthy bundle result means the proposal executed with a successful
+    // receipt — the submission's outcome is no longer unknown.
+    if (bundleBroadcast) {
+      options?.onBroadcast?.({ ...bundleBroadcast, status: 'confirmed' })
+    }
     finalizeExecution(bundled)
     return bundled
   }
--- a/composables/useEulerTx.ts
+++ b/composables/useEulerTx.ts
@@ -49,6 +49,8 @@ import {
   type PlainTxRequest,
 } from '~/utils/migrationAuthorizationTxs'
 import { logWarn } from '~/utils/errorHandling'
+import { assertNoConflictingPendingSubmission } from '~/utils/pendingSubmissionGate'
+import { walletNeverAcceptedSubmission } from '~/utils/pendingSubmissions'
 import { invalidateSdkQueries } from '~/utils/sdk-query-cache'
 import { INVALIDATE_AFTER_TX } from '~/utils/sdk-query-policy'
 import { waitForSubgraphBlock } from '~/utils/subgraph'
@@ -100,18 +102,24 @@ const isOkxWallet = async (connector?: { id?: string, name?: string, getProvider
 }
 
 /**
- * A plan submission the wallet has accepted: an on-chain transaction hash for
- * standard wallets, or a safeTxHash proposal id for Safes (resolvable only
- * through the Safe wallet provider).
+ * One plan step's lifecycle at the wallet boundary.
+ *
+ * 'armed' fires immediately before the wallet is invoked for the step — the
+ * wallet call is itself an external side-effect: it may accept and submit the
+ * request, then disconnect or fail to return an id. From 'armed' on, the step
+ * must be treated as possibly-sent until a terminal signal. 'submitted' fires
+ * once the wallet returned an id (an on-chain transaction hash for standard
+ * wallets, or a safeTxHash proposal id for Safes, resolvable only through the
+ * Safe wallet provider). 'rejected' fires only when the wallet provably never
+ * accepted the armed request (explicit user rejection or a pre-dispatch
+ * connector failure). 'confirmed' re-fires the submitted broadcast once its
+ * receipt succeeded.
  *
  * The executor waits for each submission's receipt before sending the next,
- * so it re-fires the same broadcast with status 'confirmed' once the receipt
- * succeeded — at any failure, at most the single latest 'submitted' broadcast
- * without a matching 'confirmed' has an unknown outcome.
+ * so at any failure at most the single latest armed/submitted broadcast
+ * without a terminal signal has an unknown outcome.
  */
-export interface PreparedPlanBroadcast {
-  kind: 'transaction' | 'proposal'
-  hash: Hash
+export type PreparedPlanBroadcast = {
   /**
    * Which plan step produced the submission. 'bundle' (one Safe proposal for
    * the whole plan) and 'evcBatch' move value; 'approval' and 'pluginCall'
@@ -122,8 +130,10 @@ export interface PreparedPlanBroadcast {
   index: number
   /** True when no value-moving submission follows this one in the plan. */
   completesPlan: boolean
-  status: 'submitted' | 'confirmed'
-}
+} & (
+  | { status: 'armed' | 'rejected', kind?: undefined, hash?: undefined }
+  | { status: 'submitted' | 'confirmed', kind: 'transaction' | 'proposal', hash: Hash }
+)
 
 export interface PlanDepositInput {
   vaultAddress: Address
@@ -1250,6 +1260,8 @@ export const useEulerTx = () => {
     connector,
     resolveHash,
     beforeBroadcast,
+    onArm,
+    onSubmitError,
     onBroadcast,
   }: {
     isOkx: boolean
@@ -1265,6 +1277,19 @@ export const useEulerTx = () => {
     resolveHash?: (hash: Hash) => Promise<Hash>
     /** Final caller-owned freshness check at the wallet submission boundary. */
     beforeBroadcast?: () => void
+    /**
+     * Fires immediately before the wallet is invoked. The wallet call is
+     * itself an external side-effect — it may accept and dispatch the
+     * request, then fail before returning its id — so callers persist replay
+     * protection here, with no hash yet.
+     */
+    onArm?: () => void
+    /**
+     * Fires when the wallet invocation threw, before the error propagates —
+     * so callers can release the armed state when the error proves the
+     * wallet never accepted the request.
+     */
+    onSubmitError?: (error: unknown) => void
     /**
      * Fires as soon as the wallet accepted the submission, before any
      * confirmation wait — from here on the transaction may land even if
@@ -1287,14 +1312,22 @@ export const useEulerTx = () => {
         currentChainId: currentAccount.chainId,
       })
       beforeBroadcast?.()
-      const hash = await sendTransactionAsync({
-        account: expectedAccount,
-        chainId: expectedChainId,
-        ...(connector ? { connector } : {}),
-        to,
-        data: data as Hex,
-        value: value ?? 0n,
-      })
+      onArm?.()
+      let hash: Hash
+      try {
+        hash = await sendTransactionAsync({
+          account: expectedAccount,
+          chainId: expectedChainId,
+          ...(connector ? { connector } : {}),
+          to,
+          data: data as Hex,
+          value: value ?? 0n,
+        }) as Hash
+      }
+      catch (error) {
+        onSubmitError?.(error)
+        throw error
+      }
       if (isOkx && (data as Hex).toLowerCase().startsWith(ERC20_APPROVE_SELECTOR)) {
         okxDelayPending = true
       }
@@ -1479,7 +1512,7 @@ export const useEulerTx = () => {
    * bundling brings no benefit (fewer than two calls) — callers fall back to
    * sequential execution.
    */
-  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall, beforeBroadcast, onBroadcast }: {
+  const executePlanAsSafeBundle = async ({ plan, chainId, owner, provider, connector, safeWalletProvider, sdk, extraCalls, allowSingleCall, beforeBroadcast, onArm, onSubmitError, onBroadcast }: {
     plan: TransactionPlan
     chainId: number
     owner: Address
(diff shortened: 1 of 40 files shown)


LeonardEulerXYZ (approved):
Approved at exact head `cdba521a04496323a3fa2ac0d0934cb31c15d44c`.

This resolves the #806 Safe calldata-parity blocker:

- the freshly merged bundled execution plan is prepared once when review opens;
- that exact `TransactionPlanPrepared` envelope is retained in the ceremony latch;
- `prepareBatchPlan()` returns the latched envelope to the review/export path;
- `executeBatch()` passes the same envelope, plus the same latched grants and revocations, to the Safe bundle executor;
- cart edits and account/chain resets invalidate the latch.

The regression is discriminating: transplanted onto blocked head `3c73cf214605e0358a997b2c54e271456648307d`, it fails because preparation occurs twice; it passes here and asserts object identity across review and execution.

Validation at this SHA:

- full Vitest suite: 180 files / 1,752 tests passed;
- focused batch/review/call-encoding suites: 73 tests passed;
- canonical Nuxt typecheck passed;
- full lint passed with 0 errors and 6 pre-existing warnings;
- production build passed;
- GitHub lint, typecheck, test, CodeRabbit, Cursor security, and Railway preview checks passed;
- Railway preview root and bounded public API smoke returned HTTP 200.

No blocking findings remain in this remediation diff.

Promotion note: #806 is ready for fresh review only after this commit is merged into `development` and appears in the new immutable promotion head.

LeonardEulerXYZ (changes requested):
Superseding my earlier approval at exact head `cdba521a04496323a3fa2ac0d0934cb31c15d44c` after an independent complement audit found the remaining human-review-side fork.

## Blocking — expanded operation rows still decode the add-time preview, not the latched Safe execution plan

#810 correctly makes **Copy calldata** and Safe submission consume the same latched `TransactionPlanPrepared`. However, the modal's main per-operation rows still prefer `entry.review.displayPlan` (`components/BatchReviewModal.vue:128-140`). Migration entries populate that field from the add-time prepared preview (`pages/position/[number]/migrate.vue:1149-1157`; `pages/position/[number]/borrow/swap.vue:3653-3662`), while `prepareBundledExecution()` rebuilds and latches a fresh simulation/execution plan when review opens (`pages/position/[number]/migrate.vue:1127-1139`; `pages/position/[number]/borrow/swap.vue:3635-3647`).

The new regression demonstrates this state explicitly: it creates a `stale-preview` plan and a distinct `fresh-latched-execution` plan (`tests/composables/useTxBatch.test.ts:1383-1391`). The copied core and submitted core now correctly use the latter, but the expanded operation body still decodes the former. Only the authorization grant/revocation rows are replaced from the latch.

Thus Safe owners can read one human-readable core operation sequence while **Copy calldata** and confirmation use another. The fix should make the exact latched prepared plan (or per-entry latched display plans derived from the same collection pass) own the expanded operation rows as well. Please add a discriminating UI/display regression proving stale preview steps are absent and fresh latched execution steps are shown.

The retry-time partial-source warning disappearance reported separately is a real UX warning but is not part of this funds-sensitive blocker.

kasperpawlowski (dismissed):

LeonardEulerXYZ (changes requested):
Fresh exact-head rereview of `bc1eb43a22356a6a755ecdb9dd869170683304fe`.

The prior blockers are fixed: overlapping preparations are deduplicated, stale completions after ordinary cart/account/chain invalidation are rejected, expanded rows use the ceremony plan, and copy/submission share the same prepared call encoder and wrapper order.

One blocking stale-cart race remains.

## Blocking — an async add can commit without invalidating a ceremony prepared during the add

`addEntry()` increments `batchGeneration` when the edit starts (`composables/useTxBatch.ts:2254-2256`), then awaits `buildPlan()` (`:2279-2282`). During that await, `prepareBundledExecution()` can prepare and cache a ceremony for the incremented generation but the **old entries array**. When the pending add later commits at `:2300-2306`, there is no second invalidation/generation increment.

The old-cart ceremony consequently remains current after the new operation is present and simulated. `executeBatch()` accepts it at `:2652-2667` and submits its prepared plan.

I reproduced this at the exact head with a deferred second `buildPlan()`:

1. Start with one bundled migration entry.
2. Begin adding a second entry and hold its plan promise pending.
3. Prepare a bundled ceremony while the second add is pending; it contains only entry 1.
4. Resolve the add and wait until the cart has two entries / three simulation layers.
5. Call `executeBatch(undefined, oldCeremony)`.

Observed:

- `isBundledExecutionCurrent(oldCeremony) === true` after the second entry commits.
- `executePreparedPlanWithPlainCalls` is called once with the old one-entry ceremony.

This permits the wallet proposal to omit an operation shown in the current cart.

### Required remediation

Make the cart mutation and generation transition atomic from the ceremony's perspective. Suitable approaches include:

- keep review preparation/execution unavailable while any add is pending **and** include that pending-edit state in the ceremony freshness predicate; or
- increment/invalidate again when the built entry is committed, so no ceremony prepared against the pre-commit entries can remain current.

Please add a regression with a deferred `buildPlan()` that asserts a ceremony prepared during the pending add is rejected after commit and never reaches the Safe executor. The test should cover the final executor call, not only the modal disabled state.

All other reviewed invariants pass; this is the only remaining blocker I found.

LeonardEulerXYZ (approved):
Approved exact head `18708ad6455aa75a8197f9528df50b66759ca881`.

The remaining async-add stale-cart blocker is resolved. `addEntry()` now advances the review generation synchronously at the commit edge, immediately before appending the built entry, so a ceremony prepared against the prior cart while `buildPlan()` was pending becomes stale before the new operation is visible.

The committed regression covers the full irreversible path: deferred second `buildPlan()`, old-cart ceremony preparation, second-entry commit, stale-current assertion, public `executeBatch()` attempt, and proof that `executePreparedPlanWithPlainCalls` is not called.

Validation:

- named regression at current head: passed;
- same regression overlaid onto prior head `bc1eb43a…`: failed for the expected stale-current reason;
- focused race/parity cohort: 4 files / 97 tests passed;
- Nuxt typecheck: passed;
- targeted ESLint: passed;
- CI, security review, CodeRabbit, and Railway preview: green.

The prior overlapping-preparation, display-plan, copy/submission parity, and pending-add findings are all closed. I found no remaining blocker in #810.

LeonardEulerXYZ (changes requested):
Fresh exact-head rereview of `7dd04c30d977b402141621738385f77b39cb83e6`.

The CodeRabbit findings addressed by this commit are fixed: pending queued adds now block review/execution until all settle; add-time plans are rejected across account, chain, clear, and remove transitions; Safe classification invalidates bundled ceremonies; and delayed add-time snapshot/slot-hint writes are context-guarded.

One separate stale-context blocker remains.

### Blocker: stale resimulation can repopulate `baseAccountSnapshot` after an account reset

`resimulate()` captures `owner`/`chainId`, then can await `fetchBaseAccountSnapshot()` at `composables/useTxBatch.ts:1831-1833`. It assigns the result to the module-scoped `baseAccountSnapshot` at line 1834 **before** checking `token !== resimToken` at lines 1866-1873.

That permits this ordering:

1. Account A has a cart and its resimulation suspends in `fetchBaseAccountSnapshot()`.
2. The wallet switches to account B. The watcher at lines 2177-2194 invalidates the resimulation, clears the cart, and sets `baseAccountSnapshot = null`.
3. A's old fetch resolves. The stale resimulation writes A's snapshot back at line 1834; its later token check only prevents layer publication, not the cache write.
4. A new empty-cart add starts under B. `getEntryPlanningAccount()` returns the repopulated module cache at lines 2269-2271 without checking that the cached account belongs to B.
5. B's `buildPlan()` receives A's exact `Account` object. The new `BatchAddContext` assertions pass because the add itself never leaves B.

I reproduced this at exact head with a disposable deferred-fetch Vitest probe. After switching A → B and resolving A's old fetch, the subsequent B add received the exact old `Account` object; the assertion `expect(planningAccount).not.toBe(oldAccount)` failed (`1 failed | 69 skipped`). The probe was removed and the worktree restored byte-for-byte clean.

Guard the `baseAccountSnapshot` write with the captured resimulation token plus owner/chain context, or bind the cached snapshot itself to owner+chain and validate on every read. A regression should suspend A's resimulation fetch, switch to B, resolve A, then prove B's next `buildPlan()` cannot receive A's snapshot.

Validation otherwise passed: focused race/parity cohort 4 files / 103 tests, typecheck, targeted ESLint, production build, CI, CodeRabbit, security review, and Railway preview.

LeonardEulerXYZ (changes requested):
Supplement to my exact-head `7dd04c30d977b402141621738385f77b39cb83e6` changes-requested review. Two additional execution-boundary blockers were independently identified and reproduced after the initial review.

### 1. A pending add can start after `executeBatch()`'s entry gate and the old sequential cart still submits

`executeBatch()` checks `pendingAddCount` only on entry (`composables/useTxBatch.ts:2733-2736`), then awaits `restorePendingBeforeRetry()` at line 2746. If an add starts during that await, it invalidates review state and increments `pendingAddCount`, but the non-bundled path performs no second generation/pending-add check before prerequisite broadcasts and final execution at lines 2783-2787.

Exact-head deferred probe:

1. Create and simulate one ordinary entry.
2. Start `executeBatch()` and suspend `restorePendingBeforeRetry()`.
3. Start a second deferred add; verify `hasPendingAdds === true`.
4. Release restore.

Expected: no irreversible execution for the old cart. Observed: `executePreparedPlan` was called once. The pending add is rejected only later because successful execution clears its context; the old cart has already been submitted.

This affects EOA and non-bundled execution. Revalidate a captured execution/cart generation, wallet context, and pending-add state after every pre-broadcast await and immediately before each irreversible grant/core broadcast.

### 2. Ordinary Safe reviews have no owned ceremony; reviewed/copied and executed envelopes can diverge

For a Safe cart without prerequisite-bearing entries, `prepareBundledExecution()` returns `null` (`useTxBatch.ts:2597-2602`). The modal then prepares and exports an envelope through `prepareBatchPlan()` (`BatchReviewModal.vue:397-400, 433-448`), but confirmation passes no ceremony and `executeBatch()` independently rebuilds and prepares another envelope at `useTxBatch.ts:2783-2787`.

Exact-head probe made the first preparation return `{ kind: 'reviewed' }` and the execution-time preparation return `{ kind: 'rebuilt' }`. `executePreparedPlan` received the rebuilt envelope, not the reviewed one.

The same null path has no freshness token: `BatchReviewModal.vue:383-385` treats `bundledExecutionRef === null` as unconditionally current. Cart or Safe-classification changes can therefore leave copied UI state stale while confirmation rebuilds current state.

Latch an immutable generation/owner/chain/Safe-mode/prepared-envelope ceremony for every review mode, including ordinary Safe and sequential flows, and consume that exact artifact or abort/re-review.

Reviewer-only probes were removed; the exact-head worktree is restored clean and its test blob matches Git.

LeonardEulerXYZ (changes requested):
Exact-head re-review of `18718286d1c0d10266410ca1f4b7eba138695c31`. Most previously reported boundaries are now closed: signatures wait for confirmation, wallet/connector and generation freshness reaches the final broadcast, review/copy use ceremony-owned plans, and stale completion side effects are mostly suppressed. Four parity/lifecycle issues remain.

Validation on the exact head: focused transaction cohort 6 files / 121 tests passed; full Vitest 181 files / 1,781 tests passed; typecheck, lint, production build, `git diff --check`, all GitHub checks, Railway render, and authenticated app shell smoke passed. A first local build hit `ENOSPC`; after removing only the clean superseded `554b6bcc…` review worktree, the unchanged build passed.

Reviewer probes reproduced (1) an unrestricted post-confirm plan replacement, (2) two duplicate standalone grants/restorations being executed while the summary consolidates one restoration row, and (3) a stale decoded error writing into a successor wallet context. Independent path tracing also found that the Tenderly action simulates module-level `lastMerged` from add-time resimulation rather than the immutable review/execution ceremony. Please keep this blocked until the inline findings are resolved with discriminating regressions.

LeonardEulerXYZ (approved):
Approved at exact head `abcf73dd39ea30b16c845ecc668fcda5fff73350`.

Verified the two latest remediations against the production paths:
- confirm-time migration parity now accepts only matching raw or ABI-encoded `v/r/s` signature substitutions while still rejecting target, value, ordering, approval, plugin-call, and adjacent-calldata drift;
- final standard-path gas estimation uses `estimateGasForPreparedTransactionPlan(prepared)`, so it estimates the same frozen prepared envelope that is broadcast without rerunning TOS, Keyring, or Pyth plugins.

The earlier consent ordering, connector/freshness, duplicate authorization/restoration visibility, stale decoded-error ownership, and whole-batch Tenderly findings remain resolved.

Validation on the immutable head: 137 focused tests across 7 files; full Vitest 1,793 passed and 1 skipped across 183 files; typecheck passed; ESLint passed with 0 errors and 6 pre-existing warnings; production build and diff checks passed; Railway app-shell smoke passed with no JavaScript errors; all GitHub checks are green.

pages/position/[number]/migrate.vue
```diff
@@ -609,13 +610,17 @@ function buildMigrationInput(
...
+function mintMigrationDeadline(): bigint {
+  return BigInt(Math.floor(Date.now() / 1000) + 60 * 60)
+}
+
 async function getAuthorizationRequest(
   input: OutgoingMigrationInput,
   migrationPosition?: MigrationPosition,
   account?: Account<IHasVaultAddress>,
   useSignatures = signaturesEnabled.value,
+  deadline: bigint = mintMigrationDeadline(),
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/migrate.vue]: **[High] Pin the deadline for standalone review and confirmation too.** This default is overridden only by the batch ceremony. `prepareOutgoingMigrationPreview()` calls the authorization and simulation builders without a deadline, then `sendMigration()` calls the authorization and execution builders again without one. Once the clock advances, the fresh typed-data request no longer matches the reviewed request and the payload-key gate throws `Authorization requirements changed since review` before the signed migration can proceed. The inbound standalone path has the same split at `swap.vue:2943` and `swap.vue:3441`. I reproduced the invariant directly: changing only the typed-data deadline by +1 second changes `migrationAuthorizationPayloadKey`. Mint once when preparing the preview, store it on the preview, and reuse it for fresh authorization validation plus the simulated/displayed/executed plan in both directions. Please add fake-clock regressions for direct outgoing and inbound confirmation.

pages/position/[number]/borrow/swap.vue
```diff
@@ -2749,15 +2750,28 @@ const buildInboundExternalMigrationInput = async (): Promise<InboundExternalMigr
...
+// A supplied plan deadline replaces the quote-derived verifier deadlines the
+// connectors would otherwise use, so clamp it to the quotes' own bounds —
+// pinning a ceremony deadline must never extend how long a reviewed swap
+// stays executable on-chain.
+const pinInboundMigrationCeremonyDeadline = (input: InboundExternalMigrationInput): bigint =>
+  [input.collateralSwapQuote, input.debtSwapQuote]
+    .map(quote => quote?.verify.deadline)
+    .filter((deadline): deadline is number => typeof deadline === 'number' && deadline > 0)
+    .map(deadline => BigInt(deadline))
+    .reduce((earliest, deadline) => (deadline < earliest ? deadline : earliest), mintInboundMigrationDeadline())
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/borrow/swap.vue]: **[High] Do not replace the collateral quote's verifier deadline with an older debt-quote deadline.** Collateral and debt quotes are fetched and refreshed independently (`onRefreshExternalCollateralQuotes` / `onRefreshExternalDebtQuotes`), so their deadlines can differ. The locked SDK passes this supplied value into `encodeSwapQuoteVerificationItem(collateralSwapQuote, verifierDeadline)`, whose verifier-data check requires exact equality with the deadline encoded in that collateral quote. If debt expires at `1000` and collateral at `2000`, this reduction supplies `1000` and planning throws `SwapVerifier data mismatch`; I reproduced that directly against SDK 2.0.0, while `2000` passes. Preserve the exact collateral quote deadline when its immutable verifier data is emitted (and let the debt quote retain its own embedded expiry); use a minted deadline only when no such quote governs the verifier. Add an inbound two-quote regression with distinct deadlines.

composables/useTxBatch.ts
```diff
@@ -2551,76 +2812,118 @@ export const useTxBatch = () => {
...
+          allowSingleCall: true,
+          beforeBroadcast: () => assertBatchExecutionCurrent(execution),
+        })
         if (!result) {
           // The review described ONE proposal; never silently degrade to
           // the sequential multi-proposal ceremony.
           throw new Error('Safe connection unavailable — the reviewed single-proposal submission cannot run. Reconnect your Safe and retry.')
         }
-        latchedBundledExecution.value = null
+        if (!isBatchExecutionCurrent(execution)) {
```
LeonardEulerXYZ: **[High] A successful execution must retire its own entries even if a same-wallet cart edit advanced the generation.** `addEntry()` is allowed while an execution is pending and invalidates this ceremony. If the already-broadcast transaction later succeeds, this branch marks success and returns without clearing or removing the executed entries. The new same-wallet entry is therefore left beside operations that already landed, and a later Execute can submit them again. I reproduced this at the exact head with a gated execution: add `Already executed operation`, start execution, add `New same-wallet operation`, release a successful completion; the cart remains `[Already executed operation, New same-wallet operation]`. The existing late-completion test changes wallets, whose watcher first clears the old cart, so it does not cover this case. Either centrally reject edits while an execution is pending, or snapshot entry IDs in the ceremony and remove exactly those IDs after known success while preserving successor entries. Cover both Safe-bundled and standard completion paths.

LeonardEulerXYZ (changes requested):
Exact-head re-review of `8c56f8504d9a90f166042e8135dcbdd181af0b43`. The new batch-only deadline latch closes the reported placeholder/signed-plan drift for that ceremony, but this head is not promotion-ready.

Three blocking execution paths remain:
1. standalone outgoing and inbound signed migrations still mint a new authorization deadline between review and confirmation, so their payload keys self-invalidate after the clock advances;
2. inbound batch pinning can substitute an older debt-quote deadline into immutable collateral-quote verifier data and fail planning;
3. a same-wallet cart edit during a pending execution can leave successfully executed entries in the cart, allowing a later duplicate submission.

Verification at this exact head: full Vitest `182 passed / 1 skipped` files (`1793 passed / 1 skipped` tests), focused lifecycle suite `6 files / 131 tests`, typecheck, lint, and production build all passed. The PR preview loaded normally. I also ran discriminating probes for the deadline mismatch and retained-entry execution state; details are inline.

utils/preparedPlanParity.ts
```diff
@@ -0,0 +1,165 @@
...
+  resolved: unknown,
+  substitutions: ReadonlyMap<string, SignatureSubstitutionGroup>,
+  observed: Map<string, number>,
+): void => {
+  if (typeof reviewed === 'string' && typeof resolved === 'string') {
+    const normalizedReviewed = reviewed.startsWith('0x') ? reviewed.toLowerCase() : reviewed
+    let normalizedResolved = resolved.startsWith('0x') ? resolved.toLowerCase() : resolved
+    if (normalizedReviewed === normalizedResolved) return
+    if (normalizedReviewed.startsWith('0x') && normalizedResolved.startsWith('0x')) {
+      for (const [signature, substitution] of substitutions) {
```
LeonardEulerXYZ: **[Medium] Bind each signature substitution to its reviewed structural occurrence.** This loop tries every real signature at every placeholder-shaped location, while the final check validates only global counts per signature. With two identical reviewed placeholders and two distinct signatures, swapping signature A and B between their call locations is accepted. I reproduced this directly at the exact head: `assertPreparedPlanSignatureParity` returned successfully for the reversed plan. Those signatures authorize different typed-data messages, so the accepted plan normally fails only after the user has supplied both signatures—the failure class this PR is meant to eliminate. Carry a structural path/occurrence index with each substitution, or deterministically pair ordered reviewed occurrences with ordered resolved signatures before deep equality. Add a two-authorization regression that reverses the signatures and must reject.

composables/useEulerTx.ts
```diff
@@ -1694,7 +1717,7 @@ export const useEulerTx = () => {
       before?: readonly PlainTxRequest[]
       after?: readonly PlainTxRequest[]
     },
-    options?: { allowSingleCall?: boolean },
+    options?: { allowSingleCall?: boolean, beforeBroadcast?: () => void },
```
LeonardEulerXYZ: **[Medium] Use this final freshness hook in the direct Safe migration callers too.** Batch execution supplies `beforeBroadcast`, but `sendMigrationAsSafeBundle()` and `sendInboundExternalMigrationAsSafeBundle()` call this helper without it. Their account/chain checks occur before authorization lookup, planning, preparation, and simulation; neither flow binds the reviewed connector key. A same-account/same-chain Safe connector switch during those awaits can therefore submit through a connector different from the reviewed context (the helper captures the then-current connector at line 1730). Capture the review connector/Safe classification and revalidate it through `beforeBroadcast` for both direct migration directions, with a delayed-preparation connector-switch test.

LeonardEulerXYZ (commented):
Supplement to my exact-head changes-requested review after the independent complement lanes completed. Two additional execution-boundary issues are inline below.

The earlier exact-ceremony simulation finding also remains live at this head: continuous simulation and wallet/health summaries still use immutable add-time plans, while review preparation can rebuild a different ceremony plan that is displayed, copied, and submitted—especially for migrations. Safe execution skips an exact-plan estimate and relies on that older simulation. Existing thread: https://github.com/euler-xyz/euler-lite/pull/810#discussion_r3782988442

This supplement does not change the verdict: `8c56f8504d9a90f166042e8135dcbdd181af0b43` remains blocked.

composables/useTxBatch.ts
```diff
@@ -2436,49 +2557,217 @@ export const useTxBatch = () => {
...
+      })))
+      const ceremony: PreparedBatchExecution = Object.freeze({
+        generation,
+        owner: reviewOwner,
+        chainId: currentChainId,
+        safeWallet: reviewSafeWallet,
+        safeWalletResolved: reviewSafeWalletResolved,
+        connectorKey: reviewConnectorKey,
+        mode,
+        prepared,
```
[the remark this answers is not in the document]
Seranged: Resolved structurally rather than by rebinding the simulation: the batch review modal no longer exposes a whole-batch Tenderly action at all.

The reviewed ceremony is a multi-transaction call vector — sequential authorization grants, the prepared plugin/approval/core plan, and restoration calls. The Tenderly integration here simulates a single transaction (`buildTenderlySimulationPayload` produces one `from`/`to`/`data`/`value`, and the internal simulate endpoint accepts exactly that), so no payload built from any single plan — `lastMerged` or the ceremony core — can represent what actually executes. Binding the button to the ceremony core would still drop the before/after calls and report on a partial vector while presenting it as whole-batch simulation.

Current state on this branch:
- `components/BatchReviewModal.vue` has no Tenderly action; `docs/batch-review-modal-handoff.md` ("External simulation") records why.
- `tests/components/batchReviewModal.test.ts` pins the absence ("does not offer a whole-batch simulation for a different call vector") and asserts the expanded per-entry rows derive from the ceremony's `reviewByEntryId` plans, not the add-time preview cache.
- The displayed rows, copied calldata, and execution all come from the one immutable `PreparedBatchExecution`, so nothing shown in review is sourced from `lastMerged`.

Restoring a trustworthy whole-batch simulation would need bundle-simulation support (Tenderly simulate-bundle) in the server API and payload builder so the grants, prepared core, and restorations simulate as one sequence. That is a separate feature; happy to spec it in a follow-up if wanted.

pages/position/[number]/migrate.vue
```diff
@@ -947,7 +983,15 @@ async function sendMigration(execution: TrackedExecutionScope, preview: Outgoing
           invalidateOutgoingMigrationPreview(preview.key)
           throw new Error('Authorization requirements changed since review — please review the migration again.')
         }
-        const outcome = await sendMigrationAsSafeBundle(input, migrationPosition, authorizationRequest, reviewedAccount)
+        const beforeBroadcast = createSafeBundleBroadcastGuard({
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/migrate.vue]: **[Medium] Snapshot the reviewed connector before the first confirmation await, not here.** This factory runs only after `restorePendingBeforeRetry()` and `getAuthorizationRequest(...)` have both awaited. If the user switches to another Safe connector with the same account and chain during either await, `confirmedConnectorKey` captures the replacement connector and the final guard passes. The preview stores no connector key, and the current utility tests construct the guard before mutating state, so they cannot catch this call-site race. Bind the connector/Safe context in `OutgoingMigrationPreview` (with the same change for the inbound preview), revalidate that expected key at broadcast, and cover a delayed authorization lookup where the connector changes before this line.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `2f48514c1ab4e9384fc47e26d70a502d9700e67f` from the prior blocked head `8c56f8504d9a90f166042e8135dcbdd181af0b43`.

The deadline pinning, collateral-verifier deadline selection, positional signature parity, successful-entry retirement, and final Safe-bundle guard are real improvements, and their focused regressions pass. I also consider the earlier whole-batch Tenderly objection closed: the modal no longer claims to externally simulate a multi-transaction ceremony, while review rows, copied calldata, signatures, and submission consume the immutable `PreparedBatchExecution`.

One connector-binding gap remains in both direct Safe migration directions. The new guard snapshots the connector only after confirmation has already awaited pending restoration and fresh authorization lookup/payload comparison. A same-account/same-chain Safe connector switch completed during those awaits becomes the guard's accepted baseline, so the proposal can be submitted through a connector/session that was not the reviewed one. The utility tests mutate wallet state only after constructing the guard and therefore do not exercise this call-site timing. Capture the connector/Safe context in the reviewed preview (or at minimum at confirmation entry before the first await), pass that expected context through to the final broadcast check, and add delayed-authorization connector-switch tests for both outgoing and inbound direct Safe flows.

Validation at this head:
- focused lifecycle cohort: 147/147 passed
- full Vitest suite: 184 files, 1,814 tests passed
- `npm run typecheck`: passed
- `npm run lint`: passed with 6 pre-existing `no-explicit-any` warnings, 0 errors
- `npm run build`: passed
- hosted lint/typecheck/test/security checks: green; preview deployment is present

Requesting changes on that remaining irreversible-boundary binding issue.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `3b7909671e3cb754b15b053d94fdea866f3448a8`. The new preview-owned connector key correctly fixes the previously reported direct Safe guard-construction race and now fails closed when the review had no connector identity.

The completed independent complement audit identified four separate lifecycle blockers that this one-purpose delta does not address:

1. Direct outgoing and inbound confirmation rebuild fresh prepared envelopes without proving parity with the reviewed/copied prepared envelope.
2. Non-signature batch prerequisites freeze review-time grant/revoke restoration state and do not reject authorization drift before broadcast.
3. A core transaction/Safe proposal accepted by the wallet but followed by receipt/status-poll failure is treated as never submitted; the cart remains immediately retriable.
4. Standard direct migrations do not bind the reviewed connector/session through signing and final broadcast.

Focused exact-head validation: 4 files / 113 tests passed; `npm run typecheck` passed; diff hygiene is clean. Requesting changes on the four remaining irreversible-boundary issues.

pages/position/[number]/migrate.vue
```diff
@@ -1094,46 +1150,56 @@ async function addPreparedMigrationToBatch(preview: OutgoingMigrationPreview) {
...
+                return {
+                  plan: await buildMigrationPlan(input, authorization, migrationPosition, account, ceremonyDeadline),
+                  signatureSubstitutions: getMigrationAuthorizationSignatureSubstitutions(authorization),
+                }
+              },
+              usesPlaceholderSignatures: !!request,
+              grantSteps: buildSignatureSteps(input.target, request, true, false),
+              revokeSteps: [],
             }
           },
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/migrate.vue]: **[Medium] Revalidate non-signature authorization state before sending the frozen prerequisites.** This request and its restoration values are resolved during review, then `PreparedBatchExecution` sends the frozen grants/revokes later without the payload-key drift check used by direct migrations. If an allowance changes meanwhile, the post-transaction revoke can overwrite the newer value with review-time state—particularly material for pending Safe proposals. Re-resolve at the last pre-broadcast boundary, compare the authorization payload key, and require re-review on drift; mirror this for inbound batches.

composables/useTxBatch.ts
```diff
@@ -2551,76 +2846,120 @@ export const useTxBatch = () => {
...
+        return
+      }
       clearBatch()
       if (shouldRefreshExternalMigrationPositions) scheduleExternalMigrationRefreshes()
-      await revokeAfterSuccess(grantedRevokes)
       await redirectAfterBatchExecution(scope)
     }
     catch (error) {
       logWarn('useTxBatch/executeBatch', error)
       // The batch never landed, so no granted authorization should be left
```
LeonardEulerXYZ: **[Medium] Do not classify every execution rejection as ‘the batch never landed’.** `executePreparedPlan` and Safe bundle execution can reject after the wallet accepted a transaction/proposal—for example receipt or Safe-status polling failure. This catch then leaves all entries immediately retriable, so a retry can duplicate a landed borrow/withdraw/migration. Track the core hash/proposal ID at submission (the prerequisite path already has an `onBroadcast` pattern), distinguish pre-submit failure from submitted-but-unconfirmed, and quarantine/reconcile ambiguous entries instead of offering an immediate replay.

pages/position/[number]/migrate.vue
```diff
@@ -947,7 +991,16 @@ async function sendMigration(execution: TrackedExecutionScope, preview: Outgoing
...
+          expectedAccount: reviewedInput.owner,
+          expectedChainId: target.chainId,
+          expectedConnectorKey: preview.connectorContextKey,
+          currentAccount: () => address.value as Address | undefined,
+          currentChainId: () => walletChainId.value,
+          isSafeWallet: () => isSafeWallet.value,
+          connectorContextKey: walletConnectorContextKey,
+        })
+        const outcome = await sendMigrationAsSafeBundle(input, migrationPosition, authorizationRequest, reviewedAccount, preview.authorizationDeadline, beforeBroadcast)
         if (outcome === 'aborted') return
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/migrate.vue]: **[Medium] Bind the reviewed connector/session for standard direct execution too.** The Safe-only branch has the preview-owned connector guard, but this signature and the later `executePreparedPlan` run after several awaits with only account/chain checks. A same-account/same-chain connector switch is accepted and the currently active session performs the wallet actions. Capture the reviewed connector context, revalidate it before every signature and irreversible broadcast, and cover delayed authorization/planning switches in both outgoing and inbound direct flows.

pages/position/[number]/borrow/swap.vue
```diff
@@ -3452,14 +3500,23 @@ const sendInboundExternalMigration = async (execution: TrackedExecutionScope, pr
...
+          expectedAccount: input.owner,
+          expectedChainId: migrationChainId,
+          expectedConnectorKey: preview.connectorContextKey,
+          currentAccount: () => address.value as Address | undefined,
+          currentChainId: () => walletChainId.value,
+          isSafeWallet: () => isSafeWallet.value,
+          connectorContextKey: walletConnectorContextKey,
+        })
+        const outcome = await sendInboundExternalMigrationAsSafeBundle(input, authorizationRequest, account, useSignatures, preview.authorizationDeadline, beforeBroadcast)
         if (outcome === 'aborted') return
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/borrow/swap.vue]: **[High] Prove that the confirm-time prepared envelope matches the reviewed/copied envelope.** The modal exposes `preview.calldataPrepared`, but confirmation independently signs, rebuilds, and prepares here before submission. Deadline and authorization-payload equality do not constrain plugin output, resolved approvals, targets, values, or call ordering. Apply ordered `assertPreparedPlanSignatureParity`-style validation against `preview.calldataPrepared` (and do the same in the inbound direct flow), permitting only the reviewed authorization-signature substitutions before simulation or broadcast.

composables/useTxBatch.ts
```diff
@@ -2552,75 +2887,251 @@ export const useTxBatch = () => {
...
+      assertBatchExecutionCurrent(execution)
+      // Final on-chain gas estimate before the irreversible core broadcast. If
+      // chain state moved since review, surface the decoded reason and retain
+      // the cart. A pending edit during this await invalidates the ceremony.
+      await estimateGasForPreparedPlan(prepared)
+      assertBatchExecutionCurrent(execution)
+      await executePreparedPlan(prepared, {
+        beforeBroadcast: () => assertBatchExecutionCurrent(execution),
+        onBroadcast: (broadcast) => {
+          coreBroadcast = broadcast
```
LeonardEulerXYZ: **[High] Track the actual execution sequence, not the latest prepared-plan broadcast.** `executePreparedPlan` invokes this callback for every transaction emitted by the SDK executor, including plugin calls and resolved ERC-20 approvals before the EVC batch. If one of those succeeds and the wallet rejects the later core send (or its RPC call fails before returning a hash), `coreBroadcast` still contains the earlier hash. The catch quarantines that hash as the core; reconciliation sees its successful receipt and `retireExecutedEntries` removes the whole cart even though the value-moving batch never ran. The reverse partial-execution cases also cannot be represented by one mutable hash. Carry item/status/index through the broadcast callback and reconcile the ordered prepared-plan execution state; only retire entries after every required economic item is terminal-success. Add a regression with successful plugin/approval broadcast followed by rejected core submission.

composables/useTxBatch.ts
```diff
@@ -2331,12 +2474,16 @@ export const useTxBatch = () => {
...
       syncOverlay()
     }
   }
 
   const clearBatch = () => {
+    invalidatePendingAddContext()
+    invalidateBatchReview()
+    // Emptying the cart discards the entries a quarantined submission covered,
+    // so a replay can no longer duplicate them — the quarantine is moot.
+    pendingCoreSubmission.value = null
```
LeonardEulerXYZ: **[High] Do not let cart clearing erase an unknown on-chain outcome.** “Clear batch” is directly exposed in `BatchContents.vue`, and this assignment discards the only hash/proposal record while the original submission may still land. The user can then re-add the same borrow, withdraw, or migration and submit it again; clearing entry IDs does not make the underlying operation non-repeatable. A page reload likewise loses this module-scoped ref. Keep the quarantine independent of cart contents and persist the minimal wallet/chain/submission identity across clear, account switches, and reloads until reconciliation reaches a terminal verdict. Cover clear-and-readd plus reload/reconnect retries.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `5158b2022a7704d83ac945b881b536af1b37121c` against all four findings from my previous review. Prepared-envelope parity, authorization prerequisite revalidation, and reviewed-connector binding are now resolved. The ambiguous-submission remediation remains unsafe in two independent ways: it can mistake a successful pre-core transaction for completion of the whole batch, and its quarantine is user-clearable/non-durable. Both permit incorrect retirement or duplicate execution of value-moving entries, so this head remains blocked.

Exact-head validation passed: focused transaction cohort 6 files / 156 tests; full Vitest 184 files passed, 1 skipped / 1,824 tests passed, 1 skipped; typecheck; lint with 0 errors and 6 existing warnings; production build; diff check. The new quarantine/reconciliation behavior has no discriminating tests for these lifecycle cases.

pages/position/[number]/migrate.vue
```diff
@@ -947,34 +1023,56 @@ async function sendMigration(execution: TrackedExecutionScope, preview: Outgoing
...
         usePermit2: useSignatures,
       })
+      assertMigrationPreparedParity(preview, prepared, authorization)
       const ok = await runPreparedSimulation(prepared, buildStateOverrideOptions({ noBalanceOverride: true }))
       if (!ok) {
         await revokeAfterAbort(revokeTxs)
         return
       }
-      await executePreparedPlan(prepared)
+      await executePreparedPlan(prepared, { beforeBroadcast: assertReviewedWalletContext })
```
Comment (LeonardEulerXYZ) [inline pages/position/[number]/migrate.vue]: **[High] Quarantine accepted direct migration submissions too.** This direct path passes only `beforeBroadcast`; it does not consume `onBroadcast` or retain the accepted transaction hash. The direct Safe-bundle helper likewise omits the proposal callback, and the inbound paths mirror both omissions. If the wallet accepts the migration and receipt/Safe-status polling then fails or remains unknown, the catch performs abort cleanup and presents an ordinary retry even though the first migration may still execute. That can duplicate a borrow, withdrawal, or migration. Give direct execution a submission owner keyed by wallet/chain/hash or Safe proposal ID, reconcile terminal status before enabling another attempt, and add EOA receipt-failure plus Safe unknown-status regressions for both directions.

composables/useEulerTx.ts
```diff
@@ -1019,9 +1030,10 @@ export const useEulerTx = () => {
       currentAccount: currentAccount.address,
       currentChainId: currentAccount.chainId,
     })
+    options?.beforeSignature?.()
     const signature = await signTypedDataAsync(request.typedData as unknown as Parameters<typeof signTypedDataAsync>[0])
```
LeonardEulerXYZ: **[Medium] Pin typed-data signing to the reviewed connector instead of checking then using the active one.** `beforeSignature` verifies the reviewed connector, but `signTypedDataAsync` is then called without its supported `connector` parameter; Wagmi resolves the active connector asynchronously. A same-account/same-chain connector switch after the guard can therefore route the signature request through the replacement session. The prepared-plan Permit2 signer at `executePreparedPlan` has the same pattern, while transaction sends correctly pass the captured connector. Carry the reviewed connector object through the ceremony and pass it to every typed-data mutation; add an integration test that switches connectors between the guard and connector-client resolution.

LeonardEulerXYZ (commented):
Supplement after the independent exact-head complement lanes completed. Correction to my main review summary: prepared-envelope parity remains substantively resolved, but connector continuity is not fully closed because typed-data mutations are checked and then dispatched through the unpinned active connector. The ambiguity remediation is also batch-only; direct outgoing/inbound migrations and their Safe proposals retain no accepted-submission identity. The existing `CHANGES_REQUESTED` verdict is unchanged.

utils/directSubmissionQuarantine.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Arm tracking for one attempt. Must run before any wallet action. */
+  const begin = (input: { owner: Address, chainId: number }) => {
+    attempt = { owner: getAddress(input.owner), chainId: input.chainId }
+    unconfirmed = undefined
+  }
+
+  /** onBroadcast sink for executePreparedPlan / …WithPlainCalls. */
+  const track = (broadcast: PreparedPlanBroadcast) => {
+    unconfirmed = broadcast.status === 'submitted' ? broadcast : undefined
```
LeonardEulerXYZ: **[P1] Persist the quarantine at wallet acceptance, before waiting for confirmation**

`track()` only updates this in-memory variable. The durable write does not happen until `sealFailure()`, which the callers invoke only after the awaited executor rejects. If the user reloads/closes the tab or the renderer dies after the wallet has returned the transaction hash/Safe proposal ID but while receipt/status polling is still pending, the catch never runs, storage remains empty, and the next page load treats the flow as clear and can submit the operation again while the first submission may still land. The batch tracker has the same acceptance-to-catch gap.

There is also no fail-closed path when `localStorage.setItem` is unavailable or throws: `writePendingSubmission` swallows it and `sealFailure()` still reports success. I reproduced the reload window with a temporary discriminating test: call `begin`, deliver a value-moving `submitted` broadcast, instantiate a fresh quarantine as a reload would, and `reconcileBeforeAttempt` returns `clear`.

Please persist synchronously when the value-moving `submitted` callback fires (and verify/fail closed if durable storage is unavailable), then clear only on a terminal confirmed/reverted verdict. Add regressions for reload between the submitted callback and executor settlement, plus storage-write failure.

composables/useTxBatch.ts
```diff
@@ -2331,12 +2492,17 @@ export const useTxBatch = () => {
...
       resimulatePromise = null
-      tenderly.clearSimulation()
       syncOverlay()
     }
   }
 
   const clearBatch = () => {
+    invalidatePendingAddContext()
+    invalidateBatchReview()
+    // A quarantined core submission deliberately survives cart clearing: the
```
LeonardEulerXYZ: **[P1] Gate equivalent direct operations while the batch submission is unresolved**

Keeping the record through `clearBatch()` is necessary but not sufficient. This empties `entries`; `VaultFormSubmit.vue` pauses direct execution only while `entryCount > 0`, and `hasPendingCoreSubmission` has no production consumer. The user can therefore clear an ambiguously submitted borrow/withdraw batch and immediately execute the equivalent operation through the direct form without ever invoking `reconcilePendingCoreSubmission()`. The first batch may then confirm as well. Migration records are likewise split among `batch`, `outgoing-migration`, and `inbound-migration`, so changing direct/batched execution mode walks around the flow-specific gate.

Please enforce unresolved-submission quarantine at the shared value-moving action boundary (at least owner+chain, and operation identity if safely available), not only on the next invocation of the same executor. A regression should exercise ambiguous batch submission → clear cart → attempt the equivalent direct operation, and the reverse direct-migration → add-to-batch path.

utils/pendingSubmissions.ts
```diff
@@ -0,0 +1,188 @@
...
+    // the flow forever on unparseable state.
+    clearPendingSubmission(flow)
+    return undefined
+  }
+  return record
+}
+
+export const writePendingSubmission = (flow: PendingSubmissionFlow, record: PendingSubmissionRecord) => {
+  try {
+    getStorage()?.setItem(storageKey(flow), JSON.stringify(record))
```
LeonardEulerXYZ: **[P1] Do not overwrite another wallet's unresolved submission**

Storage has one slot per flow (`euler_pending_submission:${flow}`), while reconciliation intentionally lets a different owner/chain proceed without resolving the existing record. If wallet A has an ambiguous submission and wallet B later has one in the same flow, this write replaces A's only durable record. When A reconnects, it sees B's record as unrelated and proceeds as clear even though A's original transaction can still confirm. I reproduced exactly that sequence with the real quarantine helper. Multiple tabs create the same class of stale/competing record.

Please retain a collection keyed by flow + owner + chain + submission identity, or conservatively block new value-moving submissions while any unresolved record could overlap. Add a regression where A is quarantined, B becomes ambiguous, and both records remain independently reconcilable.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `d27c99987dd3c82e77888df6a1d8e4f2d3942788` against all prior Leonard findings and with a fresh complement pass. Prepared-plan parity, authorization revalidation, prerequisite/core attribution, and connector-pinned signing are resolved. The new quarantine closes the ordinary retry path, but it is still bypassable before persistence, across execution surfaces, and when another wallet overwrites the flow-wide record.

Validation: focused lifecycle cohort 218/218; full Vitest 1,886 passed / 1 skipped; typecheck passed; lint 0 errors / 6 existing warnings; production build passed; `git diff --check` passed. I also ran temporary adversarial regressions demonstrating acceptance-time reload loss and cross-wallet overwrite, then removed them; the exact-head worktree is clean.

composables/useTxBatch.ts
```diff
@@ -253,6 +313,36 @@ const isSimulating = ref(false)
...
+ * no ceremony, so a landed verdict then resets the cart instead of retiring
+ * the exact covered entries.
+ */
+const pendingCoreSubmission = shallowRef<{
+  record: PendingSubmissionRecord
+  context?: { ceremony: PreparedBatchExecution }
+} | null>(null)
+
+/**
+ * Rehydrate the quarantine mirror from its durable record. Runs once at
```
LeonardEulerXYZ: **[P1] Refresh the durable quarantine before every batch attempt, not only at module initialization**

An already-open tab hydrates this mirror once. If tab A later stores an ambiguous batch submission, tab B keeps its stale `null` mirror: `reconcilePendingCoreSubmission()` consults that mirror rather than re-reading storage, and there is no `storage` listener or `BroadcastChannel`. Tab B can therefore broadcast the same reviewed operations while A's transaction may still confirm. Reload coverage does not exercise this lifecycle because reload itself re-runs this initializer.

Please re-read/merge durable records immediately before every value-moving attempt and synchronize already-open tabs (while still handling the simultaneous-attempt race conservatively). Add a two-module/two-tab regression over shared storage: initialize both, quarantine in A, then assert B performs zero wallet broadcasts.

composables/useEulerTx.ts
```diff
@@ -1474,6 +1541,7 @@ export const useEulerTx = () => {
       currentAccount: currentAccount.address,
       currentChainId: currentAccount.chainId,
     })
+    beforeBroadcast?.()
```
LeonardEulerXYZ: **[P1] Arm replay protection before crossing the wallet boundary**

The current model can quarantine only after a hash-shaped ID is returned. But `sendCalls`/`sendTransaction` is itself an external side-effect boundary: the wallet or connector may accept and submit the request, then disconnect, reject the response, or return a malformed/missing Safe ID. This path throws before `onBroadcast`, so no accepted-submission record exists and an ordinary retry can duplicate an operation whose first outcome is unknowable. The existing unexpected-ID test checks only the error, not replay protection.

Please persist a pre-broadcast armed reservation before invoking the wallet (with no hash yet), transition it to the returned hash/proposal ID when available, and fail closed/manual-recovery when acceptance cannot be disproved. Add EOA and Safe regressions where the wallet boundary accepts but rejects or returns malformed data before an ID reaches `onBroadcast`; after reload, retry must remain blocked.

LeonardEulerXYZ (changes requested):
Late independent lifecycle passes confirmed the three existing replay findings and identified two distinct complements: already-open tabs never refresh another tab's batch quarantine, and wallet submission can become ambiguous before a hash/proposal ID is returned. These require separate regressions from the post-hash reload cases already noted.

utils/pendingSubmissions.ts
```diff
@@ -0,0 +1,333 @@
...
+      if (storage.getItem(key) === value) {
+        memoryFallback.delete(key)
+        return
+      }
+    }
+    catch (err) {
+      logWarn('pendingSubmissions/write', err)
+    }
+  }
+  memoryFallback.set(key, value)
```
LeonardEulerXYZ: **Blocker — abort before the wallet when durable reservation cannot be proven.** A failed/non-sticking `localStorage` write falls back to this process-local map and returns normally, so `onArm` immediately proceeds into `sendTransactionAsync`/`sendCalls`. Reloading or crashing then drops the only record and the next attempt sees a clear gate. The read side has the symmetric problem: a transient getter/`getItem` failure with an existing durable record falls through to an initially empty map. I reproduced the write→reload complement with a fresh module realm. Make reservation persistence report durable success or throw, and stop before invoking the wallet when storage is unavailable, unreadable, corrupt, or fails read-back; tests should assert zero wallet calls and safety across a fresh realm.

utils/pendingSubmissions.ts
```diff
@@ -0,0 +1,333 @@
...
+  const records: PendingSubmissionRecord[] = []
+  for (const key of listRawKeys(pendingSubmissionStorageKeyPrefix(flow))) {
+    const record = readRecordAtKey(key)
+    if (record) records.push(record)
+  }
+  return records
+}
+
+export const writePendingSubmission = (flow: PendingSubmissionFlow, record: PendingSubmissionRecord) => {
+  writeRaw(storageKey(flow, record.owner, record.chainId), JSON.stringify(record))
```
LeonardEulerXYZ: **Blocker — this write is not an atomic reservation acquisition.** The global gate performs an earlier read-only check, then asynchronous preparation occurs before this unconditional single-key overwrite. Two tabs can both observe no record, each arm, and each invoke the wallet; only the last record survives. Worse, either attempt's later `rejected`/`confirmed` callback unconditionally clears the shared key and can delete the other attempt's pending reservation. The existing two-tab test starts tab B only after A has persisted, so it misses this interleaving. Serialize check+reserve per owner/chain (for example with a browser-wide lock) and give records an attempt owner/token so upgrades and terminal deletes are ownership-checked. Add simultaneous empty-read and stale-completion tests requiring exactly one wallet call.

composables/useEulerTx.ts
```diff
@@ -1527,6 +1647,18 @@ export const useEulerTx = () => {
     // must still never take the permit2 path.
     const isKnownSafe = Boolean(safeWalletProvider) || isSafeConnectorIdentity(connector)
 
+    // Cross-surface quarantine: an unresolved value-moving submission from
+    // any flow (batch cart or a direct migration) for this wallet/chain
+    // blocks new sends until it reconciles — switching execution surfaces
+    // must not walk around a flow's own retry gate.
+    await assertNoConflictingPendingSubmission({
```
LeonardEulerXYZ: **Blocker — ordinary direct executors consume the quarantine but do not create one.** This checks records, but `executePlan`'s Safe bundle and sequential sender below have no `onArm`/`onBroadcast` persistence callbacks. Thus an ordinary lend/borrow/swap/multiply/etc. accepted by the wallet followed by transport/status failure leaves no record; a same-surface retry or batch equivalent can submit again. `executePreparedPlan` has the same property whenever callers omit the optional callback, which most ordinary forms do. The executor-level guarantee needs built-in reservation ownership for every value-moving path, not optional caller wiring limited to batch/direct-migration pages. Add callback-less EOA and Safe ambiguous-outcome tests followed by direct and batch retries, asserting a durable record and zero second wallet calls.

composables/useEulerTx.ts
```diff
@@ -1234,18 +1311,28 @@ export const useEulerTx = () => {
...
-        ...(connector ? { connector } : {}),
-        to,
-        data: data as Hex,
-        value: value ?? 0n,
-      })
+      beforeBroadcast?.()
+      onArm?.()
+      let hash: Hash
+      try {
+        hash = await sendTransactionAsync({
```
LeonardEulerXYZ: **Blocker — validate an EOA submission ID before replacing the armed record.** This external wallet result is only TypeScript-cast to `Hash`. If a connector returns `undefined`, `tx-1`, or another malformed value after possible acceptance, `onBroadcast` upgrades the valid armed record to malformed `submitted` state. On the next read, `parseRecord` rejects and deletes that record, reopening retry even though the original request may have landed. The Safe bundle path already validates its ID; apply the same runtime hash check here and leave the record `armed` on malformed output. Add a malformed EOA-ID regression that proves retry makes zero wallet calls.

utils/pendingSubmissions.ts
```diff
@@ -0,0 +1,333 @@
...
+export const resolvePendingSubmissionOutcome = async (
+  record: PendingSubmissionRecord,
+  options: {
+    provider: ReceiptClientLike | undefined
+    getSafeWalletProvider: (connector?: WalletConnectorLike) => Promise<WalletProviderLike | undefined>
+    connector?: WalletConnectorLike
+    safeStatusTimeoutMs?: number
+  },
+): Promise<PendingSubmissionOutcome> => {
+  if (record.phase === 'armed') return 'unknown'
```
LeonardEulerXYZ: **Blocker — an armed record has no post-reload recovery path.** Returning `unknown` is safely conservative, but after reload the in-memory attempt that could emit `rejected`/`confirmed` is gone and there is no ID to query. The record therefore blocks every gated transaction for this wallet/chain forever; the UI's instruction to “let it resolve” cannot change this verdict. This occurs naturally if the page reloads while a wallet prompt is open and the user rejects afterward. Provide an explicit, risk-labelled recovery/reconciliation workflow (or another durable terminal signal) that cannot silently authorize replay, and test reload-before-rejection plus eventual safe recovery.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `34719d5b9ca4870dd1d84440af50fff5bb03af7d`, including every prior Leonard blocker and a fresh anti-anchoring pass. The remediation correctly adds pre-wallet arming, owner/chain-scoped records, attempt-boundary rereads, and cross-surface checks; the earlier prepared-plan, authorization, broadcast-attribution, and connector-pinning fixes remain intact. Five lifecycle complements still leave replay or recovery unsafe, detailed inline.

Executable validation is green: focused lifecycle cohort **261/261**, full Vitest **1,929 passed / 1 skipped**, typecheck, lint (**0 errors; 6 existing warnings**), production build, and `git diff --check`. Two temporary adversarial tests reproduced the failed-durable-write reload loss and the concurrent check→overwrite window; they were removed and the worktree is clean. Green checks do not close those state-machine gaps.

LeonardEulerXYZ (changes requested):
Supplemental exact-head blocker from the independent anti-anchoring lane:

**The cross-surface quarantine is not yet global at the actual wallet/signature boundaries.** `sendPlainTransactions` can invoke `sendTransactionAsync` before the new `executePreparedPlan` gate (`composables/useEulerTx.ts:1349-1395`). Direct migrations send authorization grants before reaching that gate (`pages/position/[number]/migrate.vue:1093-1119`; `pages/position/[number]/borrow/swap.vue:3596-3615`), and batch execution can do the same for standalone grants (`composables/useTxBatch.ts:3131-3133, 3191-3202`). More importantly, direct CoW execution calls `executeCowSwapTransactionPlan` with wallet send/sign callbacks without consulting pending submission records at all (`composables/cowswap/useCowSwapExecutionCore.ts:109-130`). An unresolved batch/migration can therefore coexist with a new economically equivalent CoW order despite the claimed cross-surface gate.

Gate before every wallet/signature side effect, including prerequisites and CoW, and add tests that seed a conflicting flow then assert zero transaction/signature calls from each bypass path.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `776610f54764c44d5df96ac3427d8d99d09101f4` against all six findings from the previous Leonard review and with a fresh transaction-lifecycle complement pass.

Three parts are now solid: durable-or-abort storage/read-back behavior (F1), built-in quarantine for ordinary direct executors (F3), and runtime EOA hash validation (F4). The cross-surface preflight gate from F6 also blocks a *previously existing* conflicting record. This head is still not safe to approve: F2 and F5 are incomplete, and CoW execution introduces an untracked ambiguity window at its own submission boundary.

## 1. Blocker — the no-Web-Locks fallback is not cross-tab atomic (F2)

`withPendingSubmissionLock()` uses `navigator.locks` when available, but otherwise serializes only through module-local `realmLockTails` (`utils/pendingSubmissions.ts:456-483`). Separate tabs/realms have separate maps, so two attempts can both enter the owner/chain critical section, both observe no record, both persist their own reservation, and both invoke the wallet. `attemptId` prevents one completion from deleting another record, but it cannot undo the second wallet request.

I reproduced the missing invariant with a disposable exact-head Vitest probe: two separately loaded module graphs shared storage with `navigator.locks` absent; realm B entered the same owner/chain callback while realm A was still holding it. The probe passed only because it asserted this unsafe overlap. The committed simultaneous-arm test exercises one module graph and therefore only proves the in-realm FIFO path.

Either fail closed when an origin-wide lock cannot be acquired or provide a genuinely cross-context reservation primitive. Add a no-Web-Locks, two-module/two-tab regression that reaches the wallet boundary and requires exactly one wallet call.

## 2. Blocker — direct armed records still have no reachable recovery UI (F5)

`releaseUnverifiablePendingSubmission()` and `releaseArmedAfterManualCheck()` exist, but production calls expose release only for the `'batch'` flow (`useTxBatch.ts:3455-3470` / `BatchContents.vue`). The outgoing and inbound migration quarantines are created in their pages and can leave durable armed records, but neither page calls `releaseArmedAfterManualCheck`; its only callers are tests.

The ordinary failure remains: reload while a direct wallet prompt is open, then reject after the old page is gone. The durable armed record has no ID to reconcile and every cross-surface gate for that owner/chain remains blocked. The batch dismissal cannot clear an `outgoing-migration` or `inbound-migration` key.

Expose the same explicit, risk-labelled acknowledgement on every UI surface that can create an armed record, and test the full reload-before-rejection path through each production page/composable rather than only the helper.

## 3. Blocker — CoW order submission checks existing quarantine but never creates its own

`useCowSwapExecutionCore.executePlan()` checks for a prior conflicting record at lines 123-137, then passes raw wallet send/sign callbacks into `executeCowSwapTransactionPlan` at lines 138-145. It never begins, arms, upgrades, or seals a pending-submission attempt for the CoW operation itself. On error, lines 160-165 simply reset the local status to `idle`.

The installed SDK signs the order, emits `submitOrder`, then awaits `submitCowSwapOrder`. If the orderbook accepts the request but the response is lost, the app has neither an order UID nor a quarantine record. A retry can sign and submit another live order.

A disposable exact-head probe exercised that sequence: the SDK mock crossed `signTypedData`, emitted `submitOrder`, then threw an accepted-but-response-lost error. `executePlan()` rejected, while both `direct` and `batch` pending records remained absent. The committed CoW regression only proves that an already seeded batch record blocks entry; it does not test CoW's own ambiguous result.

Give CoW execution an owned durable attempt: arm before the first approval/signature/submission boundary that can create irreversible state, retain an armed record across ambiguous orderbook outcomes, upgrade when an order UID is known, and provide objective/manual recovery appropriate to signed CoW orders.

## Validation

- full Vitest: **187 passed / 1 skipped files; 1,967 passed / 1 skipped tests**
- `npm run typecheck`: passed
- ESLint on all 13 changed remediation source/test files: passed
- production `npm run build`: passed (existing build warnings only)
- `git diff --check`: passed
- hosted lint, typecheck, test, CodeRabbit, Cursor security, and Railway preview checks: green
- two disposable adversarial regressions reproduced the cross-realm fallback overlap and CoW accepted-response-loss gap; both probes were removed after execution

Green aggregate checks are supporting evidence, but the state-machine complements above still permit either duplicate wallet/order submissions or an unrecoverable direct-flow lockout. Requesting changes at this exact head.

LeonardEulerXYZ (commented):
Supplement to my exact-head `776610f54764c44d5df96ac3427d8d99d09101f4` changes-requested review after the three independent complement lanes completed. The existing verdict is unchanged. Four additional lifecycle blockers survived source verification.

## 1. A failed durable post-hash upgrade becomes manually releasable as “no ID returned”

`upgradePendingSubmissionToSubmitted()` catches a failed durable `armed → submitted` write and returns `undefined` (`utils/pendingSubmissions.ts:412-455`). Batch tracking correctly knows the wallet returned a hash (`useTxBatch.ts:3122-3159`), but durable storage remains the old `armed` record. After reload, `releaseUnverifiablePendingSubmission()` classifies that record as hashless and permits deletion after the user checks that the wallet shows nothing pending (`pendingSubmissions.ts:538-559`; batch CTA at `useTxBatch.ts:3435-3466`).

Reachable sequence: the initial armed write succeeds; the wallet returns a valid hash; the submitted-state write fails; the transaction mines; the page reloads; the wallet no longer shows it as pending; the user follows the CTA and releases the only replay guard. The batch can then execute again even though the first transaction landed.

A reviewer probe composed the failed upgrade through manual recovery. Expected release refusal because a hash had been returned; actual result was `true`. Preserve the known hash in a fail-closed durable form or make failed post-hash persistence unrecoverable through the armed-record path. Add the full upgrade-failure → reload → manual-release → retry regression.

## 2. Batch hydration transfers ceremony context across different attempt IDs

`hydratePendingBatchSubmissionFromStorage()` preserves the prior ceremony whenever the prior record was armed, without comparing `attemptId` (`composables/useTxBatch.ts:386-395`). This permits:

1. tab A owns armed attempt A with ceremony A;
2. A terminally deletes A;
3. tab B reserves the same owner/chain as attempt B before A's post-delete hydration;
4. A hydrates B's record but retains ceremony A because `prior.record.phase === 'armed'`;
5. B upgrades/submits/lands, and A's landed handler retires entries using ceremony A rather than treating B as an unknown cross-tab attempt.

That can preserve equivalent rebuilt entries in A after B already executed them, reopening duplicate execution. Context continuity must require the same `attemptId` (and expected phase transition), not merely “the previous mirror was armed.” Add the release-A → reserve-B → hydrate-A → land-B interleaving.

## 3. CoW wallet callbacks are not pinned to the reviewed account, chain, or connector

`useCowSwapExecutionCore` captures an address for SDK arguments, but its actual callbacks call `sendTransactionAsync({ to, data, value })` and `signTypedDataAsync(typedData)` without the reviewed account, chain, or connector (`composables/cowswap/useCowSwapExecutionCore.ts:81-94, 111-145`). Wallet drift can land during `getEulerSdkFresh()`, pending-record reconciliation, approval waits, or other SDK stages.

An approval can therefore be sent from the newly active account/on the newly active chain, and an order signature can come from a connector or signer different from the reviewed owner. Capture the reviewed connector identity at review/confirmation entry, pass account/chain/connector explicitly to wagmi, and revalidate immediately before every approval and signature callback. Tests should switch account, chain, and same-address connector independently during SDK awaits and assert zero wallet calls.

## 4. Generic prepared-plan reviews accept same-address connector drift and can change signature mode

Ordinary swap, collateral, and multiply reviews pass an immutable `TransactionPlanPrepared` into the modal, then later call `executePreparedPlan(prepared)` without any reviewed connector key (`useSwapPageLogic.ts:556-590`, `position/useCollateralForm.ts:1055-1092`, `multiply.vue:989-1019`). `executePreparedPlan()` discovers the *current* connector only after SDK/provider awaits (`useEulerTx.ts:1956-1972`). A same-owner/same-chain switch is invisible.

If the replacement connector is classified as Safe, lines 1996-2015 re-resolve the reviewed Permit2-signature plan into transaction approvals; otherwise signatures are still dispatched through an unreviewed wallet session. Thus the wallet ceremony can change from reviewed signature prompts to approval transactions, or simply move to another connector, without re-review.

Bind each prepared review to owner, chain, connector, and wallet classification, then fail closed at every irreversible boundary. Do not repair a reviewed EOA signature ceremony into a Safe approval ceremony after confirmation.

The independent lanes also reconfirmed the three blockers in my main review. Focused cohorts passed (250-268 tests depending on lane), one lane ran the full suite at 1,972 passing tests, typecheck and diff checks passed, and all reviewer probes were removed with the worktree clean. These findings are state-machine complements not exercised by the green aggregate suite.

PR #814, opened 2026-08-17
feat: make deployment manifest and ABI fetching resilient to upstream outages

kasperpawlowski:
## Context

During the 2026-08-17 GitHub incident (raw content downloads erroring at ~50%+), app.euler.finance stopped loading vaults. Root cause chain: the `/api/internal/euler-chains` proxy caches the euler-interfaces `EulerChains.json` manifest with a 5-min TTL, but its stale-on-error fallback was bounded to 10 minutes past TTL by the cache utility's default staleness ceiling. Once the outage outlasted that window, the proxy started returning 502, every `buildEulerSDK` call failed at `DeploymentService.build`, and no addresses meant no vaults. Two adjacent gaps had the same root: runtime ABIs were fetched browser-direct from raw.githubusercontent.com with no proxy at all, and server-side SDK builds (vault snapshot builder, token-list, SSR) fetched the manifest directly from GitHub, bypassing the proxy cache entirely.

## Changes

**Cache policy (`server/utils/cache.ts`)**
An explicitly passed `maxStaleMs` is now honored as-is, above the 30-minute ceiling included. The ceiling still applies to every cache that doesn't set one (unchanged behavior for all existing callers — nothing passed `maxStaleMs` before). Low-churn manifests are strictly better served days-stale than not at all; the capped default remains right for market/account data.

**Deployments manifest (`/api/internal/euler-chains`)**
- 7-day stale window (`MANIFEST_MAX_STALE_MS`) — a running instance now outlives any realistic upstream outage; the warm-cache loop keeps the entry current every 5 minutes.
- Precedence fix: an explicit `NUXT_PUBLIC_CONFIG_EULER_CHAINS_URL` now wins over the branch env vars (previously the reverse), matching the labels/oracle-checks override convention and making it usable as the emergency repoint lever during an upstream outage.
- No baked-in data: an instance cold-started mid-outage 502s until upstream recovers or the env repoints it at a mirror — a deliberate, documented failure mode.

**Runtime ABIs (new `/api/internal/abis/[contract]`)**
- Same-origin proxy with the identical cache + 7-day stale chain for the three ABIs fetched at runtime (AccountLens — SDK account reads/simulate/rewards; VaultLens/UtilsLens — projected rates and IRM overview).
- `NUXT_PUBLIC_CONFIG_EULER_ABIS_BASE_URL` as the matching mirror override (explicit URL wins over branch vars, consistent with labels, oracle-checks, and euler-chains).
- Browser SDK wired via `abiService.setQueryABI` in `configureAppProxies`; a missing setter logs an error instead of silently no-oping.
- `raw.githubusercontent.com` removed from CSP `connect-src` — the ABI fetches were the last browser-direct GitHub dependency.
- Contract allowlist returns 404 for anything unknown; CDN route rule added.

**Server-side SDK builds (new `server/plugins/sdk-deployments.ts`)**
`DeploymentService.setQueryDeployments` is installed at boot so every server-side `buildEulerSDK` (vault snapshot builder, token-list, SSR instances) resolves deployments through the shared euler-chains cache instead of fetching GitHub directly — one cached copy with the long stale window instead of N independent GitHub fetches.

**Warm cache**
The global 5-min cycle now also refreshes the three ABI manifests.

**Docs**
architecture.md, development-guide.md, sdk-integration.md, and server-side-caching.md updated for the new endpoint, the stale-window policy, the env precedence, and the server-side deployments wiring.

## Test plan

- [x] `vitest run` — full suite green (1769 passed), including new coverage: cache staleness policy (default window, ceiling clamp, explicit override above/below ceiling), euler-chains precedence + resolution chain (throws with no cache, stale served through a prolonged outage up to the window, fresh-cache short-circuit), abis route (branch resolution, base-URL override precedence, non-array rejection, allowlist 404, handler happy path), and browser SDK `setQueryABI` wiring with URL→contract mapping
- [x] `nuxt typecheck` clean, `eslint` clean (0 errors)
- [x] `nuxt build` succeeds
- [ ] Verify on a preview deployment that vault lists keep loading when raw.githubusercontent.com is blocked after first warm (e.g. DNS null-route) and that a cold boot with it blocked 502s cleanly on `/api/internal/euler-chains`
- [ ] Confirm production env doesn't rely on the old branch-over-URL precedence (no `NUXT_PUBLIC_CONFIG_EULER_CHAINS_URL` set alongside a branch var with a different intent)

## Summary by CodeRabbit

- **New Features**
  - Added same-origin runtime ABI access for supported contracts.
  - Added configurable branch and endpoint overrides for chain manifests and ABI documents.
  - Routed SDK ABI requests through the local proxy.
  - Added automatic cache warming for chain and ABI data.

- **Bug Fixes**
  - Improved validation and outage resilience with stale-cache fallback.
  - Improved deployment resolution through cached chain data.

- **Documentation**
  - Clarified configuration, caching, endpoint selection, and outage behavior.

- **Security**
  - Removed an unnecessary external source from the content security policy.


LeonardEulerXYZ:
## Leonard review — manifest outage resilience

**Verdict: changes requested** on `1b2cd19aa372ee04497932ce13411049cfad3a4b`.

### Walkthrough

This PR correctly centralizes the deployment manifest and three runtime ABIs behind same-origin proxies, installs the server-side deployment query hook, removes the final browser-direct GitHub dependency from CSP, and gives explicitly selected low-churn caches a seven-day stale horizon. The cold-start limitation is stated accurately: a new process still needs either one successful upstream load or an emergency mirror.

### Blocking finding

Both refresh paths admit any JSON array as a successful value. A `200 []` or array-shaped malformed mirror response therefore replaces the last-known-good manifest/ABI and refreshes its timestamp. Subsequent failures serve that unusable value for the long stale window, defeating the intended resilience guarantee.

The two inline comments identify the duplicated cache-admission points. The fix should validate a usable payload before `cache.set(...)`; validation failure should throw so `loadEulerChains()` / `loadAbi()` can return the existing stale value. Please pin this with success → `200 []` → last-known-good regression tests for both paths.

### Validation

- Focused PR tests: 27/27 passed
- Full Vitest suite: 1,769/1,769 passed
- `npm run typecheck`: passed
- `npm run lint`: 0 errors; six unrelated existing warnings
- `npm run build`: passed
- Built-runtime mirror smoke: all four endpoints returned 200; after stopping the mirror, all four continued returning cached 200 responses
- Adversarial local regression: 2/2 failed as expected, proving valid cached data is replaced by `[]`; the temporary test was removed and the worktree is clean
- Live Railway preview: the manifest and all three ABI endpoints currently return 502 because the upstream incident prevents the first seed; this matches the documented cold-start boundary and was not treated as a separate defect

No malicious-code, credential, dependency, or supply-chain changes found in the diff.

server/api/internal/euler-chains.get.ts
```diff
@@ -3,36 +3,35 @@ import { createRateLimiter } from '~/server/utils/rate-limit'
...
 const rateLimiter = createRateLimiter({
   max: 1000,
   windowMs: 60_000,
   label: 'euler-chains',
 })
 
-const cache = createTtlCache<unknown[]>({ ttlMs: CACHE_TTL_MS })
+const cache = createTtlCache<unknown[]>({
+  ttlMs: CACHE_TTL_MS,
+  maxStaleMs: MANIFEST_MAX_STALE_MS,
```
LeonardEulerXYZ: **Blocking — preserve last-known-good on semantic corruption.** This seven-day stale policy assumes every admitted value is usable, but `refreshEulerChains()` still accepts any array. After TTL, a `200 []` or `[{}]` response replaces the valid entry and refreshes its timestamp; later upstream failures then serve poison for up to seven days. Validate a non-empty, structurally usable EulerChains manifest before `cache.set(...)`, and throw on validation failure so `loadEulerChains()` falls back to the existing stale value. Please add a valid seed → `200 []` → valid stale result regression test.

server/api/internal/abis/[contract].get.ts
```diff
@@ -0,0 +1,119 @@
...
+/** Forced upstream refresh used by the warm-cache plugin; see euler-chains. */
+export function refreshAbi(contract: AbiContract): Promise<unknown[]> {
+  return inFlight.run(contract, async () => {
+    const resp = await fetchWithTimeout(getUpstreamUrl(contract))
+    if (!resp.ok) {
+      throw new Error(`Upstream returned ${resp.status}`)
+    }
+
+    const data: unknown = await resp.json()
+    if (!Array.isArray(data)) {
```
Comment (LeonardEulerXYZ) [inline server/api/internal/abis/[contract].get.ts]: **Blocking — do not cache an array-shaped invalid ABI.** `Array.isArray` admits `[]` and malformed ABI arrays, so the five-minute warm refresh can overwrite a working ABI and turn the seven-day fallback into a long-lived bad cache entry. I reproduced `valid ABI → advance past TTL → 200 [] → [] returned` in a local throwaway Vitest. Validate a non-empty usable ABI before `cache.set(...)`; on validation failure, throw and serve the previous stale value. Add the corresponding regression test alongside the deployment-manifest case.

LeonardEulerXYZ (changes requested):
Reviewed exact head `1b2cd19aa372ee04497932ce13411049cfad3a4b`.

**Changes requested:** the long stale window works for transport failures after a valid seed, but both new resolution chains can overwrite the last-known-good value with an array-shaped invalid `200` response. See the inline comments.

Validation:
- focused tests: 27 passed
- full suite: 1,769 passed
- typecheck passed
- lint passed with 0 errors (six unrelated existing warnings)
- production build passed
- built-runtime mirror/cache smoke passed
- local throwaway cache-admission regression failed 2/2, reproducing the blocker; removed afterward

The current Railway preview's 502s are consistent with the explicitly documented cold-start limitation during the active upstream incident, not an additional finding.

server/api/internal/euler-chains.get.ts
kasperpawlowski: Fixed in dfecb4f7. refreshEulerChains() now admission-checks the payload before cache.set: non-empty array, every entry an object with a positive integer chainId and an object addresses. Unusable 200s ([], [{}], non-object addresses) throw, so loadEulerChains() falls back to the existing stale entry and its timestamp is never reset by a poison response. Regression added: valid seed → past TTL → 200 [] → stale value served, plus a second poison response to assert the timestamp survives.

server/api/internal/abis/[contract].get.ts
Comment (kasperpawlowski) [reply server/api/internal/abis/[contract].get.ts]: Fixed in dfecb4f7. refreshAbi() now requires a non-empty array whose entries are objects with a string type (the invariant every viem ABI item satisfies) before cache.set; anything else throws and loadAbi() serves the previous stale entry. Regressions added: parameterized rejection cases ([], missing type, non-array) and valid seed → past TTL → 200 [] → stale ABI served.

server/api/internal/euler-chains.get.ts
LeonardEulerXYZ: Re-tested on `dfecb4f7`: the empty-array path is fixed, but `[{ chainId: 1, addresses: {} }]` still passes admission and replaces the cache. The installed SDK requires `addresses.coreAddrs` and `addresses.lensAddrs`; a direct consumer then fails reading `coreAddrs.evc`. Please extend this guard to the SDK's minimum usable `Deployment` shape and add the same stale-preservation regression for this non-empty malformed payload.

server/api/internal/abis/[contract].get.ts
Comment (LeonardEulerXYZ) [reply server/api/internal/abis/[contract].get.ts]: Re-tested on `dfecb4f7`: even an ABI containing the expected `getEVCAccountInfo` and `getVaultAccountInfo` names still passes when the function definitions omit inputs, outputs, and mutability. The SDK accepts that runtime ABI instead of using its bundled fallback; viem then encodes `getEVCAccountInfo` as `0x73d2c828` rather than canonical `0xf859d53c…` calldata with two addresses. Please validate full ABI item shape plus the required signatures for each allowlisted contract, and cover this malformed non-empty 200 in the stale-preservation regression.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `dfecb4f71ce99f5daec3b04406f88c1b86344202`.

The new checks and regressions correctly reject `[]` and preserve the stale timestamp, but the cache-admission invariant remains incomplete. Both `[{ chainId: 1, addresses: {} }]` and `[{ type: 'function' }]` pass the new guards and overwrite the last-known-good entry, although the installed SDK immediately requires nested core/lens addresses and a real ABI function definition. I reproduced both as failing rejection tests on this head.

Please validate the minimum SDK `Deployment` shape and usable contract-specific ABI shape before `cache.set(...)`, then extend the stale-fallback regressions to malformed non-empty payloads. Full tests, focused suites, typecheck, lint, build, diff check, and current CI otherwise pass.

server/api/internal/euler-chains.get.ts
kasperpawlowski: Fixed in c56950d1 — you were right that the guard was narrowed, not closed. Admission now requires, per entry: positive integer chainId, plus EVM-address strings (0x + 40 hex) for coreAddrs.{eVaultFactory,evc,permit2} and lensAddrs.{accountLens,oracleLens,utilsLens,vaultLens} — the keys SDK builds and the core surfaces dereference unconditionally. Peripheral keys (tokenAddrs, peripheryAddrs, remaining lens/core entries) stay unpinned deliberately, so a sparse future entry degrades one feature instead of rejecting the whole manifest. Verified all 15 entries of the current euler-interfaces master manifest pass with zero rejections. Your exact repro ([{ chainId: 1, addresses: {} }]) is now a rejection case and part of the stale-preservation regression: valid seed → past TTL → that payload as 200 → stale manifest served, timestamp untouched.

server/api/internal/abis/[contract].get.ts
Comment (kasperpawlowski) [reply server/api/internal/abis/[contract].get.ts]: Fixed in c56950d1. Two layers now: (1) every function fragment must carry name, inputs array, outputs array, and stateMutability string; (2) per allowlisted contract, the ABI must contain the canonical signatures its consumers encode against — getEVCAccountInfo(address,address) + getVaultAccountInfo(address,address) for AccountLens, getVaultInterestRateModelInfo(address,uint256[],uint256[]) for VaultLens, computeAPYs(uint256,uint256,uint256,uint256) for UtilsLens — computed with viem's toFunctionSignature, so it's exactly the name+inputs invariant selector encoding depends on. Outputs are deliberately unpinned since return-tuple drift is the reason ABIs resolve at runtime. Your repro (right names, fragments omitting inputs/outputs/mutability) is a rejection case and part of the stale-preservation regression; a wrong-signature case (getVaultInterestRateModelInfo(address)) and a cross-contract case (VaultLens ABI offered to AccountLens) are covered too. Verified all three current euler-interfaces master ABIs pass with zero rejected items.

server/api/internal/euler-chains.get.ts
```diff
@@ -3,38 +3,52 @@ import { createRateLimiter } from '~/server/utils/rate-limit'
...
+// usable manifest may overwrite the last-known-good entry. An array-shaped
+// but unusable 200 (`[]`, entries missing chainId/addresses) must throw so
+// loadEulerChains() keeps serving the previous stale value instead of
+// preserving poison for up to MANIFEST_MAX_STALE_MS.
+const isValidDeployment = (entry: unknown): boolean => {
+  if (entry === null || typeof entry !== 'object') return false
+  const { chainId, addresses } = entry as { chainId?: unknown, addresses?: unknown }
+  return Number.isInteger(chainId) && (chainId as number) > 0
+    && addresses !== null && typeof addresses === 'object'
+}
```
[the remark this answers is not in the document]
kasperpawlowski: Addressed in c56950d1: addresses, coreAddrs, and lensAddrs are now all checked with an isRecord helper that excludes arrays, and an 'array addresses' rejection test case was added.

server/api/internal/abis/[contract].get.ts
Comment (kasperpawlowski) [reply server/api/internal/abis/[contract].get.ts]: Fixed in ba972b36. The drift-tolerance line was drawn too loosely: a changed return tuple is fine, an absent one is not. Required signatures must now resolve to a fragment whose outputs tuple is non-empty, and every input/output entry must be an object with a string type. The outputs *shape* stays unpinned so return-tuple drift keeps working. Regressions added: required-function-with-empty-outputs and malformed output entries as rejection cases, plus a stale-preservation step feeding the valid AccountLens ABI with all outputs stripped — it must not displace the last-known-good entry. Re-verified all three euler-interfaces master ABIs pass with zero rejected items.

server/api/internal/euler-chains.get.ts
kasperpawlowski: Follow-up in 8be22300: the all-or-nothing data.every(...) admission had a failure mode of its own — a routine euler-interfaces commit adding a sparse new chain (e.g. no permit2 yet) would freeze the manifest and turn into a total 502 on both realms once the 7-day stale window expired. Admission is now per entry: invalid entries for non-enabled chains are dropped with a warning listing their chainIds, and the remaining valid entries are cached; every enabled chain must still be present and valid, otherwise the payload rejects and the last-known-good copy is preserved. Regressions: sparse-non-enabled-entry admits the rest, invalid-enabled-entry rejects, missing-enabled-chain rejects.

server/api/internal/euler-chains.get.ts
```diff
@@ -3,36 +3,115 @@ import { createRateLimiter } from '~/server/utils/rate-limit'
...
+// adding a sparse new chain must not freeze the manifest and time-bomb a
+// total 502 when the stale window expires), while every enabled chain must
+// be present and valid — an unusable payload for a chain users are actually
+// on is exactly the poison the stale window exists to outlive.
+//
+// Required keys are the ones whose absence breaks SDK builds or the core
+// lend/borrow surfaces on every chain (all current manifest entries carry
+// the full key set; peripheral keys are deliberately not required so a
+// sparse entry degrades a feature, not the chain).
+const REQUIRED_CORE_ADDRS = ['eVaultFactory', 'evc', 'permit2'] as const
```
LeonardEulerXYZ: **Blocking — make enabled-chain admission match the SDK deployment contract.** The installed SDK `Deployment` type requires eight core addresses (`balanceTracker`, `eVaultFactory`, `eVaultImplementation`, `eulerEarnFactory`, `evc`, `permit2`, `protocolConfig`, `sequenceRegistry`) and six lens addresses (`accountLens`, `eulerEarnVaultLens`, `irmLens`, `oracleLens`, `utilsLens`, `vaultLens`); `useEulerAddresses` dereferences that same set. This predicate checks only 3 + 4, and the committed `validManifest()` fixture therefore passes despite omitting seven SDK-required addresses. The address predicate also accepts `0x000…000`, so a structurally complete enabled entry can still route core calls to no contract. Either case overwrites last-known-good for the seven-day window. Please require every non-optional SDK field used by enabled chains, reject zero for required contracts, and cover valid seed → incomplete/zero enabled entry → stale value preserved. Non-enabled sparse entries can still be filtered as intended.

server/api/internal/euler-chains.get.ts
```diff
@@ -3,36 +3,115 @@ import { createRateLimiter } from '~/server/utils/rate-limit'
...
+  }
+
+  const valid: unknown[] = []
+  const dropped: (number | 'unknown')[] = []
+  for (const entry of data) {
+    if (isValidDeployment(entry)) valid.push(entry)
+    else dropped.push(readChainId(entry) ?? 'unknown')
+  }
+
+  const validChainIds = new Set(valid.map(entry => readChainId(entry)))
```
LeonardEulerXYZ: **Blocking — reject duplicate chain IDs before cache admission.** Two individually valid entries for the same enabled `chainId` both survive this filter, and the `Set` completeness check treats that as healthy. The installed SDK then reduces deployments into an object with `acc[deployment.chainId] = deployment`, so the later entry silently wins; reversing upstream order changes every address used for that chain. With this proxy, that ambiguous choice can become the seven-day last-known-good value. Reject duplicate IDs for the admitted manifest (rather than relying on array order) and add an adversarial regression.

server/api/internal/abis/[contract].get.ts
```diff
@@ -0,0 +1,191 @@
...
+
+const isValidAbiItem = (item: unknown): boolean => {
+  if (item === null || typeof item !== 'object') return false
+  const { type, name, inputs, outputs, stateMutability } = item as Record<string, unknown>
+  if (typeof type !== 'string') return false
+  if (type !== 'function') return true
+
+  return typeof name === 'string'
+    && Array.isArray(inputs) && inputs.every(isAbiParameter)
+    && Array.isArray(outputs) && outputs.every(isAbiParameter)
```
Comment (LeonardEulerXYZ) [inline server/api/internal/abis/[contract].get.ts]: **Blocking — validate consumer-compatible outputs, not only non-empty output entries.** `isAbiParameter` accepts any string as a type, while required-signature validation covers inputs only. More importantly, the committed “valid” fixtures give every required function `outputs: [{ type: "bytes" }]`. That payload is admitted and cached, but real `computeAPYs` returndata encoded as `(uint256,uint256)` throws viem `PositionOutOfBoundsError` when decoded through the accepted `bytes` output ABI; an arbitrary type string similarly produces `InvalidAbiDecodingType`. AccountLens and VaultLens consumers also read named tuple fields, so “some output exists” is not a usability invariant. Validate ABI parameter types recursively and require the minimum output tuple/field types each runtime consumer needs (allowing compatible extra fields if desired), then add a valid seed → incompatible non-empty output → stale value preserved regression.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `8be22300bcd5655837adece864d899b16b2ccb86`.

The reported whole-manifest rejection is fixed: the current sparse non-enabled-chain regression passes, and the same test overlaid on `ba972b3` fails as expected. The enabled-chain completeness gate also runs before `cache.set`, so that original delayed-total-outage finding is resolved.

Three remaining cache-admission blockers are called out inline: the enabled-chain deployment predicate is incomplete and admits zero required addresses; duplicate chain IDs remain ambiguous and become SDK last-wins; and required ABI functions can still carry decoder-incompatible output schemas that overwrite last-known-good.

Verification on this exact head:
- `npm run test:run`: 1,786 passed, 1 skipped
- `npm run typecheck`: passed
- `npm run build`: passed
- `npm run lint`: passed with 6 pre-existing warnings, 0 errors
- Railway preview: euler-chains plus AccountLens/UtilsLens/VaultLens endpoints returned HTTP 200; all six configured chain IDs were present
- GitHub lint/typecheck/test/security/Railway checks: green

No UI/browser smoke requested: the latest change is confined to server-side manifest admission.

LeonardEulerXYZ (approved):
Superseding my earlier changes-requested review after severity recalibration. The actual merge blocker — one sparse non-enabled entry causing whole-manifest rejection and a delayed total outage — is fixed and covered by a regression that fails on the prior implementation. The remaining inline points are defensive hardening against malformed data from the controlled interfaces source, not demonstrated merge blockers. Exact-head tests, typecheck, lint, build, preview endpoints, and hosted checks are green.

Seranged (approved):

PR #822, opened 2026-08-20
Centralize reviewed transaction execution

dglowinski:
## Summary

- Replace page-owned signing and submission callbacks with one durable reviewed-execution path across direct, batch, Safe, migration, and reward flows.
- Bind the existing handcrafted review UI to deterministic reviewed requests while preserving eager form preprocessing and cache reuse.

## Changes

- Center the design on three first-class domain concepts: `OperationIntent`, `ReviewedExecution`, and `SubmissionAttempt`.
- Add deterministic materialization and digests, internal policy/simulation/effect data, an IndexedDB journal, coordinator, recovery UI, and EOA and Safe adapters.
- Migrate every inventoried in-scope form and remove the legacy execution lifecycle entry points.
- Preserve the exact operation and batch review presentation and keep Pyth details invisible.
- Keep the manifest and lockfile pinned to SDK `2.0.1`; development and validation use the locally linked companion SDK from euler-xyz/euler-sdks#96.
- Add SDK release conformance, inventory, model, failure-injection, review-binding, and preview-cache coverage.

## SDK prerequisite

This PR intentionally remains **do not merge** until the SDK materialization API in euler-xyz/euler-sdks#96 is published. Clean installs against SDK `2.0.1` do not yet contain those exports; local integration validation uses the linked SDK branch without changing Lite package files.

## Test plan

- [x] `npm run build` with the companion SDK branch linked locally
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run test:run`
- [x] `npm run test:golden`
- [x] `npm run execution:record -- --dry-run`
- [x] Live lending smoke test through eager preview and review enablement
- [ ] Re-run clean-install CI after the companion SDK is published
- [ ] Run parity, fork execution, production-browser performance, and real multi-tab release gates before merge


LeonardEulerXYZ (changes requested):
Reviewed exact head `a841d6d0effa991c48f77d8a08f0153022730981` against `development`, including the companion SDK PR #96 head `81adcca0a7d7c646d6fd4f14a09e467437c58bfd`.

Requesting changes for two execution-boundary blockers:

1. Confirmed EOA migration grants are left active when a later core request fails or is rejected; the already-reviewed cleanup is never attempted.
2. The JIT EOA Pyth path can dispatch after wallet context changes during the refresh await because both post-refresh checks are skipped.

There is also a post-success regression for batched external migrations: their delayed external-position refresh flag is now dead data.

Validation with SDK #96 linked locally: focused reviewed-execution suite (14 files / 130 tests), typecheck, lint, full Vitest, golden fixtures, execution-record dry-run, and production build all passed. A deferred Pyth context-drift discriminator failed in the unsafe direction: dispatch still occurred after the binding changed during `refreshPyth`. Worktree remained clean.

Separately, the PR is currently conflicting with `development` and remains intentionally dependent on an unpublished SDK version, so it is not merge-ready even apart from these findings.

features/reviewed-execution/coordinator/coordinator.ts
```diff
@@ -0,0 +1,511 @@
...
+        signatures.push({ slotId: slot.slotId, signature })
+      }
+
+      adapter = dependencies.adapters[transport]
+      callbacks = {
+        assertWalletBinding: assertWallet,
+        beforeDispatch: async (stepIndex) => {
+          if (compensating) {
+            compensationStepIndex = stepIndex
+            await assertCleanupPolicyAndWallet()
```
LeonardEulerXYZ: **[P1] Keep compensating revocation reachable when volatile policy fails**

Once a migration grant is confirmed and the core request conclusively aborts, the reviewed revoke is the safety action. This branch sends it through `assertCleanupPolicyAndWallet()`, which re-runs the entire mutable policy; a ToS registry gap, screening timeout/denial, labels failure, or any other policy error therefore suppresses cleanup and leaves the grant active.

I reproduced this with a focused adapter discriminator: confirm `prerequisite`, reject `core`, then fail `revalidatePolicy` only during compensation. Only `['prerequisite', 'core']` reached dispatch, and the result reported `revocation: failed` with `authorizationMayRemain: true`. Keep the exact wallet-binding and immutable reviewed-cleanup checks, but do not let policy intended to deny new economic actions make an already-reviewed authority-reducing revoke unreachable. Please add that failure-path regression.

LeonardEulerXYZ (changes requested):
Re-reviewed exact head `78f215b2fdd8f344f483e8c6c06a241bbf9fee35` against current `development`, including companion SDK PR #96 at `81adcca0a7d7c646d6fd4f14a09e467437c58bfd`.

The four findings from my previous review are materially addressed: known-abort migration cleanup is now attempted, wallet/policy are revalidated after JIT Pyth refresh, detached Safe success completes authoritative state, and batched external migrations retain their refresh callback. Their focused regressions pass. The branch also merges cleanly with current `development`.

Requesting changes on a bounded remaining set:

1. The new compensating revoke still re-runs the full volatile policy and can be suppressed after its grant is already confirmed; see the inline discriminator below.
2. ToS evidence remains fail-open when the UI blocker is not mounted: https://github.com/euler-xyz/euler-lite/pull/822#discussion_r3842847255
3. Swap quote destination/account fields remain admitted rather than rebound to the reviewed owner/subaccount/form destination: https://github.com/euler-xyz/euler-lite/pull/822#discussion_r3842455832
4. Permit2 typed data is still not bound to the reviewed approval amount/expiration: https://github.com/euler-xyz/euler-lite/pull/822#discussion_r3842455839

There is also a release-sequencing gate: Lite still pins published SDK `2.0.1`, while this head imports APIs only present on open SDK PR #96. Hosted typecheck/tests fail for that reason.

Validation with SDK #96 linked locally under Node 24: focused reviewed-execution regressions (103/103), typecheck, lint, full Vitest (1,902/1,902), golden fixtures (6/6), execution-record dry run, and production build all passed. Build output was 10.2 MB total / 2.6 MB gzip. Worktree remained clean.

features/reviewed-execution/materialization/prepared-plan.ts
```diff
@@ -0,0 +1,635 @@
...
+    pendingRequests: pendingRequests.slice(planRequestStart),
+    liteSignatureSlots: signatureSlots,
+  })
+  for (const slot of migrationSignatureSlots) {
+    if (!signatureSlots.some(candidate => candidate.kind === 'migration' && candidate.typedDataHash === slot.typedDataHash && candidate.signer === slot.signer)) {
+      throw new Error('Migration signature has no reviewed EVC insertion point')
+    }
+  }
+
+  for (const [index, input] of [...after].reverse().entries()) {
```
LeonardEulerXYZ: **[P1] Unwind nested migration authorizations in reverse grant order**

The migration compiler already builds `after` in reverse order with `unshift`: two nested pairs produce grants `[G1, G2]` and cleanup `[R2, R1]`. Reversing `after` again here restores grant order, so the executable vector becomes `G1, G2, core, R1, R2` instead of `G1, G2, core, R2, R1`. If revoking the outer authorization first removes the authority needed to revoke the inner one, the inner grant remains active.

The current tests split the invariant: the compiler test asserts `[R2,R1]`, while the materialization test supplies forward-ordered cleanup and asserts this reverse. A composed probe reproduced `G1,G2,core,R1,R2`. Please make one layer own ordering and add an integration regression with two distinct authorization pairs asserting the complete executable order.

features/reviewed-execution/coordinator/coordinator.ts
```diff
@@ -0,0 +1,511 @@
...
+            return
+          }
+          activeStepIndex = stepIndex
+          await assertPolicyAndWallet()
```
LeonardEulerXYZ: **[P1] Keep normal post-core revocation reachable after the operation deadline**

A cleanup request reached through the original EOA dispatch still has `compensating === false`, so it uses `assertPolicyAndWallet()` and re-applies the core operation/signature deadlines. If the core confirms just after that deadline, the already-reviewed revoke is rejected locally and never reaches the wallet. Because the active phase is then `cleanup`, the catch path does not attempt compensation either.

I reproduced this with deadline `101`: grant and core dispatched at time `100`, the core receipt advanced time to `102`, and only `['prerequisite', 'core']` reached dispatch; the result reported submitted core, failed revocation, and `authorizationMayRemain: true`. Select cleanup-safe validation by request phase during the original dispatch and add the deadline-crossing regression.

LeonardEulerXYZ (changes requested):
Challenge-pass amendment to the exact-head review: two additional cleanup-order/liveness defects were independently reproduced. These are sibling cases of the migration-authorization invariant, not a restart of the broad review. The remaining review is now bounded to the previously listed stop set plus these two exact regressions.

LeonardEulerXYZ (approved):
Approved exact head `6c81aee1084d4704c6aca27ae43b8698c3566b76`.

Reviewed latest commit `6c81aee1` separately: it cleanly surfaces ToS and unverified-vault prerequisites for both outgoing migration actions without changing transaction construction. Focused validation passed: inventory 15/15, ToS guard 10/10, targeted ESLint, and `git diff --check`.

Per maintainer clarification, nested authorization ordering is not a required invariant for the supported Aave/Morpho migrations, and post-core authorization cleanup is best-effort. I am retracting those cleanup findings as blockers.

The package-publication/CI sequencing remains visible in GitHub checks and is a separate repository merge gate, not a code-review blocker for this approval.

LeonardEulerXYZ (approved):
Reviewed commits since the prior approved head 6c81aee1, including wallet-prompt retry retention, batch simulation health preservation, migration EIP-712 signature materialization, screening/Tenderly parity changes, and the final migration-success UI alignment. No new material blocker found. Verified under Node 24 with companion euler-sdks PR #96 at exact head 81adcca0: focused recent-delta suite 121/121, full suite 1,921 passed / 1 skipped, typecheck, targeted ESLint, production build, diff hygiene, and merge-tree against current development. The repository CI typecheck/test failures remain the known published-SDK sequencing gate; they reproduce with locked SDK 2.0.1 and clear with the companion SDK head.

composables/useReviewedExecution.ts
```diff
@@ -0,0 +1,685 @@
...
+    peerName: session?.peer?.metadata?.name ?? '',
+    peerUrl: session?.peer?.metadata?.url ?? '',
+  })
+}
+
+const captureWalletBinding = async (
+  config: ReturnType<typeof useConfig>,
+  signaturesEnabled: boolean,
+): Promise<WalletBinding> => {
+  const first = getWagmiAccount(config)
```
LeonardEulerXYZ: [P1] Gate retained-review retries on the current screening verdict. After a retryable wallet rejection, switching A→B→A resets the module-scoped screening state and begins an asynchronous re-screen, but this binding reads raw Wagmi state. The review still binds to A and can proceed while `useWagmi.address`/`isConnected` are intentionally hidden pending screening; neither direct nor batch confirmation gates on that state. Recheck a screening verdict bound to the captured account before signatures and again before irreversible dispatch, and cover rejected prompt + A→B→A + pending/blocked verdict.

composables/useReviewedExecution.ts
```diff
@@ -0,0 +1,685 @@
...
+        // Use wagmi/viem's established signer path. It supplies the required
+        // EIP712Domain type before calling the wallet; sending the SDK object
+        // directly makes Rabby sign an empty domain and Permit2 rejects it.
+        const signature = await signTypedDataAsync({
+          ...(slot.typedData as Parameters<typeof signTypedDataAsync>[0]),
+          account: slot.signer,
+        })
+        if (typeof signature !== 'string' || !/^0x[0-9a-f]+$/i.test(signature)) throw new Error('Wallet returned an invalid signature')
+        const normalized = signature as Hex
+        if (execution.requestSet.wallet.walletKind === 'eoa') {
```
LeonardEulerXYZ: [P1] `walletKind === eoa` currently means only ‘not recognized as Safe’, which includes supported Base Account smart wallets. The installed `@base-org/account` signer hashes typed data through `toReplaySafeHash()` and returns `wrapSignature(...)`; recovering the original EIP-712 signer here therefore rejects valid Base Account signatures before Permit2 or migration execution. Classify actual EOAs separately and validate contract-account signatures with the appropriate ERC-1271/ERC-6492/replay-safe path, or force approvals/signatures off for unsupported smart-account connectors. Add a Base Account typed-signature regression test.

LeonardEulerXYZ (changes requested):
A delayed independent challenge surfaced two execution-boundary regressions that survive verification on this exact head; this review supersedes my earlier approval. The published-SDK sequencing failure remains separately known and is not duplicated here.

LeonardEulerXYZ (commented):
Correction to my preceding review: the wallet re-screening finding is withdrawn. A second screening is not required by the intended policy, and I removed that inline comment. The review remains changes-requested only for the separate Base Account typed-signature compatibility issue.

LeonardEulerXYZ (approved):
Approved exact head `574b2c95cba09c0b8a8326be7de38a0dbcdbdf08`, reviewing only the three commits since prior reviewed head `359d9e8c8bad19fcbcebcffe580fb428863beeb4`.

No new material blocker found. The latest delta:

- restores Base Account / smart-wallet signature compatibility while keeping typed-data, slot, wallet-session, and finalized-request bindings intact;
- preserves all contributing intent revisions and policy subjects when SDK plan merging coalesces required approvals, and rejects non-exact merged amounts;
- carries the final confirmed EOA/Safe block through dispatch and performs a bounded second portfolio refresh after the serving subgraph catches up.

Validation under Node 24 with companion euler-sdks PR #96 at exact head `81adcca0a7d7c646d6fd4f14a09e467437c58bfd`: targeted recent-delta regressions 52/52, full Vitest 1,934/1,934, typecheck, targeted ESLint, production build, diff hygiene, and merge-tree against current `development`. Worktree remained clean.

The hosted `test`/`typecheck` failures remain the known published-SDK sequencing gate: Lite still pins SDK `2.0.1`, while this branch consumes APIs from SDK PR #96. They clear with the companion SDK head and remain a separate merge-readiness condition from this code approval.

pages/borrow/[collateral]/[borrow]/index.vue
```diff
@@ -298,7 +298,10 @@ const addToBatch = async () => {
       quote: borrow.borrowNeedsSwap.value ? borrow.borrowSwapEffectiveQuote.value ?? undefined : undefined,
     }
     const label = `Borrow ${snap.borrowAmount} ${bVault.asset.symbol}`
-    await addBatchEntry({ label, buildPlan: account => borrow.buildBorrowPlan(snap, account), subAccount, review: { type: 'borrow', asset: bVault.asset, amount: snap.borrowAmount, quoteFetchedAt: snap.needsSwap ? borrow.borrowSwapEffectiveQuoteFetchedAt.value : null } })
+    const quoteIntents = snap.quote
+      ? borrow.borrowSwapQuoteCards.value.find(card => card.quote === snap.quote)?.intents
+      : undefined
+    await addBatchEntry({ intent: quoteIntents?.[0] ?? borrow.createBorrowIntent(snap), label, subAccount, review: { type: 'borrow', asset: bVault.asset, amount: snap.borrowAmount, quoteFetchedAt: snap.needsSwap ? borrow.borrowSwapEffectiveQuoteFetchedAt.value : null } })
```
Comment (LeonardEulerXYZ) [inline pages/borrow/[collateral]/[borrow]/index.vue]: **[P1] Re-capture the settled swap-and-borrow intent for the batch path too.** The selected quote card can contain the preview intent created before the quote-derived borrow amount settled; the committed fixture at `tests/composables/useBorrowForm.test.ts:464-477` demonstrates `borrowAmount = 0n` in that cached intent while the live form later contains `5n`. Direct submit now correctly calls `createBorrowIntent(snapshot)`, but this path still prefers `quoteIntents[0]`, stores it unchanged in the cart, and the new `validators.ts:80-84` check deterministically rejects the otherwise valid batch as having no borrow amount. A batch-path regression should assert that the intent passed to `addBatchEntry` carries the settled amount.

features/reviewed-execution/review/post-tx-refresh.ts
```diff
@@ -0,0 +1,42 @@
...
+    confirmedBlockNumber,
+  )
+  if (!caughtUp) {
+    logWarn(
+      'reviewedExecution/subgraphPoll',
+      new Error(`subgraph did not catch up to block ${confirmedBlockNumber} in time`),
+    )
+    return
+  }
+  triggerRefresh(triggerPortfolioRefresh)
```
LeonardEulerXYZ: **[P1] Ensure this second refresh cannot join a pre-catch-up query.** `invalidateSdkQueries()` marks matching rows invalid but does not cancel/remove an in-flight `fetchQuery`; TanStack deduplicates the second consumer onto that same promise. Reproduction with the locked version produced one query-function call and returned the pre-catch-up value to both consumers, after which `isInvalidated` was false. Therefore this can poll the subgraph to the confirmed block and still publish the account/portfolio result started before catch-up. The post-catch-up path needs a freshness boundary that forces a new read (with a regression covering an in-flight first refresh).

LeonardEulerXYZ (changes requested):
A delayed independent challenge completed after my approval and reproduced two current-delta blockers. This review supersedes that approval. The remaining scope is bounded to the two inline findings below.

Validation evidence: the locked TanStack Query version issued one query-function call across invalidation and the second fetch, returning the pre-catch-up value to both consumers and clearing `isInvalidated` on resolution. The committed borrow regression fixture produces a quote-preview intent with `borrowAmount = 0n` before the settled form reaches `5n`; the batch path retains that preview intent and the new validator rejects it.

The prior SDK publication/CI sequencing gate remains separate.

LeonardEulerXYZ (approved):
Re-reviewed exact head `806e441c75948c8203ca5c9fef32afb5a5ca2c22` over `574b2c95cba09c0b8a8326be7de38a0dbcdbdf08`.

The two requested changes are resolved:

- Batch swap-and-borrow now captures the settled form snapshot immediately before creating the batch intent instead of reusing the quote-preview intent. The discriminator preserves the important mismatch (`preview borrowAmount = 0n`, live amount non-zero) and passes.
- The post-subgraph refresh now advances SDK query generations before invalidation. The prior in-flight coalescing case now uses a distinct key and performs a fresh call rather than publishing the pre-catch-up result as the second refresh.

I also independently reviewed the additional plugin-plan boundary change. Raw/preview plans are kept in a deep-frozen runtime sidecar, their canonical digests remain committed by the sealed review, exact materialized request calldata remains in the request set, and sidecar integrity is checked again before plugin refresh/execution. I found no new post-review rebuild authority or integrity gap.

Validation with Node 24/npm 11.10 and SDK PR #96 at `81adcca0a7d7c646d6fd4f14a09e467437c58bfd`:

- focused fix/trust-boundary suite: 83/83 passed
- full Vitest: 1,931 passed, 1 skipped
- typecheck: passed
- production build: passed
- changed-file ESLint: passed
- merge-tree with current `development`: clean
- desktop/mobile PR-preview smoke: rendered cleanly

No current-delta blocker found.

**Release gate remains:** Lite still pins registry SDK `2.0.1`, while this branch requires the unpublished materialized-execution API from SDK PR #96. Hosted `test` and `typecheck` are therefore red. This approval is source-level for the exact head, not merge readiness; do not merge until the SDK is published/pinned and hosted CI is green.

LeonardEulerXYZ (approved):
Approved at exact head `e84e3fa40d44c3ad9f11276091e02b3188dd640c` for the source delta.

Validated against a packed build of SDK #96 head `9a8bf0a26afa349922f91037b9800c55224e26aa` under Node 24/npm 11: 109 focused tests; full Lite suite 1,942 passed / 1 skipped; typecheck, targeted ESLint, production build, `git diff --check`, and merge-tree validation all passed. The Railway preview rendered the welcome and Explore shells without JS errors.

This is **not merge-ready yet**: SDK #96 still needs its requested documentation failure-path fixes, then must merge/publish; Lite must pin the released SDK in `package.json`/lockfile and hosted `test` + `typecheck` must turn green. Current CI failures are the expected registry-2.0.1 mismatch. Also, the preview `EULER_SDK_BRANCH` path remains mutable rather than exact-commit-pinned, so preview provenance should not substitute for exact-head composition evidence.

Dockerfile
```diff
@@ -31,13 +31,13 @@ WORKDIR /usr/src/app
 COPY package.json package-lock.json ./
 RUN npm ci
 
-COPY scripts/install-preview-sdk.sh scripts/install-preview-sdk.sh
+# Preview SDK branches are mutable. Including the Lite source snapshot in this
+# layer key makes each Lite revision resolve the current remote branch head.
+COPY . .
 ARG EULER_SDK_BRANCH
 ARG EULER_SDK_PNPM_VERSION=10
 RUN chmod +x scripts/install-preview-sdk.sh && scripts/install-preview-sdk.sh
```
LeonardEulerXYZ: Moving `COPY . .` above this step only refreshes the SDK when the Lite source snapshot changes. A rebuild of the same Lite SHA can reuse stale cached SDK bytes, while a no-cache rebuild can consume a newer unreviewed branch head. Please pass an exact reviewed SDK commit as a build argument, fetch/checkout it, verify `git rev-parse HEAD`, and include that SHA in the layer key. Two images for one Lite SHA must not contain different transaction-planner code.

tests/utils/migrationAuthorizationTxs.test.ts
```diff
@@ -62,6 +62,11 @@ const morphoAuthorizationRequest = (): MigrationAuthorizationRequest => ({
   },
 } as unknown as MigrationAuthorizationRequest)
 
+const morphoCleanupOnlyRequest = (): MigrationAuthorizationRequest => ({
+  ...morphoAuthorizationRequest(),
+  call: undefined,
```
LeonardEulerXYZ: This manually reconstructed fixture verifies Lite's interpretation, but it can drift independently from the SDK connector that produces the request. After bumping the published SDK, please retain at least one consumer integration test that feeds real installed-SDK Morpho connector output into the Lite compiler/encoder and asserts the cleanup-only result.

LeonardEulerXYZ (changes requested):
Updating my disposition after the completed producer-consumer review: the transaction behavior composes correctly, but this head is not yet safe to approve because the preview artifact is not reproducible and the committed consumer tests can drift from the SDK producer. The SDK publication/pin and green hosted CI remain hard merge gates as stated previously.

LeonardEulerXYZ (approved):
Approved exact head `7b38481eb7b99bae9c23ed54de3ca2c793cc738f`.

The final delta is limited to pinning `@eulerxyz/euler-v2-sdk` from `2.0.1` to released `2.1.0` in `package.json` and `package-lock.json`. The lockfile integrity matches npm, and npm reports `gitHead = bba9e288c28a24becfa7577ea03d040577288f64`, the merged SDK PR #96 commit.

This clears the prior producer/consumer publication gate. Validation against the exact registry artifact under Node 24/npm 11.10.0:

- clean `npm ci --ignore-scripts` and postinstall
- focused reviewed-execution / migration / refinance / query-cache / Tenderly suite: 21 files, 182 tests passed
- typecheck passed
- production build passed
- `git diff --check` passed
- synthetic merge tree with current `development` is clean
- exact installed-SDK discriminator passed: the real Morpho connector's already-authorized request produced no grant and exactly one cleanup revoke through Lite's encoder
- worktree remained clean

Hosted lint, typecheck, test, security review, and Railway preview checks are green. No current-head blocker found.

The optional mutable `EULER_SDK_BRANCH` preview override remains unsuitable as exact provenance for future unreleased SDK previews; this approval is grounded in the immutable `2.1.0` registry artifact and current-head checks, not that override.

Seranged (approved):

PR #823, opened 2026-08-20
feat: migrate address screening to the data-v3 compliance API

kasperpawlowski:
## Summary

Migrates wallet screening from the legacy TRM lambda to the data-v3 compliance endpoint (`POST /v3/compliance/address-screening`), and opens the internal screening route to the first-party SPAs that have no server of their own.

- **`server/utils/screening.ts`** (new): shared fail-closed upstream call — `X-API-Key` auth, data-v3 response-envelope parsing (`data.addressIsSuspicious === false` is the only pass verdict), `chain` omitted so upstream defaults to `ethereum`, VPN flag derived from edge-set headers only. Provider-neutral naming throughout (`services/trm.ts` → `services/screening.ts`): the screening vendor is an upstream implementation detail and must be swappable without touching this app.
- **`/api/internal/screen-address`**: unchanged client contract (`{ addressIsSuspicious }`), new upstream. Now also consumed cross-origin by create.euler.finance / redemptions.euler.finance / maglev.euler.finance via a **CORS exception in `cors.ts` scoped to exactly this path** for `https://*.euler.finance` origins (plus `CORS_ALLOWED_ORIGINS` / dev localhost) — no other internal route is exposed to sibling apps. It stays under `/api/internal/` deliberately: the consumers are our own apps, and `/api/public/` would advertise the endpoint to external integrators. Lookalike (`evil-euler.finance`, `euler.finance.evil.com`), non-https, and `null` origins are rejected with 403.
- **Env**: `WALLET_SCREENING_URI` → `ADDRESS_SCREENING_URI` + `ADDRESS_SCREENING_API_KEY`. **Both unset ⇒ screening disabled** (every address passes — fork-friendly default); only one set ⇒ fails closed, as does every error path once configured.

## Deployment prerequisites (before merge reaches prod)

- [x] data-v3 screening is **already active in production** (gitops sets `ADDRESS_SCREENING_ENABLED: true`; live probe returns 401 for missing keys, not 503)
- [x] Dedicated **restricted** API key with `compliance:screen` minted in euler-toolbox — the existing `EULER_SDK_V3_API_KEY` cannot carry this capability (data-v3 only allows it on restricted keys)
- [x] `ADDRESS_SCREENING_URI` + `ADDRESS_SCREENING_API_KEY` added to the euler-lite Doppler configs (dev/stg/prd)
- [ ] Consider raising the key's per-minute limit above the data-v3 default (100 rpm) — this key now also serves the three SPAs

## Test plan

- [x] `vitest run` — full suite, 1808 tests pass
- [x] New tests: envelope parsing, legacy flat shape now fails closed, missing env fails closed, 401/429/503/timeout/network fail closed, `X-API-Key` sent, `chain` omitted, VPN derived from headers not body, address never logged raw
- [x] CORS tests: `*.euler.finance` allowed on the screening path, lookalike/non-https/`null` origins rejected, and the exception does **not** extend to any other internal route
- [x] `vue-tsc --noEmit` clean
- [ ] Staging: verify flagged-address UX and the 503 fail-closed path against staging data-v3

## Summary by CodeRabbit

- **New Features**
  - Added address screening through the compliance API with VPN-aware checks.
  - Added scoped support for approved first-party Euler Finance web applications through the internal screening endpoint.
  - Screening can be disabled by leaving both configuration values unset; partial configuration fails closed.
  - Screening now verifies response-to-request address matching.

- **Bug Fixes**
  - Improved handling of invalid addresses, insecure endpoints, redirects, timeouts, upstream errors, and malformed responses.

- **Documentation**
  - Updated setup and architecture guidance for credentials, proxy routes, CORS behavior, and screening modes.


server/utils/screening.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (!screeningUri || !apiKey) {
+    logger.warn(
+      { ctx: logCtx },
+      'ADDRESS_SCREENING_URI or ADDRESS_SCREENING_API_KEY is not set — failing closed',
+    )
+    return { addressIsSuspicious: true }
+  }
+
+  try {
+    const resp = await fetchWithTimeout(screeningUri, UPSTREAM_FETCH_TIMEOUT_MS, {
```
LeonardEulerXYZ: **Blocking — bind the restricted key to a secure transport.** `ADDRESS_SCREENING_URI` is accepted verbatim and `fetch` follows redirects by default. In Node 24 I reproduced a cross-origin 302 target receiving the original `X-API-Key`. Please require an `https:` URL in deployed environments (allow only an explicit loopback exception for local development if needed) and set `redirect: 'error'`; test both an HTTP URI and a 3xx response. Failing closed after the key has already been disclosed is too late.

server/utils/screening.ts
```diff
@@ -0,0 +1,95 @@
...
+      body: JSON.stringify({ address, vpnIsUsed }),
+    })
+
+    if (!resp.ok) {
+      logger.warn({ ctx: logCtx, status: resp.status }, 'screening API non-ok response — failing closed')
+      return { addressIsSuspicious: true }
+    }
+
+    const body = await resp.json()
+    const isSuspicious = body?.data?.addressIsSuspicious !== false
```
LeonardEulerXYZ: **Blocking — bind a safe verdict to this request.** Data v3's response requires `address`, `addressIsSuspicious`, `screenedAt`, `resolvedChain`, `cached`, and `ruleVersion`, but this clears on the lone nested boolean. A partial response, or a response for address B while screening address A, can therefore pass. Validate the response schema and require the normalized returned `data.address` to equal the requested address before accepting `false`; add partial-shape and mismatched-address regressions.

LeonardEulerXYZ (changes requested):
## Coordinated screening migration review

**Verdict: changes requested** on exact head `cf4521bb0611f058c1156f123b815b7cdb4f007a`.

The client-side pending gate, scoped CORS path, fail-closed non-2xx/error handling, and server-only key placement are sound. Two credential/decision-boundary blockers remain:

1. The restricted API key can be sent over a non-HTTPS configured URL and is preserved across a cross-origin redirect. Node 24 reproduction: the redirect target received `X-API-Key`. Require HTTPS in production and reject redirects (`redirect: 'error'`), with regression tests.
2. A safe verdict is accepted without validating the required Data v3 response shape or binding `data.address` to the requested address. A partial or mismatched response can therefore clear screening.

Validation:
- `npx --yes npm@11.10.0 ci`
- `npm run test:run -- tests/server/screen-address.test.ts tests/server/cors-internal-api.test.ts tests/composables/useAddressScreen.test.ts` — 31/31 passed
- `git diff --check` — passed

Stack gate: Data v3 activation, restricted-key provisioning, and sufficient shared quota remain binding deployment prerequisites before Lite or any proxy-consuming SPA is deployed.

kasperpawlowski:
Review blockers addressed in 1a2946f6:

- **Safe responses bound to the request**: a verdict now clears only when the 200 response's `data.address` echoes the requested address (case-insensitive) alongside an explicit `addressIsSuspicious: false`. Missing/mismatched echo fails closed. Covered by new tests (mismatched, missing, case-variant echo).
- **Restricted-key transport**: `ADDRESS_SCREENING_URI` must be https (loopback http allowed for dev only) or the call fails closed without sending the key; the upstream fetch sets `redirect: 'error'` so `X-API-Key` can never follow a cross-origin redirect. Both tested.
- **VPN evidence**: when neither edge header is present, `vpnIsUsed` is now sent as `null` (unmeasured) instead of a fabricated `false`. Tested.

Full suite + typecheck green (11 screening tests, previously 8).

server/utils/screening.ts
```diff
@@ -0,0 +1,165 @@
...
+  logCtx: string,
+): Promise<ScreeningResult> {
+  const screeningUri = process.env.ADDRESS_SCREENING_URI
+  const apiKey = process.env.ADDRESS_SCREENING_API_KEY
+
+  // Screening is opt-in by configuration: with BOTH vars absent the app is
+  // treated as a deployment without a screening provider (e.g. a fork) and
+  // every address passes. Anything partial is a misconfiguration of a
+  // deployment that intended to screen — that fails closed.
+  if (!screeningUri && !apiKey) {
```
LeonardEulerXYZ: **Blocking — do not infer a safe opt-out from missing production configuration.** This branch conflates an intentional unscreened fork with a failed/missing production secret injection and authorizes every wallet in both cases. The architecture note acknowledges that removing both vars silently disables screening, but this PR adds no monitoring or explicit opt-out invariant. Require an explicit disable flag (or scope the absent-vars bypass to non-production); screened production must fail closed when configuration disappears.

server/utils/screening.ts
```diff
@@ -0,0 +1,165 @@
...
+      body: JSON.stringify({ address, vpnIsUsed }),
+      redirect: 'error',
+    })
+
+    if (!resp.ok) {
+      logger.warn({ ctx: logCtx, status: resp.status }, 'screening API non-ok response — failing closed')
+      return { addressIsSuspicious: true }
+    }
+
+    const body: unknown = await resp.json()
```
LeonardEulerXYZ: **Blocking — validate the complete Data v3 success contract before authorizing.** `resp.ok` admits 201/other 2xx responses, and this parser accepts an envelope containing only matching `address` plus `addressIsSuspicious:false`. The documented response requires HTTP 200 and the remaining required fields (`screenedAt`, `resolvedChain`, `cached`, `ruleVersion`). Exact-head probes confirm both partial 200 and partial 201 responses currently clear the wallet. Require status 200 and schema-validate the full response before returning safe.

LeonardEulerXYZ (changes requested):
## Exact-head screening rereview

**Verdict: changes requested** on `a41846e96ad919dfd2f1129e38e1f43acad3fc57`. Secure transport, redirect rejection, address echo binding, pending gating, and changed-file validation are materially improved. Two authorization-boundary issues remain: screening can be silently disabled by missing production configuration, and incomplete/non-200 Data v3 success envelopes are still accepted as clean.

Validation: 12 focused route tests, Nuxt typecheck, changed-file ESLint, `git diff --check`, and two reviewer schema probes passed; the probes confirm the residual acceptance behavior.

kasperpawlowski:
Exact-head review blockers addressed in f4abe062:

- **Full success-contract validation**: only HTTP 200 exactly (a clean 201 fails closed) with a complete envelope — `address`, `addressIsSuspicious`, `screenedAt`, `resolvedChain`, `cached`, `ruleVersion`, all type-checked — clears. Partial-200 and clean-201 regressions added (both reviewer probes now fail closed).
- **Unconfigured production is not an opt-out**: the both-vars-absent bypass is now scoped to non-production. Under `DOPPLER_ENVIRONMENT=prd` (the same existing convention the CORS/geo/rate middleware use — no new env vars) missing configuration fails closed with a warning log. Forks and previews keep the pass-through. Both behaviors tested; docs updated.
- Minor items: empty/whitespace edge headers now count as unmeasured VPN (`null`); the case-insensitivity test uses an address with alphabetic hex; the HTTPS/loopback transport rule was documented across .env.example/README/architecture in a41846e9 and extended here.

On the Cursor MEDIUM (origin forgeable by non-browser clients): accepted risk, by design — CORS can only ever bind browsers, the endpoint returns a single boolean, is per-IP rate-limited (CF-identity, fail-closed in prd), and the restricted key never leaves the server. Equivalent posture to the legacy public checker, now with strictly more controls.

253 server tests + typecheck green.

LeonardEulerXYZ (approved):
Exact-head rereview of `f4abe062c111b963907158da0fcaa3e5fbceaa3b`: approved. The strict success contract now requires HTTP 200 plus the complete six-field Data v3 envelope, and the previous partial-200 / clean-201 exploit fixtures fail closed. Production with `DOPPLER_ENVIRONMENT=prd` also fails closed when screening configuration is absent; non-production forks retain the documented opt-out. Whitespace-only VPN headers remain unmeasured, and the case-insensitive address-binding regression is genuine. Validation: 39 focused tests, two independent former-exploit probes, Nuxt typecheck, changed-file ESLint, `git diff --check`; hosted lint/test/typecheck and preview deployment are green.

Seranged (approved):

PR #836, opened 2026-08-28
Close retained transaction review policy gaps

dglowinski:
## Summary

- Bind transaction review and policy decisions to the exact wallet, account, chain, country, vault set, and canonical on-chain authority context.
- Fail closed when required screening or vault metadata cannot be resolved.

## Changes

- Scope unverified-vault acknowledgements to the reviewed account, chain, operation, and vault set.
- Bind asynchronous terms-of-use checks to the initiating account and chain.
- Screen both sides of rebalance and swap routes, preserve positive VPN verdicts across the browser/server boundary, and revalidate geo policy before wallet handoff.
- Key vault registry state by chain, discard stale cross-chain fetches, and verify canonical EVault, Securitize, and Earn authority before treating a vault as trusted.
- Interpret an EIP-5792 calls ID as a transaction hash only when it has the required hash shape.
- Resolve each approval row's symbol and icon from its actual token while preserving the existing copy and omitting spender/address and amount details.

## Linear tickets

Implemented:

- [LITE-282](https://linear.app/euler-labs/issue/LITE-282) apex review [EULE2-42]: Scope unverified-vault acknowledgements
- [LITE-295](https://linear.app/euler-labs/issue/LITE-295) apex review [EULE2-20]: Bind TOS checks to the current wallet
- [LITE-288](https://linear.app/euler-labs/issue/LITE-288) apex review [EULE2-5]: Geo-check both rebalance vaults
- [LITE-298](https://linear.app/euler-labs/issue/LITE-298) apex review [EULE2-31]: Preserve positive VPN screening results
- [LITE-300](https://linear.app/euler-labs/issue/LITE-300) apex review [EULE2-43]: Invalidate review when geo policy changes
- [LITE-293](https://linear.app/euler-labs/issue/LITE-293) apex review [EULE2-25]: Fail closed on unresolved vault metadata
- [LITE-299](https://linear.app/euler-labs/issue/LITE-299) apex review [EULE2-33]: Scope vault registry entries by chain
- [LITE-308](https://linear.app/euler-labs/issue/LITE-308) apex review [EULE2-36]: Verify canonical governor before trust

Partially implemented:

- [LITE-284](https://linear.app/euler-labs/issue/LITE-284) apex review [EULE2-8]: Label approvals with the actual token — approval rows now resolve symbol and icon from the actual approval token; the spender and amount details requested by the ticket are intentionally not shown.

Related:

- [LITE-310](https://linear.app/euler-labs/issue/LITE-310) apex review [EULE2-29]: Reconcile Safe execution before retry — canceled; the durable Safe reconciliation was reverted from this branch and only the EIP-5792 calls-ID hash-shape check is kept.

## Test plan

- [x] `npm run lint` (0 errors; 6 pre-existing warnings)
- [x] `npm run typecheck`
- [x] Approval decoder and review-compatibility suites (2 files, 41 tests)
- [x] `npm run test:run` (207 files, 1,966 tests)
- [x] GitHub CI build and full test suite


utils/pending-safe-reviewed-submission.ts
```diff
@@ -0,0 +1,138 @@
...
+export const PENDING_SAFE_REVIEWED_SUBMISSION_KEY = 'euler-lite:pending-safe-reviewed-submissions:v1'
+
+export interface PendingSafeReviewedSubmission {
+  reservationId: string
+  reviewId: Hash
+  reviewDigest: Hash
+  requestDigest: Hash
+  account: Address
+  chainId: number
+  callsId?: Hash
```
LeonardEulerXYZ: [blocking] Keep the EIP-5792 calls ID opaque instead of narrowing it to `Hash`. The standard permits any unique string (up to 4096 bytes), and `sendSafeAtomicCalls` already accepts any string returned as `result.id`. With a valid ID such as `safe-call-batch-123`, `attachPendingSafeCallsId` writes the record, but the next `loadPendingSafeReviewedSubmissions` rejects it at line 33 as invalid. From then on reconciliation, terminal clearing, and the manual UI all fail closed permanently for that Safe context. I reproduced this against the exact head. Please type/store the calls ID as a bounded string end-to-end; only interpret it as a transaction hash for receipt fallback after an explicit hash check, and add a non-hash calls-ID regression test.

LeonardEulerXYZ (changes requested):
One blocking Safe interoperability/liveness issue remains. The rest of the exact-head change set passed local tests, typecheck, lint, build, and preview smoke.

utils/pending-safe-reviewed-submission.ts
dglowinski: Fixed in 6cf10a97. Calls IDs are now validated and stored as opaque strings bounded to 4096 bytes end-to-end. Receipt fallback runs only after an explicit transaction-hash shape check. Added non-hash persistence, reconciliation, current-session polling, and over-bound regression coverage; Node 24 lint, typecheck, build, and the full 1,982-test suite pass.

features/reviewed-execution/policy/app-policy.ts
```diff
@@ -19,31 +23,85 @@ const addressOfSubject = (subject: string): Address | undefined => {
...
+    if ((hardGeoRequired || softGeoRequired) && country.value === undefined) {
+      throw new Error('Regional availability is still loading')
+    }
+    if ((hardGeoRequired || softGeoRequired) && country.value === null) {
+      throw new Error('Regional availability could not be determined')
+    }
+    if (hardGeoRequired && exactVaults.some(entry => isVaultBlockedByCountry(entry.address, { asset: entry.vault.asset }))) {
+      throw new Error('This operation is not available in your region')
+    }
+    if (softGeoRequired && exactVaults.some(entry => isVaultRestrictedByCountry(entry.address, { asset: entry.vault.asset }))) {
```
LeonardEulerXYZ: [blocking] Apply final geo and unverified-vault policy to both sides of a retained vault swap. The collateral-swap review intent contains only `swapQuote`; `collectPlanningRequirements` classifies `vaultIn` as the source vault and `receiver` as an account, but does not classify `swapQuote.verify.vault` as the target vault. Because `exactVaults` is then non-empty, this branch suppresses the richer effect-derived fallback. A target vault that becomes restricted or loses canonical verification after the form guard is unmounted can therefore pass final review policy. My exact-head discriminator produced `[source]` instead of `[source, target]`. Please include the quote's target vault in exact requirements and add a two-vault swap regression covering geo and acknowledgement checks.

features/reviewed-execution/adapters/app-clients.ts
```diff
@@ -22,19 +30,71 @@ export const createAppEoaClients = ({
...
+          const pending = findPendingSafeReviewedSubmission(storage, identity.account, identity.chainId)
+          if (pending?.callsId) {
+            const reconciliation = await reconcileSafeTransactionExecution({
+              callsId: pending.callsId,
+              walletProvider: provider,
+              publicClient: publicClient as never,
+            })
+            if (reconciliation.state === 'success' || reconciliation.state === 'reverted'
+              || reconciliation.state === 'cancelled' || reconciliation.state === 'failed') {
+              clearPendingSafeReviewedSubmission(storage, pending.reservationId)
```
LeonardEulerXYZ: [blocking] Do not continue into a replacement reservation after reconciliation proves the prior proposal succeeded. This clears the old record, refreshes portfolio data, then line 75 reserves the already-prepared review and `SafeExecutionAdapter` immediately submits it. Retrying the same `requestDigest` after a timeout can therefore execute the action twice; even a different digest was prepared before the reconciled success changed portfolio state. I seeded a successful prior calls ID and retried the same identity: `reserveSubmission` returned a fresh reservation. Treat success as a terminal stop that requires refresh/re-prepare/review, and add a test asserting no second `wallet_sendCalls`.

features/reviewed-execution/adapters/safe.ts
```diff
@@ -49,23 +61,48 @@ export class SafeExecutionAdapter implements ExecutionTransportAdapter {
...
     await callbacks.beforeDispatch(0)
 
+    const identity: SafeSubmissionIdentity = {
+      reviewId: execution.reviewId,
+      reviewDigest: execution.reviewDigest,
+      requestDigest: execution.requestDigest,
+      account: execution.requestSet.wallet.account,
+      chainId: execution.requestSet.wallet.chainId,
+    }
+    const reservationId = await this.client.reserveSubmission(identity)
```
LeonardEulerXYZ: [blocking] Re-run the final policy/wallet guard after `reserveSubmission` and immediately before this wallet handoff. Reservation can await calls-status reconciliation and `onReconciled`; account, chain, connector session, policy, or reviewed state may drift during that gap while this adapter retains the old provider. In my exact-head discriminator I paused reservation, changed the binding verdict, then resumed: `sendCalls` was still invoked and dispatch succeeded because the only `beforeDispatch` call is above the await. Invoke the full `beforeDispatch` callback again here and cover deferred reservation plus account/chain/session drift.

components/entities/operation/OperationStepsList.vue
```diff
@@ -1,11 +1,20 @@
 <script setup lang="ts">
 import type { DisplayStep } from '~/utils/stepDecoding'
-import { formatNumber } from '~/utils/string-utils'
+import { formatUnits } from 'viem'
+import { formatNumber, shortenAddress } from '~/utils/string-utils'
```
LeonardEulerXYZ: [blocking] Preserve the exact allowance amount when the approval token is unknown. `stepDecoding` intentionally emits values such as `123 base units`, but `formatNumber` coerces that string with `Number(...)` and renders `-`. The visible consent row becomes effectively `- 0x… for spender 0x…`; only the tooltip retains the amount, so the default review hides the allowance scope. My discriminator confirmed `formatNumber('123 base units', 8, 0) === '-'`. Render this structured/raw amount without numeric formatting and add a rendered-component test, not only a decoder test.

utils/pending-safe-reviewed-submission.ts
```diff
@@ -0,0 +1,140 @@
...
+export const clearHashlessPendingSafeReviewedSubmission = (
+  storage: StorageLike,
+  context: { reservationId: string, account: Address, chainId: number, confirmedAbsent: boolean },
+) => {
+  if (!context.confirmedAbsent) throw new Error('Confirm that Safe contains no proposal before clearing this lock.')
+  const record = loadPendingSafeReviewedSubmissions(storage).find(candidate => candidate.reservationId === context.reservationId)
+  if (!record || record.callsId || record.account !== getAddress(context.account) || record.chainId !== context.chainId) {
+    throw new Error('The pending Safe reservation context changed. Review it again before clearing.')
+  }
+  clearPendingSafeReviewedSubmission(storage, context.reservationId)
```
LeonardEulerXYZ: [blocking] A hashless record can still belong to a live wallet handoff in another tab, so confirmation alone is insufficient to delete it here. The Web Lock is released after reservation and before `wallet_sendCalls` resolves: tab A can be waiting on the Safe prompt while tab B sees the hashless record and truthfully confirms that no proposal exists yet, clears it, then tab A approves. Calls-ID attachment fails because the reservation vanished, and the durable duplicate interlock is gone. I reproduced this with a paused `wallet_sendCalls`; the cross-tab clear succeeded while the handoff was live. Add an owner/lease or conservative staleness boundary that distinguishes a live attempt from an orphan, plus a two-realm regression.

LeonardEulerXYZ (changes requested):
The opaque EIP-5792 calls-ID blocker is fixed at this head, and the full Node 24 gate passes. Five independently reproduced transaction-policy/consent/lifecycle blockers remain; see inline comments. Local validation: 208 test files / 1,978 tests passed (1 skipped), typecheck passed, lint 0 errors / 6 existing warnings, production build passed. I also ran local throwaway discriminators against the exact head; they are not committed.

components/entities/operation/OperationStepsList.vue
dglowinski: Fixed in e4b9dc25. Unknown-token approval amounts are explicitly marked as raw structured values and rendered verbatim instead of passing through numeric formatting. Added an SSR-rendered component regression asserting 123 base units and the full token address remain visible.

features/reviewed-execution/adapters/app-clients.ts
dglowinski: Fixed in e4b9dc25. Successful prior-proposal reconciliation now clears the terminal record, refreshes reconciled state, and stops before any replacement reservation or wallet dispatch. Added a regression asserting the retry is rejected and no replacement record remains.

LeonardEulerXYZ (approved):
Re-reviewed exact head `e4b9dc25664ae14c004a0a1085cf25d3131d2115` against the five prior blockers and the complete incremental delta from `6cf10a97`.

All five blockers are resolved:
- collateral-swap final policy now includes both source and target vaults;
- successful prior Safe reconciliation stops replacement submission and requires fresh preparation/review;
- wallet and policy binding is revalidated after reservation, immediately before `wallet_sendCalls`;
- unknown-token approval amounts remain visibly exact as raw base units;
- the browser-wide Safe lock remains held across the live wallet handoff until calls-ID attachment or a bounded exit path.

Validation:
- focused regression set: 8 files, 51 tests passed;
- full suite: 210 files passed, 1 skipped; 1,985 tests passed, 1 skipped;
- `npm run typecheck`: passed;
- `npm run lint`: 0 errors, 6 existing warnings;
- `npm run build`: passed;
- real headed Chromium, two-tab Web Locks probe: concurrent acquisition blocked while the handoff lock was live and succeeded after release;
- GitHub CI, security review, and Railway preview checks: green;
- incremental security/supply-chain sweep: no new dependency, workflow, secret, or hidden network-execution surface.

No remaining blocking findings.

composables/useReviewedExecution.ts
```diff
@@ -581,7 +586,7 @@ export const useReviewedExecution = () => {
         if (current.validity.policyVersionDigest !== currentPolicyVersionDigest()) {
           throw new Error('Policy configuration changed after review')
         }
-        const policy = await resolveAppPolicy(current.requestSet)
+        const policy = await resolveAppPolicy(current.requestSet, Date.now(), current.intents)
```
LeonardEulerXYZ: [P1] Enforce the screened identity at the reviewed-execution boundary

This gates ordinary `useWagmi().address` consumers, but `captureWalletBinding()` and `executionRuntime()` still call `getWagmiAccount(config)` directly. After allowed A → B → A, `address` is fail-closed while A is being screened again, yet a retained review sees raw A and can reach dispatch. Separately, `executionRuntime()` samples a connector before its async Safe classification and closes that connector into the adapters; B can be captured while later wallet-binding samples see A. Build the adapters from one twice-sampled, screened account/chain/connector/session context and recheck that same context at each irreversible handoff. Please add a discriminating A → B → A test covering pending screening and connector replacement.

features/reviewed-execution/policy/app-policy.ts
LeonardEulerXYZ: [P1] Apply final geo policy to arbitrary intent assets

`collectPlanningRequirements()` records `tokenIn`/`tokenOut` and other arbitrary assets in `requirements.assets`, but this policy resolves and checks only `requirements.vaults`. A blocked/restricted swap token with no vault therefore disappears from retained-plan enforcement once the route-local selector guard is unmounted; repreparing the retained batch still passes because the final policy never evaluates that asset. Resolve the committed assets and apply hard/soft rules with planner direction (source vs acquisition target), then cover a retained batch whose arbitrary token becomes restricted.

features/reviewed-execution/policy/app-policy.ts
```diff
@@ -19,31 +23,85 @@ const addressOfSubject = (subject: string): Address | undefined => {
...
-          .filter(subject => getVault(subject) && !isVerifiedVault(subject))
-        if (unverified.some(subject => !hasUnverifiedVaultAcknowledgement(subject))) throw new Error('Unverified vault acknowledgement does not cover the execution')
+        const unverified = exactVaults
+          ? exactVaults.filter(entry => !canonicallyVerified(entry)).map(entry => entry.address)
+          : requestSet.effects
+              .flatMap(effect => effect.policySubjects)
+              .filter(subject => subject.kind === 'vault-or-contract')
+              .map(subject => getAddress(subject.value))
+              .filter(subject => getVault(subject) && !isVerifiedVault(subject))
+        if (unverified.some(subject => !hasUnverifiedVaultAcknowledgement(subject, {
```
LeonardEulerXYZ: [P2] Preserve operation scope in the final acknowledgement check

The UI records `chainId/account/operation/vault-set`, but this lookup supplies only chain and account, so any stored acknowledgement containing the vault satisfies every later operation. A throwaway regression recording consent for `lend-vault` and checking `borrow-vault` fails with `expected true to be false`. Either carry a stable operation/planner-set identity into this final policy lookup or explicitly change the consent contract; the current implementation contradicts the operation-scoped invalidation added by this PR.

composables/useVaults.ts
```diff
@@ -994,10 +994,11 @@ const getEarnVault = async (address: string): Promise<EulerEarn> => {
   const { getVault: registryGetVault, set: registrySet } = useVaultRegistry()
   const normalizedAddress = getAddress(address)
   const { earnVaults } = useEulerLabels()
+  const targetChainId = resolveTargetChainId()
 
   const fetchAndStoreEarnVault = async () => {
-    const vault = await useVaultRegistry().fetchVaultByType(normalizedAddress, 'earn') as EulerEarn
-    registrySet(normalizedAddress, vault, 'earn')
+    const vault = await useVaultRegistry().fetchVaultByType(normalizedAddress, 'earn', targetChainId) as EulerEarn
+    registrySet(normalizedAddress, vault, 'earn', undefined, targetChainId)
```
LeonardEulerXYZ: [P2] Reject stale direct-fetch results before returning them

Passing `targetChainId` keeps the registry write on chain A, but after an A → B switch this helper still returns the fetched chain-A object to its current caller. The same pattern exists in `updateVault`, `updateEarnVault`, `getEscrowVault`, `updateEscrowVault`, and `getSecuritizeVault`; live pages immediately assign those return values to form state. A direct module probe reproduced an active chain-10 registry alongside a returned chain-1 vault. Recheck active chain/generation after each await and reject/drop the stale result before both write and return.

utils/safeWalletTransactions.ts
```diff
@@ -335,7 +341,7 @@ export const waitForSafeTransactionExecution = async ({
         try {
           const rawStatus = await withStopSignal(walletProvider.request({
             method: 'wallet_getCallsStatus',
-            params: [submittedHash],
+            params: [callsId],
```
LeonardEulerXYZ: [P2] Verify that calls status belongs to the submitted batch

For opaque IDs this response is now the authoritative bridge to an execution hash, but `parseCallsStatus()` discards the EIP-5792 `id`, `chainId`, and `version`. A stale/misrouted status for another batch can therefore provide `status: 200`, `atomic: true`, and an unrelated receipt hash; the public receipt check only proves that unrelated transaction mined. Carry the reviewed chain here and require the response identity to match `callsId`/chain/supported version before accepting terminal state or receipt hashes, with a mismatched-response regression.

LeonardEulerXYZ (changes requested):
Blocking at exact head `f60844a80f65e1d8d037bd36bb2a732ea0752f78`. I discarded the earlier approval/Fuul and durable Safe scope, then re-reviewed the surviving policy, screening, registry, and opaque calls-ID paths. Five findings remain inline.

Validation: Node 24.14.0 / npm 11.10.0; focused 9 files / 73 tests passed; full suite 206 files / 1,961 tests passed (1 file and 1 test skipped); typecheck passed; lint passed with 6 existing warnings; production build passed. Exact-head Railway preview, GitHub test/typecheck/lint/security checks, six-route prod-vs-preview browser smoke, unverified-vault Cancel/Yes smoke, and malformed/valid screening API probes all passed.

Scope notes: opaque IDs are correctly kept out of hash receipt APIs, and the now-documented current-session-only Safe timeout behavior matches the reduced scope. The reported chain-key prefix collision is not real: `10:`, `100:`, and `11155111:` do not match the `1:` prefix.

The branch is also six commits behind current `development` (including SDK 2.1.1) with one overlapping page; the synthetic merge is conflict-free, but please refresh onto current `development` before the fix-round approval.

LeonardEulerXYZ (approved):
Approved at exact head f60844a80f65e1d8d037bd36bb2a732ea0752f78 after correcting the review scope. I rechecked only the scope-removal delta e4b9dc25..f60844a8: approval/Fuul display removal, durable Safe reconciliation removal, compatibility-fixture restoration, and corresponding docs/tests. That delta is coherent and all validation passed. The five concerns in my previous changes-requested review were inherited from the already-approved head and were not introduced by this delta; they are retracted as blockers here and should be handled separately if desired.

LeonardEulerXYZ (approved):
Reviewed only the new delta `f60844a8..516ae219` (commit `fix: show correct token in approval reviews`).

No blocking findings. Approval rows now derive token identity from the resolved approval item and resolve the corresponding metadata from the review context, so a borrow denominated in USDC correctly shows WETH for its collateral approval and Permit2 signature. The change is isolated from non-approval display rows.

Validation:
- `tests/utils/stepDecoding.test.ts` + `tests/reviewed-execution/inventory.test.ts`: 41 tests passed
- Additional local throwaway `swapFromAsset` discriminator passed and was removed
- `npm run typecheck` passed
- ESLint passed on all three changed files
- `npm run build` passed
- `git diff --check f60844a8..516ae219` passed
- Current GitHub lint, test, typecheck, security-review, and preview checks passed
- Current `development` has no path overlap with this delta; synthetic merge is clean

Browser smoke not run: this is a pure decoder mapping with unchanged rendering components; the focused output tests exercise the changed behavior directly.

LeonardEulerXYZ (changes requested):
Reviewed the Apex audit-fix cohort at exact head `516ae219daab21ad93bc6e237724cc5c0ba262cf`, including composition with current `development` (`3d4812fec6cf4808cd1f7ec90d971da33bb57e47`).

## Changes requested

### [P1] Preserve the exact unverified-vault acknowledgement scope at final handoff

LITE-282 is incomplete at the authoritative `resolveAppPolicy` check. `recordUnverifiedVaultAcknowledgement` stores the intended exact tuple `(chainId, account, operation, vault set)`, but `hasUnverifiedVaultAcknowledgement` accepts only `(chainId, account, individual vault)` and scans all stored acknowledgements by membership. Consequently, consent recorded for operation A over `[vaultA, vaultB]` is accepted for operation B and for a reduced `[vaultA]` context. The UI guard does use the exact key, but reviewed execution can outlive that mounted route; the final policy must independently prove that the retained operation and exact vault set are covered.

Existing line discussion: https://github.com/euler-xyz/euler-lite/pull/836#discussion_r3916472083

Please pass a complete context derived from the retained intent set into the final check and require the exact stored key (or an equivalently strict binding), then add a regression that rejects both cross-operation reuse and vault-set subset/superset reuse.

## Remaining Apex items

No additional blocking findings in LITE-295, LITE-288, LITE-298, LITE-300, LITE-293, LITE-299, or LITE-308. LITE-284 remains partial exactly as declared in the PR description; LITE-310 is canceled/out of scope.

Validation on the clean synthetic merge:
- targeted audit-fix suite: 15 files / 155 tests passed
- full suite: 210 files passed, 1 skipped / 1,977 tests passed, 1 skipped
- typecheck passed
- lint passed (0 errors; 6 existing warnings)
- production build passed
- `git diff --check` passed
- exact LITE-282 discriminator failed as expected (`true` returned where exact-scope policy requires `false`)

The Cursor registry-prefix comment is not a finding: keys include the `:` delimiter, so `10:` and `100:` do not match the `1:` prefix.

LeonardEulerXYZ (approved):
Rechecked the LITE-282 consent model at unchanged exact head `516ae219daab21ad93bc6e237724cc5c0ba262cf`.

I am retracting my prior changes-requested finding. It conflated two distinct invariants:

- The UI guard's exact `(chainId, account, operation, vault set)` key ensures that changing the active operation or vault set reopens the acknowledgement prompt.
- The final reviewed-execution policy checks the narrower risk invariant: every unverified vault in the executable plan must have been explicitly acknowledged by this account on this chain.

The latter is correctly satisfied by per-vault membership. A reduced set contains only already-acknowledged vaults; an expanded set cannot pass unless every added unverified vault appears in an explicit same-account/same-chain acknowledgement. Reuse across operation names does not weaken the user's consent to each vault's unverified status.

The local throwaway regression I cited encoded a stricter requirement than the final policy's intended contract, so its failure was not evidence of a defect. I deleted the stale inline finding.

No blocking findings in the requested Apex audit-fix cohort. Prior validation remains SHA-bound to this unchanged head and its clean composition with current `development`: targeted 155 tests passed; full 1,977 passed / 1 skipped; typecheck, lint, build, and `git diff --check` passed.

Browser smoke not rerun: this recheck changed no source and resolved a policy-semantics interpretation, not a UI/runtime delta.

Seranged (approved):

PR #844, opened 2026-09-02
Restore spy-mode transaction review preparation

dglowinski:
## Summary
- restore direct and multi-operation batch review preparation for spy-mode accounts
- prepare complete read-only calldata and Tenderly inputs without requiring a connected wallet
- keep spy-mode preparations outside executable reviewed-execution authority

## Root cause
PR #822 routed review through strict connected-wallet capture. Spy mode has an effective account but no connector, so preparation failed before the review modal could render.

## Changes
- add deterministic approval-only wallet binding for spy-mode preparation
- route direct and batch spy reviews through the read-only path
- retain independent execution guards in both review surfaces
- add direct and two-operation batch regression coverage

## Test plan
- [x] npm run build
- [x] npm run lint
- [x] npm run typecheck
- [x] npm run test:run
- [x] local spy-mode direct review renders with Copy calldata and Tenderly available
- [x] local spy-mode batch review renders without Preparation failed while Execute remains disabled

### DIFF
--- a/components/BatchReviewModal.vue
+++ b/components/BatchReviewModal.vue
@@ -433,10 +433,11 @@ const hasTenderlyFailed = computed(() => Boolean(tenderlyUrl.value && tenderlyEr
 const simulateOnTenderly = () => simulateBatchOnTenderly(preparedExecution.value ?? undefined)
 
 const isConfirmDisabled = computed(() =>
-  isSpyMode.value || isExecuting.value || hasPendingDetachedExecution.value || isPreparing.value || isSimulating.value || !canExecuteBatch.value || !!prepareError.value,
+  isSpyMode.value || preparedExecution.value?.readOnly === true || isExecuting.value || hasPendingDetachedExecution.value || isPreparing.value || isSimulating.value || !canExecuteBatch.value || !!prepareError.value,
 )
 const blockedReason = computed(() => {
   if (isSpyMode.value) return 'Connect a wallet to execute — disabled in spy mode'
+  if (preparedExecution.value?.readOnly) return 'This read-only review cannot be executed'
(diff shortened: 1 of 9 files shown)


LeonardEulerXYZ (approved):
Reviewed exact head `c06c6d8d683a4aa3380c2064a605e2e8682c544d`.

No blocking findings. The review/execution boundary remains fail-closed: spy mode can prepare direct and batch review artifacts without a connector, while acceptance/execution still requires the correct connected wallet binding.

Verified:
- focused reviewed-execution tests: 65 passed
- full suite: 1,948 passed, 1 skipped
- lint, typecheck, and production build passed under Node 24 / npm 11.10.0
- PR preview smoke: direct withdraw review rendered prepared calldata with `Spy mode (read-only)` disabled; batch review rendered the aggregate with `Execute batch` disabled and the explicit spy-mode reason

LeonardEulerXYZ:
Non-blocking test follow-up: please consider pinning the read-only authority boundary with one real `prepareReadOnly` regression test. Assert that its returned `reviewId` is absent from `getReviewedExecution(...)`, and that `accept(reviewId, reviewDigest)` rejects before any coordinator/wallet dispatch.

The current implementation is safe: `composables/useReviewedExecution.ts` returns on `readOnly` before populating `executions`, `runtimePluginPlans`, and `reviewGenerations`, and both modals independently disable execution. However, the new direct/batch tests mock `prepareReadOnly`. A mutation that moved those three registrations above the read-only return survived both the focused 65-test set and the full 1,948-test suite, so this central invariant is not yet guarded against regression. This does not change my approval.

Seranged (approved):

PR #847, opened 2026-09-03
Fix transaction review snapshot consistency

dglowinski:
## Summary
- Bind every direct reviewed transaction to one immutable action snapshot so edits during preparation cannot change what the modal presents.
- Treat eager plans and quote-time intents as optimizations: reuse them only when their transaction semantics exactly match the fresh action intent; otherwise rebuild.
- Preserve position-funded authorization context by showing the supplying position separately from the target position.

## Changes
- Replace direct review opening with a shared capture-first API across all 21 review launch sites, including lend, earn, borrow, multiply, swap/refinance, migration, rewards, collateral controls, and every repay mode.
- Capture account and chain context across asynchronous sub-account resolution, and use captured inputs consistently for planning, preflight simulation, and reviewed execution preparation.
- Make batch entries always provide a fresh add-time intent plus an optional warmed candidate; the batch boundary performs the same semantic match before accepting cached work.
- Deep-clone and freeze captured intents and presentation inputs, and add an inventory gate that prevents direct forms from bypassing the capture boundary.
- Show `From Position N` or `From Deposits` for source-funded direct and batch operations.

## Test plan
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm run test:run` (207 files, 1,974 tests)
- [x] Deferred-preparation regressions keep intent and presentation equivalent after form or source changes
- [x] Matching warmed intents are reused; stale warmed intents rebuild in both direct and batch paths


composables/useTxBatch.ts
```diff
@@ -2195,8 +2201,13 @@ export const useTxBatch = () => {
...
 
-    const { intent, ...presentation } = entry
-    const entryId = entry.intent.intentId
+    const { intent: currentIntent, preparedIntent, ...presentation } = entry
+    const selectedIntent = selectMatchingPreparedIntents(
+      preparedIntent ? [preparedIntent] : undefined,
+      [currentIntent],
+    )[0]!
+    const intent = deepFreezeSerializable(selectedIntent) as OperationIntent
+    const entryId = intent.intentId
```
LeonardEulerXYZ: **[P2] Preserve a fresh identity for every batch insertion**

`preparedIntent` is a reusable quote-time artifact, but this now derives the cart row identity from that warmed DTO. If the same semantically matching quote-backed operation is added twice *sequentially* (a case the comment above explicitly says remains supported), both calls select the same prepared `intentId:revision`. The cart then contains duplicate identities: presentation/plan maps and Vue keys alias, removing one ID removes both rows, and authoritative preparation rejects the set in `validateIntentSet()` with `Duplicate intent revision warm:1`.

I reproduced this directly on the exact head by matching two fresh equivalent intents against one warmed candidate; both selected the warmed ID and validation rejected the pair. Please keep or mint the fresh add-time identity for each row while reusing only the warmed semantics/work, and add a regression that adds the same prepared candidate twice and reaches successful whole-cart preparation.

LeonardEulerXYZ (changes requested):
One blocking batch-liveness issue found on exact head `3d8264746bf3a163833a87cb96f1ea174d1c9e78`.

I rechecked the snapshot/TOCTOU paths, all 21 direct review launch sites, migration debt/quote invalidation, source/receiver bindings, warmed-plan matching, and whole-cart preparation. The earlier migration display-versus-intent report is fixed on this head.

Validation:
- Independently reproduced the duplicate warmed-intent identity below; authoritative validation returns `Duplicate intent revision warm:1`.
- 12 focused suites / 130 tests passed locally on Node 24 + npm 11, including batch, execution review, intent factories, borrow/multiply, repay, and inventory coverage.
- Exact-head remote lint, typecheck, test, security, CodeRabbit, and Railway checks are green.

Non-blocking follow-up: savings-funded new borrow/multiply reviews do not currently pass the distinct savings `sourceSubAccount` into the review presentation, so the new source-position pill can be absent when source and target differ. I would add route-level assertions for those two branches, but it does not change execution semantics and is not the reason for requesting changes.

LeonardEulerXYZ (approved):
Reviewed exact head `ef3fb1940d1c81254c8e3fb557141d0c8995a712`.

The latest fix resolves the prior blocker: repeated adds may reuse matching warmed semantics, but each batch row keeps its fresh `intentId`, `revision`, and metadata, so whole-cart validation no longer collapses duplicate warmed quote instances. Savings-funded borrow and multiply now also bind the actual source position in direct and batch review presentation.

Validation:
- exact-head focused regression coverage passed, including the old duplicate-identity discriminator
- exact-head full suite: 217 files, 1,997 tests passed; typecheck, lint, and production build passed
- conflict-free synthetic merge with current `development`: 218 files, 2,072 tests passed; typecheck and production build passed
- independent whole-PR adversarial pass found no remaining value-moving, liveness, or review/execution parity blocker

Seranged (approved):

